PR #12, opened 2016-12-09
misc refactoring + integration tests

mhowlett:


src/Confluent.Kafka/Consumer.cs
```diff
@@ -1,175 +1,346 @@
...
-                IntPtr opaque)
+        private LibRdKafka.RebalanceDelegate rebalanceDelegate;
+        private void RebalanceCallback(
+            IntPtr rk,
+            ErrorCode err,
+            /* rd_kafka_topic_partition_list_t * */ IntPtr partitions,
+            IntPtr opaque)
         {
-            var partitionList = SafeKafkaHandle.GetTopicPartitionOffsetList(partitions);
+            var partitionList = SafeKafkaHandle.GetTopicPartitionOffsetErrorList(partitions).Select(p => p.TopicPartition).ToList();
```
treziac: @mhowlett @edenhill 
I just came accross that. Can you explain why we don't retrieve the offset and pass Offset.Invalid instead? I must be missing something.

PR #17, opened 2017-01-24
making CLS compliant

mhowlett:
Leading underscores on enum values are not CLS compliant so I've prefixed local ErrorCode values with RD_ instead. Alternatives are prefixing with L_ or LOCAL_ (and broker error codes with B_ or BROKER_), or just removing the underscore altogether.

prefixing with RD_ is my preferred option, but I don't like any of them to be honest. Seeking the opinion of others.

(and yes, whoops, accidentally pushed this branch to the confluentinc repo, not my own).

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -657,7 +657,7 @@ namespace Confluent.Kafka
                 {
                     case ErrorCode.NoError:
                         return true;
-                    case ErrorCode.Local_PartitionEof:
+                    case ErrorCode.Local_PartitionEOF:
                         OnPartitionEOF?.Invoke(this, message.TopicPartitionOffset);
                         return false;
                     default:
@@ -761,8 +761,10 @@ namespace Confluent.Kafka
         /// </remarks>
         public async Task<CommittedOffsets> CommitAsync(Message message)
         {
-            if (message.Error.Code != ErrorCode.NO_ERROR)
+            if (message.Error.Code != ErrorCode.NoError)
+            {
                 throw new InvalidOperationException("Must not commit offset for errored message");
+            }
             return await CommitAsync(new List<TopicPartitionOffset> { new TopicPartitionOffset(message.TopicPartition, message.Offset + 1) });
         }
 
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -657,7 +657,7 @@ namespace Confluent.Kafka
                 {
                     case ErrorCode.NoError:
                         return true;
-                    case ErrorCode.Local_PartitionEOF:
+                    case ErrorCode.Local_PartitionEof:
                         OnPartitionEOF?.Invoke(this, message.TopicPartitionOffset);
                         return false;
                     default:
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -399,7 +399,7 @@ namespace Confluent.Kafka
             IntPtr opaque)
         {
             var partitionList = SafeKafkaHandle.GetTopicPartitionOffsetErrorList(partitions).Select(p => p.TopicPartition).ToList();
-            if (err == ErrorCode._ASSIGN_PARTITIONS)
+            if (err == ErrorCode.Local_AssignPartitions)
             {
                 var handler = OnPartitionsAssigned;
                 if (handler != null && handler.GetInvocationList().Length > 0)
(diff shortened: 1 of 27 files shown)


edenhill:
This came up on the Python client as well for a similar reason:
https://github.com/confluentinc/confluent-kafka-python/issues/88

Since we're not really exposing librdkafka itself I don't think RD_ is a good fit.
How about "CLIENT_" (or "LOCAL_") and "BROKER_"? We should standardize this across all three clients.

(I think confluent-repo feature-wip branch are okay as long as they are removed after merge)

mhowlett:
The main potential problem I see with prefixing with full names LOCAL_, CLIENT_ or BROKER_ is whether or not that could lead to a name being misinterpreted as I think it's not obvious these words are meant as prefixes. e.g. CLIENT_UNKNOWN_TOPIC. 

We could use a double underscore, though I'm not fully into that idea as this is unusual in most contexts I can think of. e.g. CLIENT__UNKNOWN_TOPIC - that implies fairly strongly to me that CLIENT is a prefix and not to be read as part of the name.

edenhill:
I dont think the various prefixes will be that confusing - scanning through the error codes will show two distinct groups of errors (CLIENT_ and BROKER_) - but most error codes are never meant to be handled specifically anyway. A few exceptions are:
 * CLIENT_QUEUEUUEUEUE_FULL - prefix makes sense
 * CLIENT_ALL_BROKERS_DOWN - prefix sort of makes sense
 * CLIENT_PARTITION_EOF - prefix doesn't make much sense

I do oppose some kind of underscore coding scheme (that's what librdkafka uses, which in hindsight was a bad idea) - it gives no hints on what they mean or how they differ.

At the end of the day I dont think it matters much, this stuff is only exposed to developers, not end users.

src/Confluent.Kafka/Structs/Error.cs
```diff
@@ -35,7 +35,12 @@ public string Message
         public bool HasError
             => Code != ErrorCode.NO_ERROR;
 
-        // TODO: questionably too tricky?
```
ewencp: Did we decide that the implicit bool conversion isn't too tricky?

mhowlett: I did, but we didn't. I'm happy either way on this one.

ewencp: That was the royal we. I'm fine with either option.

mhowlett:
Yep, constants should be PascalCase, good point. 

I notice in the Go client, you throw away indication of whether the enum value is local or broker in the enum name, but note it in the comment. I'm currently thinking this is best of all the not-ideal options we've come up with. In C# land, people always have good intellisense, so at least when they write code the two types will be clearly differentiated.

edenhill:
Yeah, I dont know if that was the right call.. :)

mhowlett:
we could also prepend the source to the string returned by err2str

edenhill:
Yep, err2str already does ("Local: ..", "Broker: ..")

mhowlett:
Perhaps this is a perfect place for Matt style mixed snake / pascal naming to come into play: 

```
Local_AllBrokersDown
Local_PartitionEOF
Broker_LeaderNotAvailable
```

I'm hesitant on such a visible part of the API, but this actually seems pretty pragmatic to me.

If not this, I'm in favor of dropping indication of local/broker from the enum name in line with the go client.

@edenhill @ewencp

edenhill:
Go with Matt_Casing

mhowlett:
Acronyms with three letters or more are Pascal case by convention:
https://msdn.microsoft.com/en-us/library/ms182249(VS.80).aspx

With two letter acronyms, both letters are capitalized by convention. But MS are inconsistent (see third answer down):
http://stackoverflow.com/questions/596062/net-naming-convention-for-id-anything-identification-capitalization

which makes me want to be inconsistent with `.PartitionEOF` because i don't like the look of `.PartitionEof`. This decision is sort of important because this should be consistent with `OnPartitionEOF` events on `Consumer`, and this is a commonly used part of the API.

I'm currently going with all caps for EOF. On the fence on this.

mhowlett:
Cool, microsoft are inconsistent here too:
https://msdn.microsoft.com/en-us/library/microsoft.visualbasic.filesystem.eof(v=vs.110).aspx

That pretty much locks my opinion in on this ..

src/Confluent.Kafka/Consumer.cs
```diff
@@ -659,7 +659,7 @@ public bool Consume(out Message message, int millisecondsTimeout)
                 {
                     case ErrorCode.NoError:
                         return true;
-                    case ErrorCode.Local_PartitionEof:
+                    case ErrorCode.Local_PartitionEOF:
```
edenhill: FWIW, Go client:
https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/generated_errors.go#L35

PR #20, opened 2017-01-26
fixes possible race condition in getKafkaTopicHandle

### LINKED ISSUE #18: getKafkaTopicHandle() not thread-safe
The topic handle cache isn't thread-safe as seen when trying to produce from multiple threads:
```
DEBUG: 2017-01-25 13:01:00: Produce on thread 7
DEBUG: 2017-01-25 13:01:00: Produce on thread 8

Unhandled Exception: System.ArgumentException: An item with the same key has already been added. Key: test.net
   at System.ThrowHelper.ThrowAddingDuplicateWithKeyArgumentException(Object key)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at Confluent.Kafka.Internal.SafeDictionary`2.Add(TKey key, TValue val)
   at Confluent.Kafka.Producer.getKafkaTopicHandle(String topic)
   at Confluent.Kafka.Producer.Produce(String topic, Byte[] val, Int32 valOffset, Int32 valLength, Byte[] key, Int32 keyOffset, Int32 keyLength, Nullable`1 timestamp, Int32 partition, Boolean blockIfQueueFull, IDeliveryHandler deliveryHandler)
   at Confluent.Kafka.Producer.ProduceAsync(String topic, Byte[] key, Byte[] val, IDeliveryHandler deliveryHandler, Nullable`1 timestamp, Nullable`1 partition, Boolean blockIfQueueFull)
   at Confluent.Kafka.SerializingProducer`2.ProduceAsync(String topic, TKey key, TValue val, Nullable`1 timestamp, Nullable`1 partition, Boolean blockIfQueueFull)
   at Confluent.Kafka.Producer`2.ProduceAsync(String topic, TKey key, TValue val, Nullable`1 timestamp, Nullable`1 partition, Boolean blockIfQueueFull)
   at Confluent.Kafka.VerifiableClient.VerifiableProducer.Produce(String topic, String value)
   at Confluent.Kafka.VerifiableClient.VerifiableProducer.TimedProduce(Object ignore)
   at Confluent.Kafka.VerifiableClient.VerifiableProducer.<Run>b__11_0(Object state)
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.TimerQueueTimer.CallCallback()
   at System.Threading.TimerQueueTimer.Fire()
   at System.Threading.TimerQueue.FireNextTimers()
```

The problem is most likely the unprotected read-write race in getKafkaTopicHandle() that first checks if topic is in the dict and then inserts it (if it isnt). That entire cycle needs to be done in one locked operation.

mhowlett:
fixes #18. also fixied up some topicConfigHandle related stuff.
@edenhill

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -424,12 +424,12 @@ namespace Confluent.Kafka
             LibRdKafka.conf_set_offset_commit_cb(configPtr, commitDelegate);
             if (defaultTopicConfig != null)
             {
-                var topicHandle = SafeTopicConfigHandle.Create();
+                var topicConfigHandle = SafeTopicConfigHandle.Create();
                 if (config != null)
                 {
-                    defaultTopicConfig.ToList().ForEach((kvp) => { topicHandle.Set(kvp.Key, kvp.Value.ToString()); });
+                    defaultTopicConfig.ToList().ForEach((kvp) => { topicConfigHandle.Set(kvp.Key, kvp.Value.ToString()); });
                 }
-                LibRdKafka.conf_set_default_topic_conf(configPtr, topicHandle.Dup());
+                LibRdKafka.conf_set_default_topic_conf(configPtr, topicConfigHandle.DangerousGetHandle());
             }
 
             LibRdKafka.conf_set_error_cb(configPtr, ErrorCallback);
--- a/src/Confluent.Kafka/Internal/SafeDictionary.cs
+++ b/src/Confluent.Kafka/Internal/SafeDictionary.cs
@@ -56,10 +56,13 @@ namespace Confluent.Kafka.Internal
                     throw new InvalidOperationException("Attempting to add an item to SafeDictionary which has been disposed.");
                 }
 
-                Dictionary<TKey, TValue> writeDictionary = new Dictionary<TKey, TValue>(readDictionary);
-                writeDictionary.Add(key, val);
-                // this is atomic.
-                readDictionary = writeDictionary;
+                if (!readDictionary.ContainsKey(key))
+                {
+                    Dictionary<TKey, TValue> writeDictionary = new Dictionary<TKey, TValue>(readDictionary);
+                    writeDictionary.Add(key, val);
+                    // this is atomic.
+                    readDictionary = writeDictionary;
+                }
             }
         }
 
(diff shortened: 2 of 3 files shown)


edenhill:
I'm not sure this fixes the problem, the read and write are still not protected by the same locking.

```
        private SafeTopicHandle getKafkaTopicHandle(string topic)
        {
            // TODO: We should consider getting rid of these and add proper support in librdkafka itself
            //       (producev() with RD_KAFKA_V_TOPIC() is one step closer)
```
1. Checking if topic is in the cache (no lock)
```
            if (topicHandles.ContainsKey(topic))
            {
                return topicHandles[topic];
            }

            var topicConfigHandle = SafeTopicConfigHandle.Create();
            if (topicConfig != null)
            {
                topicConfig
                    .ToList()
                    .ForEach((kvp) => { topicConfigHandle.Set(kvp.Key, kvp.Value.ToString()); });
            }
            topicConfigHandle.Set("produce.offset.report", "true");
            IntPtr configPtr = topicConfigHandle.DangerousGetHandle();

            // note: there is a possible (benign) race condition here - topicHandle could have already
            // been created for the topic (and possibly added to topicHandles). If the topicHandle has
            // already been created, rdkafka will return it and not create another. the call to rdkafka
            // is threadsafe.
            var topicHandle = kafkaHandle.Topic(topic, configPtr);
```
2. Add `new rd_kafka_topic_t *` object to cache
```

            topicHandles.Add(topic, topicHandle);

            return topicHandle;
        }
```

in topicHandles.Add:
```
       public void Add(TKey key, TValue val)
        {
            lock (writeLockObj)
            {
                if (readDictionary == null)
                {
                    throw new InvalidOperationException("Attempting to add an item to SafeDictionary which has been disposed.");
                }
```
3. Check again if topic is in the cache (now with lock)
```
                if (!readDictionary.ContainsKey(key))
                {
                    Dictionary<TKey, TValue> writeDictionary = new Dictionary<TKey, TValue>(readDictionary);
```
5. Only add if not existent
```
                    writeDictionary.Add(key, val);
                    // this is atomic.
                    readDictionary = writeDictionary;
                }
            }
        }
```

So it seems the race condition is still there if two or more threads call getKafkaTopicHandle() with the same topic then multiple topic objects will be created and only one of them will be added to the cache.
The proper fix would be to lock the entire 'read-check-create-add' cycle with one single lock taking.
Otherwise there will be multiple topic objects for the same topic and I'm not sure if the ones that are not added to the cache will be cleaned up (every topic_new() must have a corresponding topic_destroy()).

mhowlett:
> every topic_new() must have a corresponding topic_destroy()
ok, i understood that incorrectly then.
at least the current code won't throw an exception (it'll just leak memory occasionally).
will re-open issue.

edenhill:
:+1:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L1626

Issue #22, opened 2017-01-26
StringDeserializer uses hard coded encoding


bschmidtbauer:
The built-in StringDeserializer doesn't use the constructor specified encoding but uses UTF8.  The StringSerializer uses the specified encoding.

https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/src/Confluent.Kafka/Serialization/StringDeserializer.cs#L35

mhowlett:
whoops! good catch, thanks.

PR #23, opened 2017-01-27
fixes issue #22

mhowlett:
@ewencp @edenhill

### DIFF
--- a/test/Confluent.Kafka.UnitTests/Serialization/String.cs
+++ b/test/Confluent.Kafka.UnitTests/Serialization/String.cs
@@ -1,0 +1,36 @@
+// Copyright 2016-2017 Confluent Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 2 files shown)


test/Confluent.Kafka.UnitTests/Serialization/String.cs
```diff
@@ -0,0 +1,36 @@
...
+    public class StringTests
+    {
+        [Fact]
+        public void SerializeDeserialize()
+        {
+            Assert.Equal("hello world", new StringDeserializer(Encoding.UTF8).Deserialize(new StringSerializer(Encoding.UTF8).Serialize("hello world")));
+            Assert.Equal("ឆ្មាត្រូវបានហែលទឹក", new StringDeserializer(Encoding.UTF8).Deserialize(new StringSerializer(Encoding.UTF8).Serialize("ឆ្មាត្រូវបានហែលទឹក")));
+            Assert.Equal("вы не банан", new StringDeserializer(Encoding.UTF8).Deserialize(new StringSerializer(Encoding.UTF8).Serialize("вы не банан")));
+
+            Assert.NotEqual("ឆ្មាត្រូវបានហែលទឹក", new StringDeserializer(Encoding.UTF8).Deserialize(new StringSerializer(Encoding.UTF32).Serialize("ឆ្មាត្រូវបានហែលទឹក")));
```
ewencp: I suppose these are simply constructed such that decoding exceptions won't happen during the tests? It's not necessarily obvious that is the case. More generally, as this is being cleaned up, the potential for `ArgumentException` and `DecoderFallbackException` might be worth handling.

mhowlett: i was (errantly) assuming Deserialize never threw. Will remove the error case tests and replace with a TODO. Agree it's worth having these tests but it'll take a bit of time to construct them well and ROI is low i think.

Issue #25, opened 2017-01-28
OnPartitionsAssigned should get TopicPartitionOffset


edenhill:
Since onPartitionsAssigned MUST call Assign(), and Assign() takes a TopicPartitionOffset collection, I think it would make more sense to actually provide that as the argument to the delegate instead of poorer TopicPartition.
Otherwise all applications needs to perform the TopicPartition -> TopicPartitionOffset collection dance.

This also avoids any compatibility problems if librdkafka starts providing another offset in the rebalance_cb than OFFSET_INVALID (which the Consumer.cs code now throws away and then re-inserts if no OnPartitionsAssigned delegate is set).

```
Handle.OnPartitionsAssigned += (_, partitions) =>
             {
                 Console.WriteLine($"Assigned partitions: [{string.Join(", ", partitions)}], member id: {Handle.MemberId}");
                 Handle.Assign(partitions.Select(p => new TopicPartitionOffset(p, Offset.Invalid)));
             };
```

mhowlett:
Is librdkafka likely to start working in that way? that may have implications to more places than just here.

A neat solution would be to provide an overload of Assign that takes a TopicPartition - such a method would work as people anticipate. Side note: something I don't like about the current implementation is the need to specify an enum value with name 'Invalid'.

I have a big problem with OnPartitionsAssigned providing an Offset, when this is always going to be -1001. The thought process of most people, I think, will be to see that there is an offset and expect that this will typically provide an actual offset. They will see -1001, not realize that this is what is always returned and think that in their particular case there has been an error. They will then go off and try and work out what's wrong. Furthermore, I think they typically won't even go to the relevant part of the manual looking for the answer because it will be so counter intuitive to them that the API should work this way - they will assume something else is wrong. In my very limited experience talking to users of librdkafka based libraries, i've seen this confusion twice. And people get stuck on it - but it's easily avoided.

edenhill:
Okay, your suggestion of Assign(TopicPartition) sounds good.

PR #29, opened 2017-01-31
new Consumer.Assign overload

mhowlett:
@edenhill @ewencp 
Addresses issue #25

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -197,14 +197,14 @@ namespace Confluent.Kafka
         ///     Update the assignment set to <param name="partitions" />.
         ///
         ///     The assignment set is the complete set of partitions to consume
-        ///     from.
+        ///     from and will replace any previous assignment.
         /// </summary>
         /// <param name="partitions">
         ///     The set of partitions to consume from. If an offset value of
         ///     -1001 (Offset.Invalid) is specified for a partition, consumption
-        ///     will resume from one after the last committed offset on that
-        ///     partition, or according to the 'auto.offset.reset' configuration
-        ///     parameter if no offsets have been committed yet.
+        ///     will resume from the last committed offset on that partition, or
+        ///     according to the 'auto.offset.reset' configuration parameter if
+        ///     no offsets have been committed yet.
         /// </param>
         public void Assign(IEnumerable<TopicPartitionOffset> partitions)
             => consumer.Assign(partitions);
@@ -213,13 +213,13 @@ namespace Confluent.Kafka
         ///     Update the assignment set to <param name="partitions" />.
         ///
         ///     The assignment set is the complete set of partitions to consume
-        ///     from.
+        ///     from and will replace any previous assignment.
         /// </summary>
         /// <param name="partitions">
-        ///     The set of partitions to consume from. Consumption
-        ///     will resume from one after the last committed offset on each
-        ///     partition, or according to the 'auto.offset.reset' configuration
-        ///     parameter if no offsets have been committed yet.
+        ///     The set of partitions to consume from. Consumption will resume
+        ///     from the last committed offset on each partition, or according
+        ///     to the 'auto.offset.reset' configuration parameter if no offsets
+        ///     have been committed yet.
         /// </param>
         public void Assign(IEnumerable<TopicPartition> partitions)
             => consumer.Assign(partitions);
@@ -579,14 +579,14 @@ namespace Confluent.Kafka
         ///     Update the assignment set to <param name="partitions" />.
         ///
         ///     The assignment set is the complete set of partitions to consume
-        ///     from.
+        ///     from and will replace any previous assignment.
         /// </summary>
         /// <param name="partitions">
         ///     The set of partitions to consume from. If an offset value of
         ///     -1001 (Offset.Invalid) is specified for a partition, consumption
-        ///     will resume from one after the last committed offset on that
-        ///     partition, or according to the 'auto.offset.reset' configuration
-        ///     parameter if no offsets have been committed yet.
+        ///     will resume from the last committed offset on that partition, or
+        ///     according to the 'auto.offset.reset' configuration parameter
+        ///     if no offsets have been committed yet.
         /// </param>
         public void Assign(IEnumerable<TopicPartitionOffset> partitions)
             => kafkaHandle.Assign(partitions.ToList());
@@ -595,13 +595,13 @@ namespace Confluent.Kafka
         ///     Update the assignment set to <param name="partitions" />.
         ///
         ///     The assignment set is the complete set of partitions to consume
-        ///     from.
+        ///     from and will replace any previous assignment.
         /// </summary>
         /// <param name="partitions">
-        ///     The set of partitions to consume from. Consumption
-        ///     will resume from one after the last committed offset on each
-        ///     partition, or according to the 'auto.offset.reset' configuration
-        ///     parameter if no offsets have been committed yet.
+        ///     The set of partitions to consume from. Consumption will resume
+        ///     from the last committed offset on each partition, or according
+        ///     to the 'auto.offset.reset' configuration parameter if no offsets
+        ///     have been committed yet.
         /// </param>
         public void Assign(IEnumerable<TopicPartition> partitions)
             => kafkaHandle.Assign(partitions.Select(p => new TopicPartitionOffset(p, Offset.Invalid)).ToList());
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -201,7 +201,7 @@ namespace Confluent.Kafka
         /// </summary>
         /// <param name="partitions">
         ///     The set of partitions to consume from. If an offset value of
-        ///     -1001 (Offset.Invalid) is specified for a partition, consumption
+        ///     Offset.Invalid (-1001) is specified for a partition, consumption
         ///     will resume from the last committed offset on that partition, or
         ///     according to the 'auto.offset.reset' configuration parameter if
         ///     no offsets have been committed yet.
@@ -583,7 +583,7 @@ namespace Confluent.Kafka
         /// </summary>
         /// <param name="partitions">
         ///     The set of partitions to consume from. If an offset value of
-        ///     -1001 (Offset.Invalid) is specified for a partition, consumption
+        ///     Offset.Invalid (-1001) is specified for a partition, consumption
         ///     will resume from the last committed offset on that partition, or
         ///     according to the 'auto.offset.reset' configuration parameter
         ///     if no offsets have been committed yet.
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -194,15 +194,34 @@ namespace Confluent.Kafka
             => consumer.Unsubscribe();
 
         /// <summary>
-        ///     Update the assignment set to \p partitions.
+        ///     Update the assignment set to <param name="partitions" />.
         ///
-        ///     The assignment set is the set of partitions actually being consumed
-        ///     by the KafkaConsumer.
+        ///     The assignment set is the complete set of partitions to consume
+        ///     from.
         /// </summary>
-        /// <remarks>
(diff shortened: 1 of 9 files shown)


src/Confluent.Kafka/Consumer.cs
```diff
@@ -194,18 +194,37 @@ public void Unsubscribe()
...
-        /// </summary>
-        /// <remarks>
-        ///     blocks
-        /// </remarks>
+        ///     The assignment set is the complete set of partitions to consume
+        ///     from.
+        /// </summary>
+        /// <param name="partitions">
+        ///     The set of partitions to consume from. If an offset value of
+        ///     -1001 (Offset.Invalid) is specified for a partition, consumption
```
mhowlett: I'm wondering if Offset.Unset would be a better name than Offset.Invalid - Invalid is not a great name here. I guess the other main place this is used is in delivery reports. There, I think Invalid is better but Unset is ok. thoughts on changing this? one negative is it would be a departure from librdkafka (though i don't see that as a deal breaker). alternatively we could provide both .Invalid and .Unset but i think that's messy and potentially confusing as people would naturally assume they're different. Or the name could be .InvalidOrUnset. Or we leave as is.

edenhill: Could we use a Nullable field?

src/Confluent.Kafka/Consumer.cs
```diff
@@ -194,18 +194,37 @@ public void Unsubscribe()
...
+        ///     will resume from one after the last committed offset on that
```
edenhill: It will start at the committed offset (the committed offset is the next offset to consume, not the last message consumed)

src/Confluent.Kafka/Consumer.cs
```diff
@@ -557,19 +576,38 @@ public void Unsubscribe()
...
+        ///     Update the assignment set to <param name="partitions" />.
+        ///
+        ///     The assignment set is the complete set of partitions to consume
+        ///     from.
+        /// </summary>
+        /// <param name="partitions">
+        ///     The set of partitions to consume from. Consumption
+        ///     will resume from one after the last committed offset on each
+        ///     partition, or according to the 'auto.offset.reset' configuration
+        ///     parameter if no offsets have been committed yet.
```
edenhill: dito

src/Confluent.Kafka/Consumer.cs
```diff
@@ -557,19 +576,38 @@ public void Unsubscribe()
...
-        /// <remarks>
-        ///     blocks
-        /// </remarks>
+        ///     The assignment set is the complete set of partitions to consume
+        ///     from.
+        /// </summary>
+        /// <param name="partitions">
+        ///     The set of partitions to consume from. If an offset value of
+        ///     -1001 (Offset.Invalid) is specified for a partition, consumption
+        ///     will resume from one after the last committed offset on that
```
edenhill: dito

test/Confluent.Kafka.IntegrationTests/Tests/AssignOverloads.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        /// <summary>
+        ///     Simple test of both Consumer.Assign overloads.
+        /// </summary>
+        [Theory, MemberData(nameof(KafkaParameters))]
+        public static void AssignOverloads(string bootstrapServers, string topic)
+        {
+            var consumerConfig = new Dictionary<string, object>
+            {
+                { "group.id", "u-bute-group" },
```
edenhill: what's a u-bute ?

test/Confluent.Kafka.IntegrationTests/Tests/AssignPastEnd.cs
```diff
@@ -23,12 +23,12 @@
...
-    /// <summary>
-    ///     Test functionality of Consumer.Consume when assigned to offest
-    ///     higher than the offset of the last message on a partition.
-    /// </summary>
     public static partial class Tests
     {
+        /// <summary>
+        ///     Test functionality of Consumer.Consume when assigned to offest
+        ///     higher than the offset of the last message on a partition.
+        /// </summary>
```
edenhill: This should set topic auto.offset.reset=end to make it explicit what kind of behaviour is expected.

test/Confluent.Kafka.IntegrationTests/Tests/AssignOverloads.cs
mhowlett: Australian slang. bute is short for beautiful. u-bute is sort of the opposite. kind of means slapped together quickly, a bit of a hack. I'll change this, it doesn't belong in integration tests.

src/Confluent.Kafka/Consumer.cs
mhowlett: Nullable is not as elegant/simple:

1. We already have (and need to continue to use) some other mechanism for specifying special values. Using Nullable here introduces a separate mechanism.
2. It doubles memory requirements per offset.
3. It brings the GC into play.
4. There is no built in language construct in F# for Nullable.

The costs outweigh the benefit IMO.

Generally speaking, Nullable is loosing favor with me as something to be used in a public interface. Also, it's not used much in the .NET Framework BCLs and only really for peripheral things.

test/Confluent.Kafka.IntegrationTests/Tests/AssignOverloads.cs
ewencp: So disappointed it wasn't a reference to http://www.ultimatebbqs.com.au/products-page/accessories/u-bute-portable-butane-power-bbq/

src/Confluent.Kafka/Consumer.cs
ewencp: Point 4 seems like the deal breaker if it makes things a lot messier in languages that don't handle nullable.

ewencp: Also, a real optional class seems like the ideal here but since there isn't one in the C# stdlib...

test/Confluent.Kafka.IntegrationTests/Tests/AssignOverloads.cs
mhowlett: not sure I'd trust any u-bute butane powered thing..

src/Confluent.Kafka/Consumer.cs
```diff
@@ -194,18 +194,37 @@ public void Unsubscribe()
...
+        ///     Update the assignment set to <param name="partitions" />.
         ///
-        ///     The assignment set is the set of partitions actually being consumed
-        ///     by the KafkaConsumer.
-        /// </summary>
-        /// <remarks>
-        ///     blocks
-        /// </remarks>
+        ///     The assignment set is the complete set of partitions to consume
+        ///     from.
```
edenhill: ".. replacing the previous assignment" or something to that effect

src/Confluent.Kafka/Consumer.cs
edenhill: dito

src/Confluent.Kafka/Consumer.cs
```diff
@@ -557,19 +576,38 @@ public void Unsubscribe()
...
         /// <summary>
-        ///     Stop consumption and remove the current assignment.
+        ///     Update the assignment set to <param name="partitions" />.
+        ///
+        ///     The assignment set is the complete set of partitions to consume
+        ///     from.
+        /// </summary>
+        /// <param name="partitions">
+        ///     The set of partitions to consume from. Consumption
+        ///     will resume from one after the last committed offset on each
```
edenhill: dito

test/Confluent.Kafka.IntegrationTests/Tests/AssignPastEnd.cs
mhowlett: this example doesn't use kafka offset management at all - it just calls Assign on a particular partition with an offset based on the last produced message, so the auto.offset.reset property should be irrelevant.. 

However, now you've got me looking at this... the test should check the produced message offset it not invalid. Also I noticed the Offset property in Message is still of type long not Offset and I'll change that too.

src/Confluent.Kafka/Consumer.cs
mhowlett: what about '.. and will replace any previous assignment'.

test/Confluent.Kafka.IntegrationTests/Tests/AssignPastEnd.cs
edenhill: Yes, but if you assign an invalid offset then auto.offset.reset will kick in - it defaults to end, that's why you're not seeing any messages which is expected by the test, but coming in and seeing this test it might seem that it wont consume if you give it an invalid offset, which isnt true. So make it more explicit what is actually going on by setting that config property, or making a comment.

mhowlett: i tried setting auto.offset.reset to smallest and it does not change the behavior of the test.

note that by 'invalid offset' i mean 'out of range offset', not -1001.

it seems counter intuitive to me that auto.offset.reset would come into play for an explicitly specified offset, even if it's out of range.

edenhill: Not really, the application needs to decide what should happen if a sought offset doesn't exist. The client helps with this by allowing the application to select the three possible alternatives: start from beginning, start from end, error out

edenhill: The partition end offset (lastMsg+1) is actually considered a valid offset (broker wont return OffsetOutOfRange),  so you should probably set it to +1000 or something

mhowlett: ahh ok, i see that error is an option for auto.offset.reset now and also see why this is an intuitive way for this to work.

note that lastMsg+2 is the first offset to be considered out of range (makes sense), and given your other comments in this PR about what offset is, of course lastMsg+1 will be in range (I didn't properly get that before).

test/Confluent.Kafka.IntegrationTests/Tests/AssignPastEnd.cs
```diff
@@ -46,14 +46,19 @@ public static void AssignPastEnd(string bootstrapServers, string topic)
             using (var producer = new Producer<Null, string>(producerConfig, null, new StringSerializer(Encoding.UTF8)))
             {
                 dr = producer.ProduceAsync(topic, null, testString).Result;
+                Assert.NotEqual(dr.Offset, Offset.Invalid);
```
ewencp: `>= 0` might be better here since it excludes a larger set of values we don't want to see

src/Confluent.Kafka/Consumer.cs
mhowlett: My current thoughts are to leave this as is because people are not likely to want to specify this directly here now there is the overload of Assign that takes `TopicPartition`s. I may well flip-flop on that opinion, but I think this PR is a step forward so we should merge it.

test/Confluent.Kafka.IntegrationTests/Tests/AssignPastEnd.cs
```diff
@@ -63,8 +68,19 @@ public static void AssignPastEnd(string bootstrapServers, string topic)
...
                 consumer.Poll(TimeSpan.FromSeconds(10));
+                consumer.Assign(new List<TopicPartitionOffset>() { new TopicPartitionOffset(dr.TopicPartition, dr.Offset+2) });
+                consumer.Poll(TimeSpan.FromSeconds(10));
             }
 
+            consumerConfig["default.topic.config"] = new Dictionary<string, object>() { { "auto.offset.reset", "earliest" } };
+            using (var consumer = new Consumer(consumerConfig))
+            {
+                Message msg;
+                consumer.Assign(new List<TopicPartitionOffset>() { new TopicPartitionOffset(dr.TopicPartition, dr.Offset+1) });
```
ewencp: Given there was some confusion over what the behavior should be depending on lastOffset+1 vs lastOffset+2, you might want to add some comments explaining the expected behavior (inline w/ the asserts or the to comment on the test.

Issue #16, opened 2017-01-23
Inappropriate exception in Consumer constructor


mhowlett:
When configuration parameters meant for a Producer are passed to the Consumer constructor, the following exception is thrown:

"Failed to redirect the poll queue to consumer_poll queue"

An exception should be thrown, but it should provide a more appropriate error message.

mhowlett:
Notes / Questions:

1. The error returned by rd_kafka_poll_set_consumer is _UNKNOWN_GROUP. Why is a group.id required by rd_kafka_poll_set_consumer?
2. a call to rd_kafka_new (type Consumer) is successful with no group.id specified, but the resulting client is not really very useful. It seems it can be used for some things, for example querying the watermark offsets, but can't be used to consume anything.
3. The way things are then, I think I should add a check for the existence of "group.id" in the Consumer constructor and throw an appropriate exception if it doesn't exist.
4. but this seems wrong - why is a group.id a prerequisite for calling .Assign and reading messages from topics? Calling .Assign outside of .Subscribe does not create a consumer group as far as i can tell. And you can read messages having called it with no consumer group ever created. The way it seems this should work to me is for a group.id to not be required, and an exception thrown only if the user of the consumer tries to call .Subscribe (i.e. use group functionality).

ewencp:
@edenhill may have different reasoning, but I actually prefer requiring the group ID (which the Java implementation doesn't require). Not doing so has multiple paths that can cause problems. Anything with group membership or offsets require a group ID. That means offset commit would also fail. And deciding what offset to start at would differ between the two modes (one checks for committed offsets first, the other doesn't). I think the confusing part is that it's not fail fast and leaves some functionality working (e.g. presumably all the generic stuff that's available on both producers and consumers).

mhowlett:
First, I think it's pretty clear here given all the constraints that we should fail fast on no group.id and throw an exception with a good error. 

But... my preference for not having to specify group.id is strong. If I have to specify a group.id in scenarios where it's not needed I might be left wondering if there is something about those scenarios that I don't understand properly. If the requirements of the API match up with the actual operation, I'm left with more certainty - a better mental model more easily - of how things are working when I'm fiddling about trying to work things out. Kafka is complex, and I think this sort of thing is important to helping people grok it. 

[...]

PR #30, opened 2017-02-01
new Consumer() fail fast when no group.id

mhowlett:
Addresses issue #16

### DIFF
--- a/examples/AdvancedConsumer/Program.cs
+++ b/examples/AdvancedConsumer/Program.cs
@@ -250,7 +250,7 @@ namespace Confluent.Kafka.Examples.AdvancedConsumer
 
         public static void Main(string[] args)
         {
-            if (args.Length != 3)
+            if (args.Length < 3)
             {
                 PrintUsage();
                 return;
--- a/examples/AdvancedConsumer/Program.cs
+++ b/examples/AdvancedConsumer/Program.cs
@@ -245,21 +245,34 @@ namespace Confluent.Kafka.Examples.AdvancedConsumer
             }
         }
 
+        private static void PrintUsage()
+            => Console.WriteLine("usage: <poll|consume|background> <broker,broker,..> <topic> [topic..]");
+
         public static void Main(string[] args)
         {
-            var mode = args[1];
+            if (args.Length != 3)
+            {
+                PrintUsage();
+                return;
+            }
+
+            var mode = args[0];
+            var brokerList = args[1];
             var topics = args.Skip(2).ToList();
 
-            switch (args[0])
+            switch (mode)
             {
                 case "poll":
-                    Run_Poll(mode, topics);
+                    Run_Poll(brokerList, topics);
                     break;
                 case "consume":
-                    Run_Consume(mode, topics);
+                    Run_Consume(brokerList, topics);
                     break;
                 case "background":
-                    Run_Background(mode, topics);
+                    Run_Background(brokerList, topics);
+                    break;
+                default:
+                    PrintUsage();
                     break;
             }
         }
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -408,6 +408,11 @@ namespace Confluent.Kafka
         /// </remarks>
         public Consumer(IEnumerable<KeyValuePair<string, object>> config)
         {
+            if (config.FirstOrDefault(prop => prop.Key == "group.id").Value == null)
+            {
+                throw new Exception("'group.id' configuration parameter is required and was not specified.");
+            }
+
             var defaultTopicConfig = (IEnumerable<KeyValuePair<string, object>>)config.FirstOrDefault(prop => prop.Key == "default.topic.config").Value;
             var configHandle = SafeConfigHandle.Create();
             config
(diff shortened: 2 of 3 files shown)


examples/AdvancedConsumer/Program.cs
```diff
@@ -245,21 +245,34 @@ public static void Run_Consume(string brokerList, List<string> topics)
...
+
+            var mode = args[0];
+            var brokerList = args[1];
             var topics = args.Skip(2).ToList();
 
-            switch (args[0])
+            switch (mode)
             {
                 case "poll":
-                    Run_Poll(mode, topics);
```
mhowlett: I have no idea how this nonsense made it in

src/Confluent.Kafka/Consumer.cs
```diff
@@ -408,6 +408,11 @@ private void LogCallback(IntPtr rk, int level, string fac, string buf)
         /// </remarks>
         public Consumer(IEnumerable<KeyValuePair<string, object>> config)
         {
+            if (config.FirstOrDefault(prop => prop.Key == "group.id").Value == null)
```
tomasdeml: I think these configuration keys should be compared using `string.Equals(x, y, StringComparison.Ordinal)` as recommended [here](https://msdn.microsoft.com/en-US/library/dd465121(v=vs.110).aspx#choosing_a_stringcomparison_member_for_your_method_call).

src/Confluent.Kafka/Consumer.cs
```diff
@@ -408,6 +408,11 @@ private void LogCallback(IntPtr rk, int level, string fac, string buf)
...
+            {
+                throw new Exception("'group.id' configuration parameter is required and was not specified.");
```
tomasdeml: IMO the exception should be more specific, e.g. `ArgumentException`.

mhowlett: yep, I agree.

src/Confluent.Kafka/Consumer.cs
mhowlett: Thanks for the review Tomáš! I wasn't familiar with those recommendations. Yep, I agree it's better to follow convention here.

examples/AdvancedConsumer/Program.cs
```diff
@@ -245,21 +245,34 @@ public static void Run_Consume(string brokerList, List<string> topics)
             }
         }
 
+        private static void PrintUsage()
+            => Console.WriteLine("usage: <poll|consume|background> <broker,broker,..> <topic> [topic..]");
+
         public static void Main(string[] args)
         {
-            var mode = args[1];
+            if (args.Length != 3)
```
edenhill: shouldn't this be < 3 since it takes a list of topics?

ewencp: Shouldn't be `!=` or multiple topics won't work.

test/Confluent.Kafka.UnitTests/Consumer.cs
```diff
@@ -0,0 +1,38 @@
...
+    public class ConsumerTests
+    {
+        [Fact]
+        public void Constuctor()
+        {
+            var config = new Dictionary<string, object>();
+            var e = Assert.Throws<ArgumentException>(() => { var c = new Consumer(config); });
+            Assert.True(e.Message.Contains("group.id"));
+            e = Assert.Throws<ArgumentException>(() => { var c = new Consumer<Null, string>(config, null, new StringDeserializer(Encoding.UTF8)); });
+            Assert.True(e.Message.Contains("group.id"));
```
edenhill: add a positive test too where group.id is set.
Also add a note what is being tested since that won't be clear for future readers.

examples/AdvancedConsumer/Program.cs
mhowlett: yep.

test/Confluent.Kafka.UnitTests/Consumer.cs
mhowlett: I didn't do any positive tests because 1. I didn't want to call into rd_kafka_new in the unit tests which don't require a broker to be running (though now I checked, I see that's fine even without specifying any config param other than group.id) and 2. every integration tests this case already. I want to leave that as is.

But I'll put in some comments.

edenhill: That's okay, as long as this comment + the code comment is shorter than the test code I asked about ;)

mhowlett: i didn't even mention THAT. ok, i'll amend what I just pushed.

test/Confluent.Kafka.UnitTests/Consumer.cs
```diff
@@ -33,6 +39,8 @@ public void Constuctor()
             Assert.True(e.Message.Contains("group.id"));
             e = Assert.Throws<ArgumentException>(() => { var c = new Consumer<Null, string>(config, null, new StringDeserializer(Encoding.UTF8)); });
             Assert.True(e.Message.Contains("group.id"));
+
+            // positve case covered by integration tests. here, avoiding creating a rd_kafka_t instance.
```
edenhill: I would argue that this is a typical unit-test and that we shouldn't rely on integration tests in this case - they are not run as frequently, and not from CI.

Apart from that this is LGTM.

Your call on the unit test.

Issue #27, opened 2017-01-29
dotnet quit unexpectedly


mhowlett:
I'm using Confluent.Kafka in an application in which I create a (non serializing) producer and then do nothing more with it. The application does a lot of processing and at semi-random points in time (doesn't take too long ~ 20 seconds) exits with "dotnet quit unexpectedly". This is repeatable. The problem is (repeatably) resolved if I don't create the producer, so I'm sure Confluent.Kafka is at fault.

One time (but only one time, most of the time not), an error including the following was printed to the console: "pointer being freed was not allocated"

Maybe this is something to do with the pointer to the kafka config being accessed after being garbage collected. I swapped out .Dup with .DangerousGetHandle() when setting the config. Maybe the problem is related to that.

We could write a test for this - create a producer / consumer then do lots of arbitrary processing to give the GC a work out.

mhowlett:
switching .DangerousGetHandle() back to .Dup() fixes this, but I haven't yet put any effort into understanding why.

edenhill:
Not sure it has bearing on this, but the *_conf_t objects passed to rd_kafka_new() and rd_kafka_topic_new() are freed by librdkafka and the application mustn't touch them after a successful such _new call

bschmidtbauer:
FWIW - When hosting a consumer or producer in a console I get an access violation when exiting the console.

Unhandled Exception: System.AccessViolationException: Attempted to read or write protected memory. This is often an indication that other memory is corrupt.
   at Confluent.Kafka.Impl.LibRdKafka.NativeMethods.rd_kafka_conf_destroy(IntPtr conf)
   at Confluent.Kafka.Impl.LibRdKafka.conf_destroy(IntPtr conf)
   at Confluent.Kafka.Impl.SafeConfigHandle.ReleaseHandle()
   at System.Runtime.InteropServices.SafeHandle.InternalFinalize()
   at System.Runtime.InteropServices.SafeHandle.Dispose(Boolean disposing)
   at System.Runtime.InteropServices.SafeHandle.Finalize()

The code is effectively:
```
        static void Main(string[] args)
        {
            var config = new Dictionary<string, object>
            {
               // config here
            };
            using (var consumer = new Confluent.Kafka.Consumer(config))
            {
                PrintConsumerGroups(consumer);
            }

            Console.WriteLine("Press Enter to exit");
            Console.ReadLine();
        }
```
[...]

PR #35, opened 2017-02-02
no longer release config handles

mhowlett:
Fixes issue #27.

I have open questions still. Mostly, is there really any value in using SafeHandle to wrap unmanaged config pointers?

1. we don't want ownership semantics (a major service they provide). Also, the resource referenced is just memory allocated by librdkafka (not a win32 handle or anything like that). Does that make all of the advantages noted in https://msdn.microsoft.com/en-us/library/system.runtime.interopservices.safehandle(v=vs.110).aspx irrelevant? 
2. we're grabbing the associated IntPtr and passing it into many of the librdkafka methods anyway (should we change the marshaling of methods like conf_set_error_cb take SafeHandle, not IntPtr ?)

I propose getting this fix in now and deferring careful consideration of the above and related questions to another Issue. At this time, I want to prioritize getting the API locked down.

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -430,6 +430,7 @@ namespace Confluent.Kafka
                     defaultTopicConfig.ToList().ForEach((kvp) => { topicConfigHandle.Set(kvp.Key, kvp.Value.ToString()); });
                 }
                 LibRdKafka.conf_set_default_topic_conf(configPtr, topicConfigHandle.DangerousGetHandle());
+                topicConfigHandle.SetHandleAsInvalid(); // topic config object is no longer useable.
             }
 
             LibRdKafka.conf_set_error_cb(configPtr, ErrorCallback);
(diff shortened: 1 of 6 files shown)


src/Confluent.Kafka/Consumer.cs
```diff
@@ -430,13 +430,15 @@ public Consumer(IEnumerable<KeyValuePair<string, object>> config)
...
                 LibRdKafka.conf_set_default_topic_conf(configPtr, topicConfigHandle.DangerousGetHandle());
+                topicConfigHandle.SetHandleAsInvalid(); // topic config object is no longer useable.
             }
 
             LibRdKafka.conf_set_error_cb(configPtr, ErrorCallback);
             LibRdKafka.conf_set_log_cb(configPtr, LogCallback);
             LibRdKafka.conf_set_stats_cb(configPtr, StatsCallback);
 
             this.kafkaHandle = SafeKafkaHandle.Create(RdKafkaType.Consumer, configPtr);
+            configHandle.SetHandleAsInvalid(); // config object is no longer useable.
```
edenhill: The passed conf_t is only invalidated if rd_kafka_new() succeeded, if it returns error the caller will need to destroy it.

mhowlett: by this point rd_kafka_new is actually guaranteed to have succeeded as SafeKafkaHandle.Create throws if not... but SafeKafkaHandle.Create does not destroy it either in the case of failure (so I'll do that there).

edenhill:
LGTM

PR #38, opened 2017-02-03
Kafkatest client [WIP]

edenhill:
And various fixes for problems found implementing the kafkatest client.

Currently only includes the VerifiableProducer, the consumer part will be added later to this PR.
The premature push is to allow review of core client changes, rather than the **VerifiableProducer (so ignore that part for now).**

### DIFF
--- a/examples/AdvancedConsumer/Program.cs
+++ b/examples/AdvancedConsumer/Program.cs
@@ -66,7 +66,7 @@ namespace Confluent.Kafka.Examples.AdvancedConsumer
                 consumer.OnError += (_, error)
                     => Console.WriteLine($"Error: {error}");
 
-                consumer.OnOffsetCommit += (_, commit) =>
+                consumer.OnOffsetsCommitted += (_, commit) =>
                 {
                     Console.WriteLine($"[{string.Join(", ", commit.Offsets)}]");
 
@@ -126,7 +126,7 @@ namespace Confluent.Kafka.Examples.AdvancedConsumer
                 consumer.OnError += (_, error)
                     => Console.WriteLine($"Error: {error}");
 
-                consumer.OnOffsetCommit += (_, commit) =>
+                consumer.OnOffsetsCommitted += (_, commit) =>
                 {
                     Console.WriteLine($"[{string.Join(", ", commit.Offsets)}]");
 
@@ -189,7 +189,7 @@ namespace Confluent.Kafka.Examples.AdvancedConsumer
                 consumer.OnError += (_, error)
                     => Console.WriteLine($"Error: {error}");
 
-                consumer.OnOffsetCommit += (_, commit) =>
+                consumer.OnOffsetsCommitted += (_, commit) =>
                 {
                     Console.WriteLine($"[{string.Join(", ", commit.Offsets)}]");
 
--- a/examples/AdvancedConsumer/Program.cs
+++ b/examples/AdvancedConsumer/Program.cs
@@ -238,8 +238,8 @@ namespace Confluent.Kafka.Examples.AdvancedConsumer
                     if (msg.Offset % 5 == 0)
                     {
                         Console.WriteLine($"Committing offset");
-                        consumer.Commit(msg);
-                        Console.WriteLine($"Committed offset");
+                        var committedOffsets = consumer.CommitAsync(msg).Result;
+                        Console.WriteLine($"Committed offset: {committedOffsets}");
                     }
                 }
             }
--- a/examples/Misc/Program.cs
+++ b/examples/Misc/Program.cs
@@ -57,7 +57,7 @@ namespace Confluent.Kafka.Examples.Misc
             var config = new Dictionary<string, object> { { "bootstrap.servers", brokerList } };
             using (var producer = new Producer(config))
             {
-                var meta = producer.GetMetadata();
+                var meta = producer.GetMetadata(true, null);
                 Console.WriteLine($"{meta.OriginatingBrokerId} {meta.OriginatingBrokerName}");
                 meta.Brokers.ForEach(broker =>
                     Console.WriteLine($"Broker: {broker.BrokerId} {broker.Host}:{broker.Port}"));
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -312,13 +312,13 @@ namespace Confluent.Kafka
         public string MemberId
             => consumer.MemberId;
 
-        /// <summary>
+        /// <returns>
         ///     The current librdkafka out queue length.
-        /// </summary>
+        /// </returns>
         /// <remarks>
-        ///     The out queue contains messages and requests waiting to be sent to,
+        ///     The out queue contains requests waiting to be sent,
         ///     or acknowledged by, the broker.
-        /// </summary>
+        /// </remarks>
         public long OutQueueLength
             => consumer.OutQueueLength;
 
@@ -817,13 +817,13 @@ namespace Confluent.Kafka
         public string MemberId
             => kafkaHandle.MemberId;
 
-        /// <summary>
+        /// <returns>
         ///     The current librdkafka out queue length.
-        /// </summary>
+        /// </returns>
         /// <remarks>
-        ///     The out queue contains messages and requests waiting to be sent to,
+        ///     The out queue contains requests waiting to be sent,
         ///     or acknowledged by, the broker.
-        /// </summary>
+        /// </remarks>
         public long OutQueueLength
             => kafkaHandle.OutQueueLength;
 
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -142,10 +142,10 @@ namespace Confluent.Kafka
             remove { consumer.OnPartitionsRevoked -= value; }
         }
 
-        public event EventHandler<CommittedOffsets> OnOffsetCommit
+        public event EventHandler<CommittedOffsets> OnOffsetsCommitted
         {
-            add { consumer.OnOffsetCommit += value; }
-            remove { consumer.OnOffsetCommit -= value; }
+            add { consumer.OnOffsetsCommitted += value; }
+            remove { consumer.OnOffsetsCommitted -= value; }
         }
 
         public event EventHandler<LogMessage> OnLog
@@ -434,10 +434,9 @@ namespace Confluent.Kafka
             /* rd_kafka_topic_partition_list_t * */ IntPtr offsets,
             IntPtr opaque)
         {
-            OnOffsetCommit?.Invoke(this, new CommittedOffsets(
-                // TODO: check to see whether errors can ever be present here. If so, expose TPOE, not TPO.
-                SafeKafkaHandle.GetTopicPartitionOffsetErrorList(offsets).Select(tp => tp.TopicPartitionOffset).ToList(),
-                err
+            OnOffsetsCommitted?.Invoke(this, new CommittedOffsets(
+                SafeKafkaHandle.GetTopicPartitionOffsetErrorList(offsets),
+                new Error(err)
             ));
         }
 
@@ -523,7 +522,7 @@ namespace Confluent.Kafka
         /// <remarks>
         ///     Executes on the same thread as every other Consumer event handler (except OnLog which may be called from an arbitrary thread).
         /// </remarks>
-        public event EventHandler<CommittedOffsets> OnOffsetCommit;
+        public event EventHandler<CommittedOffsets> OnOffsetsCommitted;
 
         /// <summary>
         ///     Raised on critical errors, e.g. connection failures or all brokers down.
@@ -650,7 +649,7 @@ namespace Confluent.Kafka
         ///     Manually consume message or triggers events.
         ///
         ///     Will invoke events for OnPartitionsAssigned/Revoked,
-        ///     OnOffsetCommit, etc. on the calling thread.
+        ///     OnOffsetsCommitted, etc. on the calling thread.
         /// </summary>
         public bool Consume(out Message message, int millisecondsTimeout)
         {
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -732,7 +732,11 @@ namespace Confluent.Kafka
         {
             if (consumerCts == null)
             {
-                throw new Exception("Consumer background poll loop not started - cannot stop.");
+                // Consumer background poll loop not started - cannot stop.
+                // While this might be considered an error there is no point
+                // in throwing an exception here since there are no side-effects
+                // of not being able to stop a non-Started instance.
+                return;
             }
 
             consumerCts.Cancel();
@@ -796,6 +800,8 @@ namespace Confluent.Kafka
 
         public void Dispose()
         {
+            // Make sure background poller is stopped (no-op if already stopped, or not started)
+            Stop();
             kafkaHandle.ConsumerClose();
             kafkaHandle.Dispose();
         }
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -493,7 +493,8 @@ namespace Confluent.Kafka
             var pollSetConsumerError = kafkaHandle.PollSetConsumer();
             if (pollSetConsumerError != ErrorCode.NO_ERROR)
             {
-                throw new KafkaException(pollSetConsumerError, "Failed to redirect the poll queue to consumer_poll queue");
+                throw new KafkaException(new Error(pollSetConsumerError,
+                    $"Failed to redirect the poll queue to consumer_poll queue: {Error.ErrorCode2String(pollSetConsumerError)}"));
             }
         }
 
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -365,8 +365,6 @@ namespace Confluent.Kafka
         private LibRdKafka.ErrorDelegate errorDelegate;
         private void ErrorCallback(IntPtr rk, ErrorCode err, string reason, IntPtr opaque)
         {
-            // TODO: Is reason ever different from that returned by err2str?
-            //       If so, sort something else out here.
             OnError?.Invoke(this, new Error(err, reason));
         }
 
@@ -493,7 +491,7 @@ namespace Confluent.Kafka
             if (pollSetConsumerError != ErrorCode.NO_ERROR)
             {
                 throw new KafkaException(new Error(pollSetConsumerError,
-                    $"Failed to redirect the poll queue to consumer_poll queue: {Error.ErrorCode2String(pollSetConsumerError)}"));
+                    $"Failed to redirect the poll queue to consumer_poll queue: {ErrorCodeExtensions.GetReason(pollSetConsumerError)}"));
             }
         }
 
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -254,8 +254,8 @@ namespace Confluent.Kafka
         /// <summary>
         ///     Commit offsets for the current assignment.
         /// </summary>
-        public void Commit()
-            => consumer.Commit();
+        public Task<CommittedOffsets> CommitAsync()
+            => consumer.CommitAsync();
 
         /// <summary>
         ///     Commits an offset based on the topic/partition/offset of a message.
@@ -265,14 +265,14 @@ namespace Confluent.Kafka
         ///     A consumer which has position N has consumed records with offsets 0 through N-1 and will next receive the record with offset N.
         ///     Hence, this method commits an offset of <param name="message">.Offset + 1.
         /// </remarks>
-        public void Commit(Message<TKey, TValue> message)
-            => consumer.Commit(new List<TopicPartitionOffset> { new TopicPartitionOffset(message.TopicPartition, message.Offset + 1) });
+        public Task<CommittedOffsets> CommitAsync(Message<TKey, TValue> message)
+            => consumer.CommitAsync(new List<TopicPartitionOffset> { new TopicPartitionOffset(message.TopicPartition, message.Offset + 1) });
 
         /// <summary>
         ///     Commit explicit list of offsets.
         /// </summary>
-        public void Commit(ICollection<TopicPartitionOffset> offsets)
-            => consumer.Commit(offsets);
+        public Task<CommittedOffsets> CommitAsync(ICollection<TopicPartitionOffset> offsets)
+            => consumer.CommitAsync(offsets);
 
         public void Dispose()
             => consumer.Dispose();
@@ -746,8 +746,8 @@ namespace Confluent.Kafka
         /// <summary>
         ///     Commit offsets for the current assignment.
         /// </summary>
-        public void Commit()
-            => kafkaHandle.Commit();
+        public async Task<CommittedOffsets> CommitAsync()
+            => await kafkaHandle.CommitAsync();
 
         /// <summary>
         ///     Commits an offset based on the topic/partition/offset of a message.
@@ -757,14 +757,18 @@ namespace Confluent.Kafka
         ///     A consumer which has position N has consumed records with offsets 0 through N-1 and will next receive the record with offset N.
         ///     Hence, this method commits an offset of <param name="message">.Offset + 1.
         /// </remarks>
-        public void Commit(Message message)
-            => Commit(new List<TopicPartitionOffset> { new TopicPartitionOffset(message.TopicPartition, message.Offset + 1) });
+        public async Task<CommittedOffsets> CommitAsync(Message message)
+        {
+            if (message.Error.Code != ErrorCode.NO_ERROR)
+                throw new InvalidOperationException("Must not commit offset for errored message");
+            return await CommitAsync(new List<TopicPartitionOffset> { new TopicPartitionOffset(message.TopicPartition, message.Offset + 1) });
+        }
 
         /// <summary>
         ///     Commit explicit list of offsets.
         /// </summary>
-        public void Commit(ICollection<TopicPartitionOffset> offsets)
-            => kafkaHandle.Commit(offsets);
+        public async Task<CommittedOffsets> CommitAsync(ICollection<TopicPartitionOffset> offsets)
+            => await kafkaHandle.CommitAsync(offsets);
 
         /// <summary>
         ///     Retrieve current committed offsets for topics + partitions.
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -367,7 +367,7 @@ namespace Confluent.Kafka
         {
             // TODO: Is reason ever different from that returned by err2str?
             //       If so, sort something else out here.
-            OnError?.Invoke(this, err);
+            OnError?.Invoke(this, new Error(err, reason));
         }
 
         private LibRdKafka.StatsDelegate statsDelegate;
--- a/src/Confluent.Kafka/Error.cs
+++ b/src/Confluent.Kafka/Error.cs
@@ -75,11 +75,16 @@ namespace Confluent.Kafka
         public static bool operator !=(Error a, Error b)
             => !(a == b);
 
-        // Returns the string representation of the error.
-        // If a rich error string is available return that, otherwise fall
-        // back to librdkafka's static error code to string conversion.
+        /// <summary>
+        ///   Returns the string representation of the error.
+        ///   Depending on error source this might be a rich
+        ///   contextual error message, or a simple static
+        ///   string representation of the error Code.
+        /// </summary>
         public override string ToString()
         {
+            // If a rich error string is available return that, otherwise fall
+            // back to librdkafka's static error code to string conversion.
             if (!string.IsNullOrEmpty(reason))
                 return reason;
             else
--- a/src/Confluent.Kafka/Error.cs
+++ b/src/Confluent.Kafka/Error.cs
@@ -88,7 +88,7 @@ namespace Confluent.Kafka
             if (!string.IsNullOrEmpty(reason))
                 return reason;
             else
-                return ErrorCodeExtensions.GetReason(Code);
+                return Code.GetReason();
         }
     }
 }
--- a/src/Confluent.Kafka/Error.cs
+++ b/src/Confluent.Kafka/Error.cs
@@ -18,27 +18,30 @@
 
 namespace Confluent.Kafka
 {
-    public struct Error
+    public class Error
     {
         public Error(ErrorCode code)
         {
             Code = code;
-            String = null;
+            reason = null;
         }
 
         public Error(ErrorCode code, string reason)
         {
             Code = code;
-            String = reason;
+            this.reason = reason;
         }
 
         public ErrorCode Code { get; }
 
-        private string String; // Rich error string, will be empty for APIs
-                               // where there was just an ErrorCode. See ToString()
+        private string reason;
 
-        public string Message
-            => ToString();
+        // Rich error string, will be empty for APIs
+        // where there was just an ErrorCode. See ToString()
+        public string Reason
+        {
+            get { return ToString(); }
+        }
 
         public bool HasError
             => Code != ErrorCode.NO_ERROR;
@@ -77,14 +80,10 @@ namespace Confluent.Kafka
         // back to librdkafka's static error code to string conversion.
         public override string ToString()
         {
-            if (!string.IsNullOrEmpty(String))
-                return String;
+            if (!string.IsNullOrEmpty(reason))
+                return reason;
             else
-                return ErrorCode2String(Code);
+                return ErrorCodeExtensions.GetReason(Code);
         }
-
-
-        // Return the human readable representation of a (librdkafka) error code
-        static public string ErrorCode2String(ErrorCode code) => Internal.Util.Marshal.PtrToStringUTF8(Impl.LibRdKafka.err2str(code));
     }
 }
--- a/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
+++ b/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
@@ -22,7 +22,7 @@ using System.Linq;
 using System.Text;
 using System.Runtime.InteropServices;
 using Confluent.Kafka.Internal;
-
+using System.Threading.Tasks;
 
 namespace Confluent.Kafka.Impl
 {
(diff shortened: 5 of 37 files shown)


mhowlett:
Aside for it's use in KafkaException, where do you need to associate a custom string with an ErrorCode?

src/Confluent.Kafka/KafkaException.cs
```diff
@@ -23,12 +23,19 @@ namespace Confluent.Kafka
 {
     public class KafkaException : Exception
     {
-        public KafkaException(Error error, string message)
-            : base(message)
+        private Error _error;
```
mhowlett: this could be done without explicitly specifying the private member here (as in the existing code). The only thing this can't give you is immutability (i.e. making it readonly, which is another thing you could do), though it's such a trivial class i'd prefer terseness over that. Also coding convention is to not use _'s. I probably agree with you _'s is better, even though that is not official C# coding standard, but if we're going to do that we should do it across the entire codebase.

edenhill: There wasn't much choice left: Error was taken by the struct type, error is the argument name, so I resorted to what was left, underscore-prefixing. Since it is a private member I think it might be okay for now.
Also, there doesn't really seem to be one definitive style guide for .NET (as previously discussed), but we should of course decide on one style for all our code (generally speaking, there must be exceptions, for exceptions.)

edenhill:
Re error strings: search rdkafka.h for `errstr`  to see the APIs that currently provide richer errors.
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h

Plus the error_cb.
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L1142

src/Confluent.Kafka/KafkaException.cs
mhowlett: Yeah. There are recommendations by MS (but they themselves don't follow them in their code)

I mean: 

```
    public class KafkaException : Exception
    {
        public KafkaException(Error error)
            : base(error.ToString())
        {
            Error = error;
        }

        public KafkaException(ErrorCode code)
            : base(Error.ErrorCode2String(code))
        {
            Error = new Error(code);
        }

        public Error Error { get; }
    }
```

edenhill: Isnt having the same name on the member as its type more confusing than the uglyness of underscore prefix?

mhowlett:
Ok, so we potentially need a custom reason string in the error callback and in places that will map to exceptions.

I see why you like having the Error class in this PR - it can be used as the basis of KafkaException and in the OnError event. Both are important and need Code+CustomString together, so it seems natural that the thing we call Error should be that. Still, I'm not convinced we've nailed this quite yet.

My primary concern is avoiding an extra 8 bytes in every delivery report, where this seems easily avoidable. One solution is to change Message to include ErrorCode (not Error), but that's less convenient to the user and it's in a place that I think will come into play often.

Alternatively, the error event delegate could provide an ErrorCode and a Reason (not encapsulated in something else) and KafkaException can remain as is. When I implemented OnError, I was anticipating probably having to change it to this, because I thought there might be custom reason strings there (see my todo).

Error (as I have it) is effectively ErrorCode ++ some zero cost syntactic sugar that allows easy access to the associated human readable text. This pattern is actually slightly a-typical - I don't know where else I've seen something similar (though I think it works quite well). That's probably the primary reason I I'm not completely locked to it. Having ErrorCode in Message + static method on Error may make me happy.

edenhill:
I hear you and I agree that we want to keep the size/object creation cost down for the fast path (which exceptions are not, but Message is).
But we need to make this decision based on real numbers, so we'll need performance benchmarks for the following Message error alternatives:
 1. Message.ErrorCode
 2. Message.Error
 3. Message.Nullable<Error>

My preference would be: 2, 3, 1

mhowlett:
I think i've finally put my finger on the core thing thats irking me here:

We have the concept of an error code and we have a method which can be used to get a message associated with that error code. There is a 1:1 mapping between these two. 

Additionally, sometimes, a different message can be passed along with an error code. This is a different kind of thing. 

Having an Error class where the message is either the 1:1 mapping, or some other arbitrary string is not clear modeling of ideas. 

- it's potentially confusing in fact. 

Having the capability to hold an arbitrary thing in a place (Message) where no such arbitrary thing is ever going to be put does not gel well with me.

[any efficiency gains are side benefits to getting the modeling correct]

edenhill:
I'm not really buying into models and principals, at the end of the day there are two people we need to help out as best as we can: the application developer and its end-user. The better information we can provide for both of them, the more succesful they will be, and consequently us.

A error serves two purposes:
 * let the developer handle the error programmatically (error code)
 * inform the end-user what's up (error string)

I don't see the problem of composing those two things in the same construct, they're two sides of the same coin, two facets of the same thing.

If the error string is sometimes richer than a simple static err2str mapping then I'd say that is great for he end-user, it'll help them sort out what's wrong.

So, is putting an error string in the Message.Error field wasteful?
I'd say no, because most Message handling code will eventually look like this:
```
 if (!msg.Error)
    return handle_the_msg(msg);
 switch (msg.Error.Code)
 {
     case SomethingIKnowToHandle:
        dostuff();
        break;
     case SomethingElseIthinkIknowtoHandle:
       dootherstuff();
        break;
     default:
       log_to_enduser("Couldn't handle errored message at {msg.TopicPartition] offset {msg.Offset}: {msg.Error});
       break;
}
```` 
Thus, a developer will never care what the error string for an error is, because it is never programmatically used, all she wants is proper error propagation to a human.

Here are a couple of examples of rich Message errors:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_broker.c#L3773
Compare:
  "Unsupported codec" to "Unsupported compression codec 0x%x" - even though this is mildly helpful it at least gives a technically savvy person a chance to know what codec isn't supported.


https://github.com/edenhill/librdkafka/blob/master/src/rdkafka_broker.c#L4172
Compare:
  "Message too large" to "Message at offset .. too large"


https://github.com/edenhill/librdkafka/blob/dcfc3ed789d450195189502f7a5ec48b2dd643b7/src/rdkafka_cgrp.c#L941
Compare:
   "Illegal generation" to "Failed to Join group: Illegal generation"


https://github.com/edenhill/librdkafka/blob/bbee686594d009a078758cba7070dbb3964f771c/src/rdkafka_partition.c#L1900
Compare: 
 "Not coordinator for group" vs "Failed to fetch offsets from brokers: Not coordinator for group"

**Comment by edenhill** [comment]
Maybe we could convert some of these errors into TopicPartitionOffsetError's perhaps and have its own onConsumerError for them

**Comment by mhowlett** [comment]
Yeah - the end user should be able to get the better error message if there is one - of course don't want to limit that. Also, maybe those custom errors are close enough to being the same as the 1:1 mapping that the argument in my previous comment isn't that strong. Still, in Message, they *are* 1:1 and it's still dominating my thinking. 

A negative on the Error class as I have it is it's just a wrapper of ErrorCode and this may not 100% future proof for something with the name Error. Extending it at all breaks the concept.

I just realized something - you can put extension methods on Enums: 
https://msdn.microsoft.com/en-us/library/bb383974.aspx

That means we could put an ErrorCode in Message, and also define an extension method which calls  err2str.

**Comment by mhowlett** [comment]
That's currently doing it for me - keep the `Error` class as you have it but put `ErrorCode`s in `Message` (which will have a convenient method on them to get the `err2str` string via an extension method).

**Comment by mhowlett** [comment]
Actually I wanted to re-open the discussion on whether we want to break out the OnError event into consume and other errors again. We decided they should be joined before, but I feel like I don't have a good opinion on that currently. Just wanted to ask the question again.

File: src/Confluent.Kafka/KafkaException.cs (hunk 1 above)
Comment (ewencp) [inline src/Confluent.Kafka/KafkaException.cs]: @edenhill If you want the same name as a field, doesn't using `this.error = error;` work?

File: src/Confluent.Kafka/Producer.cs [hunk 2]
```diff
@@ -382,12 +381,13 @@ public long OutQueueLength
         /// <summary>
         ///     Wait until all outstanding produce requests, et.al, are completed. This should typically be done prior
         //      to destroying a producer instance to make sure all queued and in-flight produce requests are completed
-        //      before terminating.
+        //      before terminating. The wait is limited by the optional millisecondsTimeout parameter.
+        //      Returns the number of remaining messages and requests in queue.
```
ewencp: Shouldn't the return value be in `<returns>` tags?

src/Confluent.Kafka/KafkaException.cs
mhowlett: You also have a member with the same name as the type (this is something which happens very frequently, no problem) - you're just using old skool notation to write it out. Yes, what @ewencp says is how to write it out long hand in the current coding style.

edenhill: Okay, fixed according to Matt's code snippet

examples/Misc/Program.cs
```diff
@@ -57,7 +57,7 @@ static void PrintMetadata(string brokerList)
             var config = new Dictionary<string, object> { { "bootstrap.servers", brokerList } };
             using (var producer = new Producer(config))
             {
-                var meta = producer.GetMetadata();
+                var meta = producer.GetMetadata(true, null);
```
edenhill: I agree, but what exactly is the common usage? all topics? single topic? locally used topics?
What is the common timeout?

edenhill:
@mhowlett I can't seem to reply to the above thread on Message.Error, but to follow up:
There are two big concerns with decoupling the rich error from the Message:
 * the rich consumer error will most likely need to use a TopicPartitionOffsetError to provide context. Do we want to add that to the Error type (as a Nullable)? Add a OnConsumerError? or just keep the Message.Error as today since there is a Message.TopicPartitionOffsetdo we want that as an optional on the OnError?
 * sequenciality: separating in-stream error messages from OnMessage means that ordering is lost. E.g., you can get the delegate for a later message before the On*Error. I don't have a specific case where this is bad but I'm quite sure there will be problems in some use-cases doing this.

mhowlett:
Firstly, currently no context is provided with consume errors - this needs to be rectified, yes (I hadn't yet got to thinking this through thoroughly).

Regarding sequentiality: events for consumer errors and messages *are* fired in order (ordering is not lost). It's possibly confusing that consumed messages are spread over two event handlers - and probably definitely confusing that one of these events can also be fired as a result of something other than a consumed message with error. For that reason I'm now against the idea of consumer errors being exposed via `OnError` along with other errors. Either we have `OnConsumerError` or they should go to `OnMessage`. 

If we have `OnConsumeError`, we could simply provide a `Message` to the event rather than invent a new type - it contains everything we need, and it maps to what's happening under the covers.

A benefit of having an `OnConsumeError` is we're effectively helping people write their message handling logic. I'm guessing it's probably going to result in slightly cleaner implementations. The downside is it's possibly more confusing to new users - it sort of assumes working knowledge of closures / comprehension that everything happens on the same thread / realizing ordering is guaranteed between the events. I don't have a strong opinion here, but currently I'd err slightly on the side of `OnConsumeError`.

`rd_kafka_message_t` doesn't have a rich error message. `rd_kafka_topic_partition_t` also doesn't have a rich error message (unless i'm missing something?). I still can't see a good reason to have something richer than `ErrorCode` in `Message` or `TopicPartitionOffsetError`. An extension method on `ErrorCode` that calls through to `err2str` still seems to me like it'll do everything we need and be an abstraction that is unlikely to break or cause problems in the future.

I think we probably do want an abstraction for a rich error (and it can be called `Error`), but this is applicable to `OnError` and `KafkaException`.

edenhill:
> rd_kafka_message_t doesn't have a rich error message. 

Actually it does: `payload` doubles as error string in consumer mode when `err` is non-zero, see here:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L832

I posted a couple of examples where the errors will be richer than a simple error string, see earlier in this thread, please see https://github.com/confluentinc/confluent-kafka-dotnet/pull/38#issuecomment-277339909 again

mhowlett:
Ok, I hadn't understood that - this changes everything.

mhowlett:
Given this, I think `Message` should have `Error` as you define it, as you suggest.

What about `rd_kafka_topic_partition_t`? there is no rich error message there right? in that case I probably still think we should only have `ErrorCode` in that.

edenhill:
Correct, topic_partition_t has just the error code (it is mapped very closely to the underlying Kafka protocol which does the Topic,Partition,Error tuple frequently).

I agree that ErrorCode is technically sufficient on TopicPartition, but I'm wondering if that spills implementation details out into the public API and making it harder to use? Some types have ErrorCode, other have Error, ..
I'd rather see Error used consistently everywhere, and from the user's perspective they want both the error code and error string as described above.

Since errors are part of the slow-path I would suggest that we make Message.Error a Nullable, or whatever it is called, e.g., not allocated at all if there is no error.

mhowlett:
Yep, I think it makes sense to make `Error` a class (meaning it's heap allocated). It can then have a value of null without wrapping it in Nullable<T>. In the normal case of no error, this gets set to null - so cost per message in the normal case is size of a reference + there is no heap allocation except in case of error. I like it.

I'm currently not sure about `Error` vs `ErrorCode` in `TopicPartitionOffsetError`. Erring on the side of your thinking is probably right.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -367,7 +367,7 @@ private void ErrorCallback(IntPtr rk, ErrorCode err, string reason, IntPtr opaqu
         {
             // TODO: Is reason ever different from that returned by err2str?
```
mhowlett: can remove this comment now.

examples/AdvancedConsumer/Program.cs
```diff
@@ -189,7 +189,7 @@ public static void Run_Consume(string brokerList, List<string> topics)
                 consumer.OnError += (_, error)
                     => Console.WriteLine($"Error: {error}");
 
-                consumer.OnOffsetCommit += (_, commit) =>
+                consumer.OnOffsetsCommitted += (_, commit) =>
```
mhowlett: do we still want this when manually committing offsets?

edenhill: Yes, it is for auto commits.

src/Confluent.Kafka/Consumer.cs
edenhill: I missed the same spot in the Producer. Would be good if code like this was in a shared place instead. DRY and all that

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -397,42 +403,65 @@ internal void Assign(ICollection<TopicPartitionOffset> partitions)
...
-            if (list == IntPtr.Zero)
-            {
-                throw new Exception("Failed to create offset commit list");
-            }
-            foreach (var offset in offsets)
+            // Create temporary queue so we can get the offset commit results
+            // as an event instead of a callback.
+            // We still need to specify a dummy callback (that does nothing)
+            // to prevent the global offset_commit_cb to kick in.
+            IntPtr cQueue = LibRdKafka.queue_new(handle);
```
mhowlett: I still don't properly understand whether `SafeHandle` has any meaningful advantages in situations like this (where the `IntPtr` is just some memory given to us by librdkafka). Whether or not it does, we can't use it to own the pointer because commit_queue takes ownership (conditional on success/failure).

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -523,6 +552,31 @@ internal static List<TopicPartitionOffsetError> GetTopicPartitionOffsetErrorList
...
         }
 
+        /// <summary>
+        /// Creates and returns a C rd_kafka_topic_partition_list_t * populated by offsets.
+        /// </summary>
+        /// <returns>
+        /// If offsets is null a null IntPtr will be returned, else a IntPtr
+        /// which must destroyed with LibRdKafka.topic_partition_list_destroy()
+        /// </returns>
+        internal static IntPtr GetCTopicPartitionList(ICollection<TopicPartitionOffset> offsets)
```
mhowlett: there is code similar to this in quite a number of places, probably be good to remove the duplication at some point (not necessarily in this PR).

I'm not too fond of the C in the name here, though it's internal so hardly matters.

src/Confluent.Kafka/KafkaException.cs
```diff
@@ -23,12 +23,18 @@ namespace Confluent.Kafka
 {
     public class KafkaException : Exception
     {
-        public KafkaException(Error error, string message)
-            : base(message)
+        public KafkaException(Error error)
+            :base(error.ToString())
```
mhowlett: space after the : would be more consistent with styling.

src/Confluent.Kafka/KafkaException.cs
```diff
@@ -23,12 +23,18 @@ namespace Confluent.Kafka
...
         {
             Error = error;
         }
 
+        public KafkaException(ErrorCode code)
+            :base(Error.ErrorCode2String(code))
```
mhowlett: space

src/Confluent.Kafka/Producer.cs
```diff
@@ -381,13 +380,16 @@ public long OutQueueLength
 
         /// <summary>
         ///     Wait until all outstanding produce requests, et.al, are completed. This should typically be done prior
-        //      to destroying a producer instance to make sure all queued and in-flight produce requests are completed
-        //      before terminating.
+        ///     to destroying a producer instance to make sure all queued and in-flight produce requests are completed
+        ///     before terminating. The wait is limited by the optional millisecondsTimeout parameter.
         /// </summary>
-        public void Flush(TimeSpan timeout)
```
mhowlett: It is typical in the BCL to provide both TimeSpan and millisecondsTimeout overloads in analogous situations, so I'm in favor of leaving both. Eg
https://msdn.microsoft.com/en-us/library/system.threading.waithandle_methods(v=vs.110).aspx
https://msdn.microsoft.com/en-us/library/system.threading.thread_methods(v=vs.110).aspx

edenhill: We discussed that at some point and I think we came to  the conclusion that we should favour  API frugality at this point and add stuff on-demand.
In this particular case it will mostly be called with a constant hardcoded timeout so people will most likely not use TimeSpans, and if they do it is trivial enough to convert  to int

src/Confluent.Kafka/Producer.cs
```diff
@@ -605,10 +610,10 @@ public void ProduceAsync(string topic, TKey key, TValue val, IDeliveryHandler<TK
         public event EventHandler<Error> OnError;
 
 
-        public void Flush(TimeSpan timeout)
+        public long Flush(int timeout)
```
mhowlett: Again, I'm in favor of having three flush variants here.

src/Confluent.Kafka/Structs/Error.cs
```diff
@@ -23,14 +23,22 @@ public struct Error
...
+        {
+            Code = code;
+            String = reason;
         }
 
         public ErrorCode Code { get; }
 
-        // Note: In most practical scenarios there is no benefit to caching this +
-        //       significant cost in keeping the extra string reference around.
+        private string String; // Rich error string, will be empty for APIs
```
mhowlett: shouldn't be capitalized. also, I like the idea of calling this reason.

src/Confluent.Kafka/Structs/Error.cs
```diff
@@ -23,14 +23,22 @@ public struct Error
...
+                               // where there was just an ErrorCode. See ToString()
+
         public string Message
```
mhowlett: I'd change the name of `Message` to `Reason`

src/Confluent.Kafka/Structs/Error.cs
```diff
@@ -64,8 +72,19 @@ public override int GetHashCode()
         public static bool operator !=(Error a, Error b)
             => !(a == b);
 
+        // Returns the string representation of the error.
```
mhowlett: I'd put the first line in a <summary></summary> block and the second two lines in <reason></reason>

src/Confluent.Kafka/Structs/Error.cs
```diff
@@ -64,8 +72,19 @@ public override int GetHashCode()
...
+        // If a rich error string is available return that, otherwise fall
+        // back to librdkafka's static error code to string conversion.
         public override string ToString()
-            => $"[{(int)Code}] {Message}";
+        {
+            if (!string.IsNullOrEmpty(String))
+                return String;
```
mhowlett: I always use braces no matter what. can we consider that the coding style?

src/Confluent.Kafka/Structs/Error.cs
edenhill: I think you need to escape / gt lt your comment :)

src/Confluent.Kafka/Structs/Error.cs
```diff
@@ -64,8 +72,19 @@ public override int GetHashCode()
...
+            else
+                return ErrorCode2String(Code);
+        }
+
 
+        // Return the human readable representation of a (librdkafka) error code
+        static public string ErrorCode2String(ErrorCode code) => Internal.Util.Marshal.PtrToStringUTF8(Impl.LibRdKafka.err2str(code));
```
mhowlett: I'd do this as extension method on `ErrorCode`. I'd put the following in ErrorCode.cs in the Confluent.Kafka namespace:
 
```
    public static class ErrorCodeExtensions
    {
        public static string GetReason(this ErrorCode code)
        {
            return Internal.Util.Marshal.PtrToStringUTF8(Impl.LibRdKafka.err2str(code));
        }
    }
```

test/Confluent.Kafka.UnitTests/Error.cs
```diff
@@ -27,6 +27,11 @@ public void Constuctor()
             var e = new Error(ErrorCode._BAD_COMPRESSION);
             Assert.Equal(e.Code, ErrorCode._BAD_COMPRESSION);
             Assert.NotNull(e.Message);
+            Assert.Equal(e.Message, "Local: Invalid compressed data");
```
mhowlett: I realized I've been typically getting the order of arguments wrong in Assert.Equal. First argument is expected, last argument is actual. This is noted in messages when running the tests so we should change (because it's confusing otherwise). However, no need to fix here I think, I'll do a PR which changes it everywhere.

src/Confluent.Kafka/Structs/Error.cs
```diff
@@ -23,14 +23,22 @@ public struct Error
         public Error(ErrorCode code)
         {
             Code = code;
+            String = null;
+        }
+
+        public Error(ErrorCode code, string reason)
```
mhowlett: (not quite the right place for this comment, but close): I think `Error` would be better as a class (not a struct) for reasons mentioned elsewhere. Probably move Error.cs out of the structs folder as well.

examples/Misc/Program.cs
ah-: How about an enum?

src/Confluent.Kafka/Consumer.cs
mhowlett: Yes. [and in fact @ah- was doing this by deriving from `Handle`]. I didn't like having both `Handle` and `SafeKafkaHandle` + some other reasons for changing away from this. This duplicate functionality could all be put in SafeKafkaHandle easily enough and passed through.

src/Confluent.Kafka/Producer.cs
mhowlett: Yeah we did discuss. Though I think the optimal API is to be able to write `TimeSpan.FromMilliseconds(1000)` for clarity (i prefer this), even though it's longer than `1000`. `(int)TimeSpan.FromMilliseconds(1000).TotalMilliseconds` isn't very nice. Perf. isn't a consideration here.

src/Confluent.Kafka/Structs/Error.cs
edenhill: Is intended use `ErrorCodeExtensions.GetReason()`?

examples/AdvancedConsumer/Program.cs
mhowlett: whoops, I was thinking this example was manual committing but it's not.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -734,7 +732,11 @@ public void Stop()
         {
             if (consumerCts == null)
             {
-                throw new Exception("Consumer background poll loop not started - cannot stop.");
+                // Consumer background poll loop not started - cannot stop.
+                // While this might be considered an error there is no point
+                // in throwing an exception here since there are no side-effects
+                // of not being able to stop a non-Started instance.
+                return;
```
mhowlett: on the other hand it indicates a logic error and I would argue that throwing an exception to point that out is a good thing. 

note: this has all been refactored in an open PR + I think there is a good chance we'll even cut background thread mode from the upcoming release as I think there might be a better factoring than what i've got (maybe better to take it outside the `Consumer` class).

src/Confluent.Kafka/Consumer.cs
```diff
@@ -794,6 +800,8 @@ public List<TopicPartitionOffsetError> Position(ICollection<TopicPartition> part
 
         public void Dispose()
         {
+            // Make sure background poller is stopped (no-op if already stopped, or not started)
+            Stop();
```
mhowlett: yep, refactored + tests that cover this in open PR.

src/Confluent.Kafka/Error.cs
```diff
@@ -18,19 +18,30 @@
 
 namespace Confluent.Kafka
 {
-    public struct Error
+    public class Error
```
mhowlett: 👍

src/Confluent.Kafka/Error.cs
```diff
@@ -64,8 +75,15 @@ public override int GetHashCode()
...
+        // If a rich error string is available return that, otherwise fall
+        // back to librdkafka's static error code to string conversion.
         public override string ToString()
-            => $"[{(int)Code}] {Message}";
-
+        {
+            if (!string.IsNullOrEmpty(reason))
+                return reason;
+            else
+                return ErrorCodeExtensions.GetReason(Code);
```
mhowlett: `Code.GetReason()`

src/Confluent.Kafka/Error.cs
```diff
@@ -18,19 +18,30 @@
...
         public ErrorCode Code { get; }
 
-        // Note: In most practical scenarios there is no benefit to caching this +
-        //       significant cost in keeping the extra string reference around.
-        public string Message
-            => Internal.Util.Marshal.PtrToStringUTF8(Impl.LibRdKafka.err2str(Code));
+        private string reason;
+
+        // Rich error string, will be empty for APIs
+        // where there was just an ErrorCode. See ToString()
```
mhowlett: Comment like `/// <summary></summary>` tags (see elsewhere for formatting).

Also, comment is wrong (it'll be rich error string or that returned by `ErrorCode.GetReason`

src/Confluent.Kafka/project.json
```diff
@@ -14,8 +14,7 @@
     },
 
     "frameworks": {
-        "net451": { },
```
mhowlett: these shouldn't change. we should continue to target the lowest version of netstandard we can and only increase this for good reason. and unfortunately we need to target net451 separately as this this is not compatible with netstandard1.3. changing to netcoreapp1.0 means we can't link with .net framework apps.

more info on netstandard versions.
https://blogs.msdn.microsoft.com/dotnet/2016/09/26/introducing-net-standard/

unfortunately I don't know any way to make dotnet happy with this configuration on linux/mac. I have a script  that allows me to comment/uncomment net451 (yes, in a JSON file, but dotnet doesn't complain) which i run before and after checking in (have a terminal window dedicated to running this script). There must be a better way... surely. 

```
#!/usr/local/bin/python

with open ("/git/confluent-kafka-dotnet/src/Confluent.Kafka/project.json", "r") as myfile:
        data = myfile.readlines()

out = []
for line in data:
        line2 = line[:-1]
        if "net451" in line2:
                if line2[0:2] == "/*":
                        line2 = line2[2:-2]
                else:
                        line2 = "/*" + line2 + "*/"
        out.append(line2)

with open ("/git/confluent-kafka-dotnet/src/Confluent.Kafka/project.json", "w") as myfile:
        for line in out:
                myfile.write(line + "\n")
```

src/Confluent.Kafka/Producer.cs
edenhill: We'll add them later.

src/Confluent.Kafka/Structs/Error.cs
edenhill: Okay, can we make a later pass when we have code formatter or lint checker in place?

src/Confluent.Kafka/Consumer.cs
edenhill: The problem stems from (auto) disposing a consumer without calling Stop() first: it will leave threads behind and eventually assert on some refcount issue.
I also think we should be forgiving in APIs like these, otherwise app developers will need to keep track state (did I start or not?) that we already have.

src/Confluent.Kafka/Error.cs
edenhill: Didnt know that, cool thing (this .. )

src/Confluent.Kafka/project.json
edenhill: Yeah, that's messy, couldn't get it to work on Linux wiht that line still there.
It might be possible with build --framework 

Will revert that change.

edenhill: Yeah, -f works. I'll add that to the Makefile for convenience

src/Confluent.Kafka/KafkaException.cs
edenhill: VS formatting
fixed though

src/Confluent.Kafka/project.json
mhowlett: I just realized the reason I use that script is mostly VS Code / omni-sharp which can't deal with net451 being in there (by default.. and if there's a config setting I didn't find it).

examples/AdvancedConsumer/Program.cs
mhowlett: comment above was just in the wrong spot: In this example we are manually committing offsets and still registered for the OnOffsetsCommitted event which is no longer relevant.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -398,42 +404,65 @@ internal void Assign(ICollection<TopicPartitionOffset> partitions)
...
-                    offset.Offset);
+                LibRdKafka.queue_destroy(cQueue);
+                return new CommittedOffsets(new Error(err));
             }
-            ErrorCode err = LibRdKafka.commit(handle, list, false);
-            LibRdKafka.topic_partition_list_destroy(list);
-            if (err != ErrorCode.NO_ERROR)
+
+            // Wait for commit to finish
+            IntPtr rkev = LibRdKafka.queue_poll(cQueue, -1/*infinite*/);
```
mhowlett: I think having an infinite timeout here won't cut it - when the consumer is disposed this needs to be able to be cancelled (search for `CancellationToken` for examples of how to do this).

Also, do we need a timeout variant of `ConsumerAsync`?

src/Confluent.Kafka/Consumer.cs
mhowlett: the refactor in my open PR changes how all this works and thinks all this (stopping on dispose) through carefully. alternatively, we may cut background processing from this PR (I want to think through more whether this actually belongs on consumer or separate, see comments in my open pr on that. there's also the issue of how exceptions on the background thread are surfaced). Anyway, this bit of code is going away regardless.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -757,14 +759,18 @@ public void Commit()
         ///     A consumer which has position N has consumed records with offsets 0 through N-1 and will next receive the record with offset N.
         ///     Hence, this method commits an offset of <param name="message">.Offset + 1.
         /// </remarks>
-        public void Commit(Message message)
-            => Commit(new List<TopicPartitionOffset> { new TopicPartitionOffset(message.TopicPartition, message.Offset + 1) });
+        public async Task<CommittedOffsets> CommitAsync(Message message)
+        {
+            if (message.Error.Code != ErrorCode.NO_ERROR)
```
mhowlett: coding standard used throughout code is to always use braces where optional.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
edenhill: That's a good point, maybe put that poll(with a short timeout) in a cancellation-checking loop?
This requires the cancellation to be blocking, e.g., the cancelour must block until the cancelee returns, is that possible?

I dont think a timeout variant is needed for the async interface, the caller can use Task.WaitAny(task, Task.Delay(..)) or something like that.

mhowlett: > That's a good point, maybe put that poll(with a short timeout) in a cancellation-checking loop?

yep

> This requires the cancellation to be blocking, e.g., the cancelour must block until the cancelee returns, is that possible?

In `Dispose`, we'd need to block for up to the poll loop interval (e.g. 100ms). Not sure I quite understand the question, but I think everything that needs to be done is possible.

`Task.WaitAny(task, Task.Delay(..))` would wait for a timeout, but would not cancel the poll loop. 95% sure we either need a timeout or overload that takes a `CancellationToken` (which is more sophisticated and i think not necessary).

src/Confluent.Kafka/Consumer.cs
edenhill: Aye, lets clean that up later when we have a lint-checker in place.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
edenhill: Can't really use timeouts to control thread synchronization, proper thread synchronization is required.
We must prohibit the caller (Dispose()) to proceed with destructing the client instance until all blocking tasks (CommitAsync()) are cancelled.
If cancellation tokens provides synchronized and blocking signalling that solves that problem for us, if not we'll have to figure something else out.

mhowlett: yes, the cancellation token stuff is exactly for this purpose.

the timeout variant would be implemented using the thread cancellation stuff, effectively hiding the details of this from the user.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -312,13 +312,13 @@ public void Flush()
         public string MemberId
             => consumer.MemberId;
 
-        /// <summary>
+        /// <returns>
```
mhowlett: not sure how auto-generated docs look but i suspect having something in the `summary` section is going to be good for that.

mhowlett:
I'm going to take the liberty of merging this even though not completely LGTM yet as i want test these changes on windows and rebase + fix conflicts on other PRs. 

List of things remaining to fix up:

1. CommitAsync cancellation
2. Superfluous OnOffsetsCommitted in AdvancedConsumer example
3. Various linting related things
4. GetMetadata method signature
5. TimeSpan Flush variants

We can do the above in later PRs.

PR #39, opened 2017-02-06
Reworking Consumer Poll/Consume/Start/Stop [Experiment]

### LINKED ISSUE #37: Consumer.OnMessage has not been raised
with consumer created by not generic constructor

mhowlett:
Resolves #37 
@ewencp @edenhill 

When calling `Start` on the non-deserializing `Consumer`, an exception was errantly thrown due to checking that the consumer was not in background thread mode. A fix would have been messy without a refactor and investigating the problem also made me realize this alternate factoring was better.


src/Confluent.Kafka/Internal/CancellablePollThread.cs
```diff
@@ -0,0 +1,84 @@
...
+                    }
+                    else
+                    {
+                        return;
+                    }
+                }
+
+                consumerCts.Cancel();
+                consumerTask.Wait();
+                consumerCts = null;
```
tomasdeml: I think it would be best to [dispose](https://msdn.microsoft.com/en-US/library/system.threading.cancellationtokensource(v=vs.110).aspx#Remarks) `consumerCts` when no longer needed (if possible).

src/Confluent.Kafka/Internal/CancellablePollThread.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        public void Stop(bool throwIfNotStarted)
+        {
+            lock (startStopLock)
+            {
+                if (consumerCts == null)
+                {
+                    if (throwIfNotStarted)
+                    {
+                        throw new Exception("Poll loop not started - cannot stop.");
```
tomasdeml: Maybe a more specific exception - like `InvalidOperationException` - would be a better fit here?

src/Confluent.Kafka/Internal/CancellablePollThread.cs
```diff
@@ -0,0 +1,84 @@
...
+            this.millisecondsTimeout = millisecondsTimeout;
+        }
+
+        public void Start(Action<int> pollMethod)
+        {
+            lock (startStopLock)
+            {
+                if (consumerCts != null)
+                {
+                    throw new Exception("Poll loop cannot be started twice.");
```
tomasdeml: Maybe a more specific exception - like `InvalidOperationException` - would be a better fit here?

src/Confluent.Kafka/Consumer.cs
```diff
@@ -671,77 +680,59 @@ public bool Consume(out Message message, int millisecondsTimeout)
...
+        ///     Manually consume message or triggers events.
+        ///
+        ///     Will invoke events for OnPartitionsAssigned/Revoked,
+        ///     OnOffsetCommit, etc. on the calling thread.
+        /// </summary>
+        public bool Consume(out Message message, int millisecondsTimeout)
+        {
+            if (backgroundPollThread.IsStarted)
+            {
+                throw new Exception("Poll/Consume cannot be called on a Consumer when a background poll thread running.");
```
tomasdeml: Maybe a more specific exception - like `InvalidOperationException` - would be a better fit here?

src/Confluent.Kafka/Consumer.cs
```diff
@@ -85,8 +85,13 @@ public class Consumer<TKey, TValue> : IDisposable
 
         public bool Consume(out Message<TKey, TValue> message, int millisecondsTimeout)
         {
+            if (backgroundPollThread.IsStarted)
+            {
+                throw new Exception("Poll/Consume cannot be called on a Consumer when a background poll thread running.");
```
tomasdeml: Maybe a more specific exception - like `InvalidOperationException` - would be a better fit here?

src/Confluent.Kafka/Consumer.cs
```diff
@@ -645,13 +659,8 @@ public void Assign(IEnumerable<TopicPartition> partitions)
...
 
-        /// <summary>
-        ///     Manually consume message or triggers events.
-        ///
-        ///     Will invoke events for OnPartitionsAssigned/Revoked,
-        ///     OnOffsetCommit, etc. on the calling thread.
-        /// </summary>
-        public bool Consume(out Message message, int millisecondsTimeout)
+
+        internal bool ConsumeImpl(out Message message, int millisecondsTimeout)
```
tomasdeml: IMO the `out` parameter [should be ordered last](https://msdn.microsoft.com/en-us/library/ms229015(v=vs.110).aspx). It would be also consistent with other BCL APIs.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -671,77 +680,59 @@ public bool Consume(out Message message, int millisecondsTimeout)
             return false;
         }
 
+        /// <summary>
+        ///     Manually consume message or triggers events.
+        ///
+        ///     Will invoke events for OnPartitionsAssigned/Revoked,
+        ///     OnOffsetCommit, etc. on the calling thread.
+        /// </summary>
+        public bool Consume(out Message message, int millisecondsTimeout)
```
tomasdeml: IMO the `out` parameter [should be ordered last](https://msdn.microsoft.com/en-us/library/ms229015(v=vs.110).aspx). It would be also consistent with other BCL APIs. However I understand it may be a breaking change.

src/Confluent.Kafka/Internal/CancellablePollThread.cs
```diff
@@ -0,0 +1,84 @@
...
+                }
+
+                consumerCts = new CancellationTokenSource();
+                var ct = consumerCts.Token;
+                consumerTask = Task.Factory.StartNew(() =>
```
tomasdeml: IMO it would be better to use `Task.Run` instead as recommended [here](http://blog.stephencleary.com/2013/08/startnew-is-dangerous.html).

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_OperationalScenarios.cs
```diff
@@ -0,0 +1,140 @@
...
+            var consumerConfig = new Dictionary<string, object>
+            {
+                { "group.id", "simple-produce-consume" },
+                { "bootstrap.servers", bootstrapServers },
+                { "session.timeout.ms", 6000 }
+            };
+
+            // VALID Scenarios
+
+            // start then stop then start a Consumer
```
tomasdeml: May I suggest splitting these blocks to separate tests with distinct names? IMO it would greatly improve readability of the tests. Also as you will get multiple test failures instead of just one, it may help to find out the root cause faster.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Background.cs
```diff
@@ -0,0 +1,78 @@
...
+
+namespace Confluent.Kafka.IntegrationTests
+{
+    public static partial class Tests
+    {
+        /// <summary>
+        ///     Basic Consumer test (background poll mode).
+        /// </summary>
+        [Theory, MemberData(nameof(KafkaParameters))]
+        public static void Consumer_Background(string bootstrapServers, string topic)
```
tomasdeml: May I suggest a more intent-revealing name for the test, e.g. `WhenConsumingMessageWithHighLevelConsumer_AllExpectedLifecycleEventsAreRaised`?

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Consume.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace Confluent.Kafka.IntegrationTests
+{
+    public static partial class Tests
+    {
+        /// <summary>
+        ///     Basic Consumer test (consume mode).
+        /// </summary>
+        [Theory, MemberData(nameof(KafkaParameters))]
+        public static void Consumer_Consume(string bootstrapServers, string topic)
```
tomasdeml: Maybe a more descriptive name would convey the intent better?

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_OperationalScenarios.cs
```diff
@@ -0,0 +1,140 @@
...
+
+namespace Confluent.Kafka.IntegrationTests
+{
+    public static partial class Tests
+    {
+        /// <summary>
+        ///     Tests various
+        /// </summary>
+        [Theory, MemberData(nameof(KafkaParameters))]
+        public static void Consumer_OperationalScenarios(string bootstrapServers, string topic)
```
tomasdeml: Maybe a more descriptive name would convey the intent better?

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Poll.cs
```diff
@@ -0,0 +1,79 @@
...
+
+namespace Confluent.Kafka.IntegrationTests
+{
+    public static partial class Tests
+    {
+        /// <summary>
+        ///     Basic Consumer test (poll mode).
+        /// </summary>
+        [Theory, MemberData(nameof(KafkaParameters))]
+        public static void Consumer_Poll(string bootstrapServers, string topic)
```
tomasdeml: Maybe a more descriptive name would convey the intent better?

src/Confluent.Kafka/Internal/CancellablePollThread.cs
```diff
@@ -0,0 +1,84 @@
...
+
+
+namespace Confluent.Kafka
+{
+    internal class CancellablePollThread
+    {
+        private readonly int millisecondsTimeout;
+        private readonly Object startStopLock = new Object();
+
+        private CancellationTokenSource consumerCts = null;
```
tomasdeml: As the fields are initialized to `null` by default, isn't the assignment redundant?

src/Confluent.Kafka/Consumer.cs
edenhill: + "is running"

src/Confluent.Kafka/Consumer.cs
edenhill: "is running"

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Poll.cs
```diff
@@ -0,0 +1,79 @@
+// Copyright 2016-2017 Confluent Inc.
```
edenhill: Filename: PascalCase and snake_case?

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Poll.cs
```diff
@@ -0,0 +1,79 @@
...
+                int msgCnt = 0;
+                bool done = false;
+
+                consumer.OnMessage += (_, msg) =>
+                {
+                    Assert.Equal(msg.Error.Code, ErrorCode.NO_ERROR);
+                    msgCnt += 1;
+                };
+
+                consumer.OnPartitionEOF += (_, partition)
```
edenhill: This will only work if there is just one partition, which is not asserted anywhere.
It is better to do something like:
`if (++eofCnt == currentAssignment.Length) done = truth;`

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Poll.cs
```diff
@@ -0,0 +1,79 @@
...
+                    => done = true;
+
+                consumer.OnPartitionsAssigned += (_, partitions) =>
+                {
+                    Assert.Equal(partitions.Count, 1);
```
edenhill: What is setting up the topic to only have one partition?

It might be better not to have such strict requirements on the test cluster but instead do something like: is better 
`if (++eofCnt == currentAssignment.Length) done = true;`

edenhill:
It would be great if the commit message said how and why this was reworked.
What's the underlying problem that is fixed?

src/Confluent.Kafka/Internal/CancellablePollThread.cs
mhowlett: wow! thanks again for reviewing Tomáš - you're bringing a different perspective in and it's much appreciated.

On this point, I often prefer to show explicit intent even when not required (so will leave this as is). I believe doing this is better (though you could argue it's a matter of style).

src/Confluent.Kafka/Internal/CancellablePollThread.cs
mhowlett: yep

src/Confluent.Kafka/Internal/CancellablePollThread.cs
mhowlett: As you may have guessed, I was using `StartNew` so as to be able to specify the `LongRunning` flag. In the referenced article, Mr Cleary quite confidently states that this is not necessary. Ok, I'm convinced.

src/Confluent.Kafka/Internal/CancellablePollThread.cs
mhowlett: yep

src/Confluent.Kafka/Internal/CancellablePollThread.cs
mhowlett: Well spotted. I think `Dispose` should be called both immediately below `consumerCts.Cancel()` and in a continuation of the `Task.Run` (which will also set `consumerCts` and `consumerTask` to `null` synchronized via `startStopLock`). The latter in case an exception is thrown in one of the event handlers called by the poll method. You've started me thinking about exactly how that scenario should work. I'll write some tests to try and clarify this.

src/Confluent.Kafka/Consumer.cs
mhowlett: yep and yep.

src/Confluent.Kafka/Consumer.cs
mhowlett: yep and yep.

src/Confluent.Kafka/Consumer.cs
mhowlett: I find myself in conflict with the Microsoft recommendations here. 

First I want to defend the use of the out parameter in this scenario - doing it this way here results in both terser usage patterns and avoids making `Message` a class or wrapping it in `Nullable`.

Second, ordering as I have seems more natural (timeout is optional - put last). In fact, if we had have used a default parameter value for the timeout, it would have been required to be specified last. 

I'm quite happy to argue on this one out more, but at this point I feel like I want to leave as is.

side question: @tomasdeml - would love to know if you have any opinion on whether default parameter values (like we use in ProduceAsync methods) should be avoided? BCL doesn't use them as far as I can tell, but we came to the conclusion that it was ok.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Poll.cs
mhowlett: You're not the first to interrogate me on that :-)
Sometimes I think this naming convention helps a lot in grouping like things together.
Particularly when names get long (as @tomasdeml  is suggesting above) it can help a lot.
In the past I've reluctantly conceded I'm wrong but in my own projects I still use it all the time without hesitation.

mhowlett: And yeah, I've heard of namespaces and folders. 

We're starting to get enough tests that we need to start structuring them a bit more. I'll try to avoid doing this.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Poll.cs
mhowlett: actually, it's asserted in `OnPartitionAssign` :-)
but I'll change this (and other tests) so they work in the more general case.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Poll.cs
mhowlett: yes

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Poll.cs
edenhill: Yeah, a sub-directory probably makes sense.

src/Confluent.Kafka/Consumer.cs
tomasdeml: Well, there are many opinions on default parameter values :). They entered the game quite late, so it may be one of the reasons they are not used in BCL. But I guess the most important reason may be potential [compile/runtime time issues](http://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/). E.g. if you make some param optional or change its default value, the callers may need to be recompiled. As these breaking changes can be quite subtle, IMO they may not be that useful in public APIs...

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Poll.cs
mhowlett: The entire test (and other like it) is built around the assumption that there is only one partition (start consuming from the first produced offset). For that reason, I want to leave this as is for now.

edenhill: Okay, that might not be optimal but will do for now, it is better if we can rely on topic auto creation and assume the topic has *at least* one partition instead.
Anyway, until then, can you clearly document this prerequirement? (preferably with the full kafka-topics.sh --create .. command to create it according to your wishes.

src/Confluent.Kafka/Consumer.cs
bschmidtbauer: I would be in favor of leveraging overloads instead of optional parameters, then you can add new signatures without potentially breaking existing clients. e.g. like @tomasdeml mentioned, if you add a second optional param it will break existing clients until they recompile.

mhowlett: thanks for the feedback @tomasdeml , @bschmidtbauer

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Background.cs
mhowlett: Agree with all the suggestions about renaming tests but I'm going to leave as is in this PR and do a test restructure at a later time (priority right now is to get the API locked down).

src/Confluent.Kafka/Consumer.cs
```diff
@@ -184,6 +186,11 @@ public void Stop()
             remove { consumer.OnPartitionEOF -= value; }
         }
 
+        public event EventHandler<AggregateException> OnBackgroundPollException;
```
mhowlett: i'm not sure this is the best name for this.

i've taken the approach that any exception that occurs on the background poll loop (e.g. in OnMessage etc) should be surfaced via an event on `Consumer`, currently called `OnBackgroundPollException`.

I think this works nicely and I can't think of any sane alternative, but welcome any feedback on this.

Note: had i thought through the complexities of correctly implementing a background poll loop on `Consumer`, i probably would have been in favor of omitting it from version 1...

test/Confluent.Kafka.IntegrationTests/Tests/AssignOverloads.cs
```diff
@@ -28,12 +28,12 @@ public static partial class Tests
...
         ///     Simple test of both Consumer.Assign overloads.
         /// </summary>
-        [Theory, MemberData(nameof(KafkaParameters))]
+        [Theory, ClassData(typeof(KafkaParameters))]
         public static void AssignOverloads(string bootstrapServers, string topic)
         {
             var consumerConfig = new Dictionary<string, object>
             {
-                { "group.id", "test-consumer-group" },
+                { "group.id", "assign-overloads-cg" },
```
mhowlett: every test now uses a different consumer group id because if a test fails, subsequent tests with the same consumer group will not be able to join the group because it was often still be alive from the failed test (session.timeout.ms min = 6000 is a relatively long time).

src/Confluent.Kafka/Consumer.cs
mhowlett: I'm currently thinking that background poll functionality would likely be better taken outside the `Consumer` class. Having different ways of operating a consumer (Start/Stop vs Poll or Consume) is not quite as straightforward an interface as I'd like. Now having  `OnBackgroundPollException` as well which is only relevant in the Start/Stop scenario, the situation worse. This is not really the end of the world, but I expect there is a better factoring.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_OperationalScenarios.cs
edenhill: I don't think there's a use-case for multiple start+stop cycles for a single instance. 
The only reason to have start() is to allow delegates to be registered before starting consumtion, right?

mhowlett: I don't think so either, but the API allows it so i think it needs to work...

edenhill: I disagree, solve it by documentation for now and enforce it in the future by exceptions.

mhowlett: If that behaviour is enforced, I think there shouldn't be a `.Stop` function at all (`.Dispose` completely allows for that intended usage). I'm happy with removing `.Stop`. If `.Stop` is retained, the behaviour as I have it is better I think as it is what is more implied by the API (less confusing).

test/Confluent.Kafka.IntegrationTests/Tests/AssignOverloads.cs
edenhill: For the future it is probably a good idea to use unique group ids for each invocation to have consistent tests (e.g., uuid or similar)

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_OperationalScenarios.cs
edenhill: While removing .Stop seems like a good idea we need to make sure that the behaviour of consumer_close() is properly abstracted:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L2311

E.g., it probably isnt good if Dispose() is blocking and triggers callbacks/delegates.

test/Confluent.Kafka.IntegrationTests/Tests/AssignOverloads.cs
mhowlett: good idea

mhowlett:
closing - background polling on the Consumer was cut due to uncertainties over the API and questions over whether it's necessary. We have no near term plans to re-include.

PR #41, opened 2017-02-09
ProduceAsync default params -> method overloads

mhowlett:
Advantages:
1. default parameter values suffer from issues outlined here: http://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/ also see @tomasdeml and @bschmidtbauer comments in #39
2. `Nullable` is avoided for the timestamp and partition parameters (no value is specified implicitly via the chosen overload). I think this is definitely a good interface in the partition case because at any particular point of calling ProduceAsync, the user is very unlikely to want to specify a partition some of the time and 'unassigned' the rest of the time. It's more likely they may want to specify null timestamps intermixed with actual timestamps, but I would still expect such scenarios to be very rare.

Disadvantages
1. Lots repetition (however, users of .NET BCLs are used to this, and the IDEs are designed for it - this is mostly an issue for the library maintainer).
2. I think we will probably have other parameters in the future. Specifying overloads for all combinations of 4 parameters is prohibitively verbose. We would need to start wrapping parameters up - introduce a `Payload` or `MessageValue` class that wraps the 'what' and have a variant of ProduceAsync that takes that.

TODO:
1. Some things are not quite right (eg RD_KAFKA_PARTITION_UA is duplicated) - I wanted to put a quick version out for discussion before putting too much effort in.
2. Possibly remove some variants.
3. Probably don't need `MessageValue` now, but maybe.

Commentary from everyone encouraged. I believe we probably want some variant of this PR.

Regarding ArraySegment? vs byte[], int, int - I both want to avoid `Nullable` in a public API and believe this would actually be more convenient in most practical situations.

### DIFF
--- a/src/Confluent.Kafka/Producer.cs
+++ b/src/Confluent.Kafka/Producer.cs
@@ -227,7 +227,7 @@ namespace Confluent.Kafka
             return;
         }
 
-        internal Task<Message> Produce(
+        private Task<Message> Produce(
             string topic,
             byte[] val, int valOffset, int valLength,
             byte[] key, int keyOffset, int keyLength,
--- a/src/Confluent.Kafka/Producer.cs
+++ b/src/Confluent.Kafka/Producer.cs
@@ -191,11 +191,11 @@ namespace Confluent.Kafka
             }
         }
 
-        private void Produce(
+        internal void Produce(
             string topic,
             byte[] val, int valOffset, int valLength,
             byte[] key, int keyOffset, int keyLength,
-            long? timestamp,
+            DateTime? timestamp,
             Int32 partition, bool blockIfQueueFull,
             IDeliveryHandler deliveryHandler)
         {
@@ -208,7 +208,7 @@ namespace Confluent.Kafka
                 var gch = GCHandle.Alloc(deliveryCompletionSource);
                 var ptr = GCHandle.ToIntPtr(gch);
 
-                if (topicHandle.Produce(val, valOffset, valLength, key, keyOffset, keyLength, partition, timestamp, ptr, blockIfQueueFull) != 0)
+                if (topicHandle.Produce(val, valOffset, valLength, key, keyOffset, keyLength, partition, timestamp?.Ticks, ptr, blockIfQueueFull) != 0)
                 {
                     var err = LibRdKafka.last_error();
                     gch.Free();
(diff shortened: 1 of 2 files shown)


src/Confluent.Kafka/Producer.cs
```diff
@@ -37,7 +37,7 @@ public class Producer : IDisposable
         private bool manualPoll;
         private bool disableDeliveryReports;
 
-        const int RD_KAFKA_PARTITION_UA = -1;
+        internal const int RD_KAFKA_PARTITION_UA = -1;
```
mhowlett: internal because `Producer<TKey, TValue>` needs to know this. could possibly be better defined somewhere else (though I couldn't think where). bit ugly, but implementation detail which isn't publicly exposed.

src/Confluent.Kafka/Producer.cs
```diff
@@ -191,11 +191,11 @@ public void HandleDeliveryReport(Message message)
             }
         }
 
-        private void Produce(
+        internal void Produce(
```
mhowlett: again a bit ugly that this is exposed internally and accessed directly by `Produce<TKey, TValue>`. But having `DateTime` nullable leads to a more convenient implementation (and 1 less level of indirection).

mhowlett:
wonderful

src/Confluent.Kafka/Producer.cs
```diff
@@ -510,13 +551,37 @@ public void HandleDeliveryReport(Message message)
             }
         }
 
-        public Task<Message<TKey, TValue>> ProduceAsync(string topic, TKey key, TValue val, DateTime? timestamp = null, int? partition = null, bool blockIfQueueFull = true)
+        private Task<Message<TKey, TValue>> Produce(string topic, TKey key, TValue val, DateTime? timestamp, int partition, bool blockIfQueueFull)
         {
             var handler = new TypedTaskDeliveryHandlerShim(key, val);
-            producer.ProduceAsync(topic, KeySerializer?.Serialize(key), ValueSerializer?.Serialize(val), handler, timestamp, partition, blockIfQueueFull);
+            var keyBytes = KeySerializer?.Serialize(key);
```
tomasdeml: Sorry for chiming in so late, but this statements seems like a bug to me. The `keyBytes` will be set to null when the key serializer is `null`. Ditto for `valBytes`. Is this intentional? If the serializer cannot ever be set to null, I think the elvis operator should be removed.

src/Confluent.Kafka/Producer.cs
```diff
@@ -548,12 +613,35 @@ public void HandleDeliveryReport(Message message)
             }
         }
 
-        public void ProduceAsync(string topic, TKey key, TValue val, IDeliveryHandler<TKey, TValue> deliveryHandler, DateTime? timestamp = null, int? partition = null, bool blockIfQueueFull = true)
+        private void Produce(string topic, TKey key, TValue val, IDeliveryHandler<TKey, TValue> deliveryHandler, DateTime? timestamp, int partition, bool blockIfQueueFull)
         {
             var handler = new TypedDeliveryHandlerShim(key, val, deliveryHandler);
-            producer.ProduceAsync(topic, KeySerializer?.Serialize(key), ValueSerializer?.Serialize(val), handler, timestamp, partition, blockIfQueueFull);
+            var keyBytes = KeySerializer?.Serialize(key);
```
tomasdeml: Ditto.

mhowlett: hi @tomasdeml, thanks for pointing that out.. 

the last few PRs were pushed without sufficient testing, sorry about this. FWIW this is fixed in the prodtest branch.

note that 0.9.2.x (currently == master) is what we intend to release (+ bug fixes). we are now completely dedicated to testing.

tomasdeml: Oh, ok :). Please see my comment https://github.com/confluentinc/confluent-kafka-dotnet/commit/10ee94437b13e9e5057966345737941f3235a3b3#commitcomment-20891385.

tomasdeml: Sorry for confusion, please see my comment https://github.com/confluentinc/confluent-kafka-dotnet/commit/10ee94437b13e9e5057966345737941f3235a3b3#commitcomment-20891497.

PR #42, opened 2017-02-10
How to think about Consumers and Producers [Experiment]

mhowlett:
It might be good if general client functions were split out from `Producer` and `Consumer` as in this PR creating three classes with more focus than the existing two.

this is a quick experiment (not polished, some stuff broken, don't merge) - just putting it out there for general comment.

I'm on the fence on this one. The way things are bugs me a bit, but I don't think it's going to cause anyone too many problems.

I don't know what to call `Client`/`ClientInstance`.

### DIFF
--- a/examples/Misc/Program.cs
+++ b/examples/Misc/Program.cs
@@ -34,7 +34,7 @@ namespace Confluent.Kafka.Examples.Misc
 
             using (var producer = new Producer(config))
             {
-                var groups = producer.ListGroups(TimeSpan.FromSeconds(10));
+                var groups = producer.ClientInstance.ListGroups(TimeSpan.FromSeconds(10));
                 Console.WriteLine($"Consumer Groups:");
                 foreach (var g in groups)
                 {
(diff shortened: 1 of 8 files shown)


examples/Misc/Program.cs
```diff
@@ -34,7 +34,7 @@ static void ListGroups(string brokerList)
 
             using (var producer = new Producer(config))
             {
-                var groups = producer.ListGroups(TimeSpan.FromSeconds(10));
+                var groups = producer.ClientInstance.ListGroups(TimeSpan.FromSeconds(10));
```
edenhill: I really dont like this, it looks like internal implementation details bleeding through to the public API.
From a user's perspective I'd much rather have those methods on my Producer instance, as is today, but using a base-class to avoid the repetetiveness.
This would also make it easier to make integrate with the library since you could make a method that takes a Client instance, rather than one for Producer or one for Consumer.

mhowlett: I agree this approach is worse than what we have, but I just wanted to have something related to this up for general comment (will close this PR out at some point). Maybe there is some other way of thinking about this.

With re-sharper enabled (not sure if not), VS shows direct methods of a class in bold in intellisense and methods on base classes in normal weight. So deriving from a common base class to get these methods does go some way towards helping.

Note though that `Consumer<TKey,TValue>` doesn't derive from `Consumer` (because otherwise it would expose the `byte[]` methods which is not desirable. Same for `Producer`.

mhowlett:
Closing (this was only ever meant to generate discussion).

I have related refactoring idea which is better and I'll put that up for discussion at some point.

PR #45, opened 2017-02-13
structs -> classes

mhowlett:
`struct` is currently used in many places where `class` is more appropriate ( https://msdn.microsoft.com/en-us/library/ms229017.aspx ).

The most important instance of this (because of frequency of use) is `Message` and `Message<TKey, TValue>`. I wanted to do some profiling before making the change because I thought `struct` may have some perf benefit here (even if it's use is not advised). I ran the benchmark test using both a class and struct variant of `Message` / `Message<TKey,TValue>` and profiled with dotTrace. Quick and dirty, but give some indication.

Averaged over 10 runs with N=5M, the mean run time for the class variant was 80457ms (std: 3134)
Averaged over 10 runs with N=5M, the mean run time for the struct variant was 81690ms (std: 2065)

Time spent in GC:
11.34%	+-1.03% (class)
11.35%	+-1.10% (struct)

the difference isn't statistically significant.


mhowlett:
wonderful

PR #46, opened 2017-02-14
remove background poll mode from consumer

mhowlett:
due to questions over design of this feature, and whether it's really needed, cutting.


mhowlett:
LGTM

PR #47, opened 2017-02-14
not subscribed, don't commit offsets

mhowlett:
this blocks indefinitely causing tests to hang.
I believe that is actually intended operation given current implementation.


mhowlett:
LGTM

PR #51, opened 2017-02-16
Target netstandard

treziac:
A library should target netstandard, not netcoreapp
Also updated dependencies version


mhowlett:
thanks @treziac 
hmm. I actually caught the netstandard -> netcoreapp issue before, I'll investigate how it managed to get through.

treziac:
Seems to be in Kafkatest client [WIP]
Also I didn't push it, but a sdk property in global.json needs to be adde when any sdk msbuild (preview3+) is installed. I assume project will migrate to csproj when release come in march

mhowlett:
FYI: I'm going to open a new PR on 0.9.4.x with only the netcoreapp -> netstandard one line change and leave this PR open on master.

An thanks for your additional notes. We're waiting for MS on the build stuff before putting effort into fixing all this up. Also note we haven't been developing under Windows (just building and testing the nuget package).

treziac:
fixed in #57, will eventually create a new one for sln/csproj when vs2017 is out

PR #63, opened 2017-02-19
API docs

mhowlett:
opening this PR early to highlight API changes.
continuing to work on docs (not for review yet).


mhowlett:
Summary of all changes made that don't relate to documentation:

- Consume errors were previously reported via the same event as other Errors. This is bad because it throws away context (topic, partition etc). Consume errors now have their own event: OnConsumeError.
- The non-deserialising consumer field in the deserialising Consumer is now private (previously protected -> errantly publicly exposed).
- Removed the Flush method from Consumer (which contained an empty implementation - this was a TODO, but in fact we never want this method implemented).
- Removed the OutQueueLength property from Producer and Consumer which exposes internal implementation details of librdkafka which is not a good thing to do on a public API. 
- Changed the return type of Producer.Flush to int (long is not required so a bad interface - may lead to users wondering why).
- Removed Producer.PartitionsAvailable (which should never be used except inside custom partitioners, but we don't have custom partitioners yet).
- Producer.Poll now returns the number of events served rather than void. feedback on this change requested. 
- Removed GetWatermarkOffsets from Producer as it will never return anything other than Invalid/Invalid. It's only useful on Consumers.
- Added a Producer constructor overload in favour of using default arguments (see reasoning elsewhere why default arguments are best avoided).
- Updated projects referenced by the .sln file so that it is no longer broken.

src/Confluent.Kafka/Metadata.cs
```diff
@@ -42,15 +57,17 @@ public Metadata(List<BrokerMetadata> brokers, List<TopicMetadata> topics, int or
         public List<BrokerMetadata> Brokers { get; }
 
         /// <summary>
-        ///     Information about every topic managed by in the cluster.
+        ///     Information about each topic in the cluster.
```
edenhill: Since this might be all topics, single requested topic, or locally used topics, this should probably say:
Information about requested topics in cluster

src/Confluent.Kafka/Metadata.cs
```diff
@@ -28,6 +28,21 @@ namespace Confluent.Kafka
...
     public class Metadata
     {
+        /// <summary>
+        ///     Instantiates a new Metadata class instance.
+        /// </summary>
+        /// <param name="brokers">
+        ///     Information about each constituent broker of the cluster.
+        /// </param>
+        /// <param name="topics">
+        ///     Information about each topic in the cluster.
```
edenhill: see comment below

src/Confluent.Kafka/Offset.cs
```diff
@@ -19,27 +19,63 @@
...
         public static Offset End { get { return new Offset(RD_KAFKA_OFFSET_END); } }
+
+        /// <summary>
+        ///     A special value thet refers to the stored offset for a partition.
+        /// </summary>
+        /// <returns></returns>
         public static Offset Stored { get { return new Offset(RD_KAFKA_OFFSET_STORED); } }
+
+        /// <summary>
+        ///     A special value that refers to an invalid or unassigned partition offset.
```
edenhill: invalid, unassigned or default

probably covers most cases

src/Confluent.Kafka/TimestampType.cs
```diff
@@ -16,6 +16,9 @@
 
 namespace Confluent.Kafka
 {
+    /// <summary>
+    ///     Enumerates the different meanings of a message timestamp value.
```
edenhill: The enums themselves should be documented too

src/Confluent.Kafka/WatermarkOffsets.cs
```diff
@@ -33,7 +33,7 @@ public class WatermarkOffsets
         ///     The offset of the earlist message in the topic/partition.
         /// </param>
         /// <param name="high">
-        ///     The offset of the last committed message in the topic/partition.
+        ///     The offset of the last stored message in the topic/partition.
```
edenhill: committed is actually more accurate since store sometimes refers to the local in-memory offset-to-commit-store

src/Confluent.Kafka/WatermarkOffsets.cs
```diff
@@ -33,7 +33,7 @@ public class WatermarkOffsets
...
         /// </param>
         public WatermarkOffsets(Offset low, Offset high)
```
edenhill: is this Query or Get?
Are they both exposed?

src/Confluent.Kafka/WatermarkOffsets.cs
```diff
@@ -47,7 +47,7 @@ public WatermarkOffsets(Offset low, Offset high)
         public Offset Low { get; }
 
         /// <summary>
-        ///     The offset of the last committed message in the topic/partition.
+        ///     The offset of the last stored message in the topic/partition.
```
edenhill: dito

src/Confluent.Kafka/Metadata.cs
mhowlett: yep

src/Confluent.Kafka/Metadata.cs
mhowlett: yep

src/Confluent.Kafka/Offset.cs
mhowlett: yep

src/Confluent.Kafka/WatermarkOffsets.cs
mhowlett: this is a general data structure used to store a low and high watermark offset values. it's returned by both the get and query methods.

src/Confluent.Kafka/TimestampType.cs
mhowlett: so should a lot of other things :-)
(but will document)

src/Confluent.Kafka/WatermarkOffsets.cs
edenhill: oh right, sorry, I Was confusing things. nevermind me!

src/Confluent.Kafka/WatermarkOffsets.cs
mhowlett: yeah, I think so too but someone was telling me they thought stored was better because of possible confusion of the word committed with committing offsets. I'll change it back to committed for now.

edenhill: No, dont! I was in the wrong, the watermark offsets have nothing to do with committed offsets. they are just the oldest and newest message offsets of a partition.

src/Confluent.Kafka/WatermarkOffsets.cs
mhowlett: yep

src/Confluent.Kafka/WatermarkOffsets.cs
mhowlett: probably should give your jet lag a break and go to bed! 

the initial design docs for this actually use the word committed in relation to high watermark.

Confluent.Kafka.sln
```diff
@@ -1,101 +1,101 @@
...
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "SimpleProducer", "examples\SimpleProducer\SimpleProducer.xproj", "{A7BF0A75-D3E7-4024-8597-5FCCC567D372}"
+EndProject
+Project("{8BB2217D-0F2D-49D1-97BC-3654ED321F3B}") = "SimpleConsumer", "examples\SimpleConsumer\SimpleConsumer.xproj", "{85F08EF0-9E30-49D3-B86F-DBA36267843C}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{56321620-6DE2-44DC-942B-E3701C01EBFE}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "test", "test", "{03D93818-5A54-4913-A5CD-CA0C373CEE05}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "examples", "examples", "{5E1F754C-FCE9-4B80-B13D-5F9ACE13305A}"
+EndProject
```
edenhill: ayay, whitespace diff.
I think you should set your git client to auto-convert Windows line-endings to Unix. see auto crlf.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -316,13 +319,13 @@ public void Poll()
         public event EventHandler<Message<TKey, TValue>> OnMessage;
 
         /// <summary>
-        ///     Returns the current partition assignment as set by Assign.
+        ///     The current partition assignment as set by Assign.
```
edenhill: Nitpick but I think it is clearer with the Returns the  prefix.
Alternatively just use the returns tag

src/Confluent.Kafka/Error.cs
```diff
@@ -20,39 +20,76 @@ namespace Confluent.Kafka
...
         }
 
+        /// <summary>
+        ///     true if Code != ErrorCode.NoError.
+        /// </summary>
         public bool HasError
             => Code != ErrorCode.NoError;
 
+        /// <summary>
+        ///     true if this is error originated locally (within librdkafka). false otherwise.
```
edenhill: ", false otherwise", not ". false .."

src/Confluent.Kafka/Error.cs
```diff
@@ -20,39 +20,76 @@ namespace Confluent.Kafka
...
+        /// </summary>
         public bool IsLocalError
             => (int)Code < -1;
 
+        /// <summary>
+        ///     true if this error originated on a broker. false otherwise.
```
edenhill: comma, not period

src/Confluent.Kafka/IDeliveryHandler.cs
```diff
@@ -21,9 +21,8 @@ namespace Confluent.Kafka
 {
     /// <summary>
     ///     This interface is implmented by types that handle delivery report
```
edenhill: implemented

src/Confluent.Kafka/IDeliveryHandler.cs
```diff
@@ -51,9 +50,8 @@ public interface IDeliveryHandler
 
     /// <summary>
     ///     This interface is implmented by types that handle delivery report
```
edenhill: implemented

src/Confluent.Kafka/Producer.cs
```diff
@@ -246,8 +247,8 @@ public void HandleDeliveryReport(Message message)
         ///     Initializes a new Producer instance.
         /// </summary>
         /// <param name="config">
-        ///     librdkafka configuration parameters (refer to https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md)
-        ///     topic configuration parameters should be specified via the "default.topic.config" config parameter.
+        ///     librdkafka configuration parameters (refer to https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md).
+        ///     Topic configuration parameters should be specified via the "default.topic.config" config parameter.
```
edenhill: "must be" or "are", "not should be".
Also clarify the default.topic.config, something like:
"via the default.topi... sub-dictionary"

src/Confluent.Kafka/Producer.cs
```diff
@@ -302,16 +303,17 @@ public Producer(IEnumerable<KeyValuePair<string, object>> config, bool manualPol
...
```
edenhill: dito

src/Confluent.Kafka/Producer.cs
```diff
@@ -302,16 +303,17 @@ public Producer(IEnumerable<KeyValuePair<string, object>> config, bool manualPol
...
         /// </param>
         public Producer(IEnumerable<KeyValuePair<string, object>> config)
             : this(config, false, false) {}
 
 
         /// <summary>
-        ///     Poll for callback events. Only use this method on producer
-        ///     instances where background polling has been disabled.
+        ///     Poll for callback events. You will not typically need
```
edenhill: Why was this changed?
Is there a case where you should call Poll() with background polling enabled?

src/Confluent.Kafka/Producer.cs
```diff
@@ -330,8 +332,9 @@ public int Poll(int millisecondsTimeout)
         }
 
         /// <summary>
-        ///     Poll for callback events. Only use this method on producer
-        ///     instances where background polling has been disabled.
+        ///     Poll for callback events. You will not typically need
+        ///     to call this method. Only call on producer instances
```
edenhill: dito.

src/Confluent.Kafka/Producer.cs
```diff
@@ -341,12 +344,13 @@ public int Poll(int millisecondsTimeout)
...
         public int Poll(TimeSpan timeout)
-            => Poll(timeout.TotalMillisecondsAsInt());
-
+            => Poll(timeout.TotalMillisecondsAsInt());
+
         /// <summary>
-        ///     Poll for callback events. Only use this method on producer
-        ///     instances where background polling is disabled. Blocks
-        ///     until there is a callback event ready to be served.
+        ///     Poll for callback events. You will not typically need
```
edenhill: dito.

another thing: isnt it better to avoid all this doc duplication and simply reference the first one?

src/Confluent.Kafka/Producer.cs
```diff
@@ -903,8 +912,8 @@ public class Producer<TKey, TValue> : ISerializingProducer<TKey, TValue>, IDispo
         ///     Creates a new Producer instance.
         /// </summary>
         /// <param name="config">
-        ///     librdkafka configuration parameters (refer to https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md)
-        ///     topic configuration parameters should be specified via the "default.topic.config" config parameter.
+        ///     librdkafka configuration parameters (refer to https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md).
+        ///     Topic configuration parameters should be specified via the "default.topic.config" config parameter.
```
edenhill: dito

src/Confluent.Kafka/Producer.cs
```diff
@@ -938,8 +947,8 @@ public class Producer<TKey, TValue> : ISerializingProducer<TKey, TValue>, IDispo
         ///     Initializes a new Producer instance.
         /// </summary>
         /// <param name="config">
-        ///     librdkafka configuration parameters (refer to https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md)
-        ///     topic configuration parameters should be specified via the "default.topic.config" config parameter.
+        ///     librdkafka configuration parameters (refer to https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md).
+        ///     Topic configuration parameters should be specified via the "default.topic.config" config parameter.
```
edenhill: dito

src/Confluent.Kafka/Producer.cs
```diff
@@ -1034,25 +1043,58 @@ public string Name
...
+
+
+        /// <summary>
+        ///     Asynchronously send a single message to the broker (order of delivery reports strictly guarenteed).
+        ///     See <see cref="ProduceAsync(string, TKey, TValue, int, bool, IDeliveryHandler{TKey, TValue})"/> for more information.
+        /// </summary>
+        /// <remarks>
+        ///     The partition the message is produced to is determined using the configured partitioner.
+        ///     
+        ///     Blocks if the send queue is full.
```
edenhill: Requires background polling to be enabled, or a separate thread doing polll, otherwise it will block indefinately.

src/Confluent.Kafka/Producer.cs
```diff
@@ -1034,25 +1043,58 @@ public string Name
...
+        /// </remarks>
         public void ProduceAsync(string topic, TKey key, TValue val, IDeliveryHandler<TKey, TValue> deliveryHandler)
-            => serializingProducer.ProduceAsync(topic, key, val, deliveryHandler);
-
+            => serializingProducer.ProduceAsync(topic, key, val, deliveryHandler);
+
+        /// <summary>
+        ///     Asynchronously send a single message to the broker (order of delivery reports strictly guarenteed).
```
edenhill: dito

src/Confluent.Kafka/Producer.cs
```diff
@@ -1034,25 +1043,58 @@ public string Name
...
         /// </remarks>
         public Task<Message<TKey, TValue>> ProduceAsync(string topic, TKey key, TValue val, bool blockIfQueueFull)
-            => serializingProducer.ProduceAsync(topic, key, val, blockIfQueueFull);
-
-
+            => serializingProducer.ProduceAsync(topic, key, val, blockIfQueueFull);
+
+
+        /// <summary>
+        ///     Asynchronously send a single message to the broker (order of delivery reports strictly guarenteed).
```
edenhill: remove (order of delivery reports...) part since it is better explained in the see cref link

src/Confluent.Kafka/Producer.cs
```diff
@@ -1034,25 +1043,58 @@ public string Name
...
+        ///     Refer to <see cref="ProduceAsync(string, TKey, TValue, int, bool)" />
+        ///     for more information.
+        /// </remarks>
         public void ProduceAsync(string topic, TKey key, TValue val, int partition, bool blockIfQueueFull, IDeliveryHandler<TKey, TValue> deliveryHandler)
-            => serializingProducer.ProduceAsync(topic, key, val, partition, blockIfQueueFull, deliveryHandler);
-
+            => serializingProducer.ProduceAsync(topic, key, val, partition, blockIfQueueFull, deliveryHandler);
+
+        /// <summary>
+        ///     Asynchronously send a single message to the broker (order of delivery reports strictly guarenteed).
```
edenhill: dito

src/Confluent.Kafka/Producer.cs
```diff
@@ -1034,25 +1043,58 @@ public string Name
...
+        /// <remarks>
+        ///     Blocks if the send queue is full.
+        /// </remarks>
         public void ProduceAsync(string topic, TKey key, TValue val, int partition, IDeliveryHandler<TKey, TValue> deliveryHandler)
-            => serializingProducer.ProduceAsync(topic, key, val, partition, deliveryHandler);
-
+            => serializingProducer.ProduceAsync(topic, key, val, partition, deliveryHandler);
+
+        /// <summary>
+        ///     Asynchronously send a single message to the broker (order of delivery reports strictly guarenteed).
```
edenhill: dito

src/Confluent.Kafka/Serialization/IDeserializer.cs
```diff
@@ -16,8 +16,23 @@
 
 namespace Confluent.Kafka.Serialization
 {
+    /// <summary>
+    ///     Implement this interface to define a de-serializer
```
edenhill: deserializer without -

src/Confluent.Kafka/Timestamp.cs
```diff
@@ -80,6 +88,15 @@ public static long DateTimeToUnixTimestampMs(DateTime dateTime)
...
+        ///     Convert a unix timestamp to a DateTime value.
+        /// </summary>
+        /// <param name="timestamp">
+        ///     The unix timestamp to convert.
+        /// </param>
+        /// <returns>
+        ///     The DateTime value associated with <paramref name="timestamp"/>
+        /// </returns>
         public static DateTime UnixTimestampMsToDateTime(long timestamp)
             => new DateTime(1970, 1, 1, 0, 0, 0, 0, System.DateTimeKind.Unspecified)
```
edenhill: do we have a unit-test that makes sure this is correct, mainly with regards to timezones?

treziac: Not doc related, bu this should be Utc kind - UnixTimestamp is Utc, not Unspecified.
Also i saw that producer use timestamp?.Ticks instead of DateTimeToUnixTimestampMs(datetime) in Produce, will post an issue/PR related to this tomorrow

treziac: Regarding TimeZone, perhaps we should use DateTimeOffset instead? Note that it already contains a method for From/ToUnixTimestampMs and can be converted from / to datetime

src/Confluent.Kafka/Timestamp.cs
```diff
@@ -72,6 +71,15 @@ public override int GetHashCode()
         public static bool operator !=(Timestamp a, Timestamp b)
             => !(a == b);
 
+        /// <summary>
+        ///     Convert a DateTime instance to a unix timestamp.
```
treziac: It's already in the name of function, but maybe specify it is millisecond and not second. Also precise in remarks that it converts in Utc before - local DateTime with unspecified Kind will be false

src/Confluent.Kafka/Timestamp.cs
```diff
@@ -72,6 +71,15 @@ public override int GetHashCode()
...
             => !(a == b);
 
+        /// <summary>
+        ///     Convert a DateTime instance to a unix timestamp.
+        /// </summary>
+        /// <param name="dateTime">
+        ///     The DateTime value to convert.
+        /// </param>
+        /// <returns>
+        ///     The unix timestamp corresponding to <paramref name="dateTime"/>
```
treziac: unix millisecond timestamp

src/Confluent.Kafka/Consumer.cs
mhowlett: Returns is not the best terminology for a property... but now you're made me think about it, properties are usually documented as "Gets ..." or "Gets and Sets ..." - I'll go through and change a lot of them over.

src/Confluent.Kafka/Error.cs
mhowlett: yep

src/Confluent.Kafka/Error.cs
mhowlett: fixed

src/Confluent.Kafka/IDeliveryHandler.cs
mhowlett: fixed

src/Confluent.Kafka/IDeliveryHandler.cs
mhowlett: fixed

src/Confluent.Kafka/Timestamp.cs
mhowlett: well spotted. 

btw, we're hiring at Confluent. If you're interested in an interview, shoot me an email.

src/Confluent.Kafka/Producer.cs
mhowlett: yep. this also deserves an inline example i think, but later.

src/Confluent.Kafka/Producer.cs
mhowlett: yep

src/Confluent.Kafka/Producer.cs
mhowlett: This hasn't changed - it's just clarifying. Background polling has always been enabled by default on the producer and i was thinking disabling it is an advanced use case. 

I guess you might want to poll from multiple threads to process delivery reports in parallel.. but how often is anyone going to want to do this? is it often going to be a bottleneck? I hadn't thought this through before. I think it would all work ok with the current API. we should sync up about this to discuss.

mhowlett: Assuming there are no foreseeable implications to API, I don't think think this should be changed right now.

src/Confluent.Kafka/Producer.cs
mhowlett: fixed

src/Confluent.Kafka/Producer.cs
mhowlett: fixed

src/Confluent.Kafka/Producer.cs
mhowlett: i've just discovered you can reference text blocks in an external file - we should probably use this approach where it makes sense.

it's very desirable to have the docs on every method from an intellisense point of view at least. but yes the duplication is highly undesirable, I don't like it.

src/Confluent.Kafka/Producer.cs
mhowlett: The summary is what gets displayed in intellisense - this note is there because I was thinking a user should see it in that. Ideally (ultimately) we don't want the see cref thing at all (see above note about including blocks).

src/Confluent.Kafka/Producer.cs
mhowlett: yep. noted in the fully documented variant as well.

src/Confluent.Kafka/Producer.cs
mhowlett: fixed.

src/Confluent.Kafka/Producer.cs
mhowlett: fixed

src/Confluent.Kafka/Producer.cs
mhowlett: fixed

src/Confluent.Kafka/Serialization/IDeserializer.cs
mhowlett: yep.

src/Confluent.Kafka/Timestamp.cs
mhowlett: @edenhill I was considering timestamps reliant on the move to 0.9.4 (remembering the bug in 0.9.2 i mentioned), so had been putting off testing until now. Will do some unit an integration tests on master (master).

@treziac I'm not convinced by DateTimeOffset - I think we should keep everything as DateTime - DateTimeOffset seems a more complex format than strictly necessary.

src/Confluent.Kafka/Timestamp.cs
mhowlett: good points

src/Confluent.Kafka/Timestamp.cs
mhowlett: yep.

src/Confluent.Kafka/Timestamp.cs
mhowlett: Oh, we've cut explicit production of timestamps in the upcoming release, so that also goes some way to explaining why there is no testing for it yet.

mhowlett: I corrected the produce method as you pointed out, though this won't come into play just yet as we have disabled all variants of ProduceAsync that allow a timestamp to be explicitly specified in the upcoming release (because of lack of time and there is stuff around this to be tested, including the marshalling to producev which is definitely wrong). Also, we will move away from long? for timestamps because timestamp of 0 is interpreted by librdkafka to mean 'current time', the reason I introduced null in the first place.

Also fixed Unspecified -> Utc. Good point.

src/Confluent.Kafka/Producer.cs
```diff
@@ -425,42 +640,123 @@ public void Dispose()
...
+        /// </summary>
+        /// <param name="timeout">
+        ///     The maximum period of time the call should block.
+        /// </param>
         public List<GroupInfo> ListGroups(TimeSpan timeout)
             => kafkaHandle.ListGroups(timeout.TotalMillisecondsAsInt());
 
+        /// <summary>
+        ///     Get information pertaining to all groups in the Kafka
+        ///     cluster.
```
edenhill: Add not on indefinate blockage

src/Confluent.Kafka/Producer.cs
```diff
@@ -425,42 +640,123 @@ public void Dispose()
...
+        ///     Returns information pertaining to the specified group
+        ///     or null if this group does not exist.
+        /// </returns>
         public GroupInfo ListGroup(string group, TimeSpan timeout)
             => kafkaHandle.ListGroup(group, timeout.TotalMillisecondsAsInt());
 
+        /// <summary>
+        ///     Get information pertaining to a particular group in the
+        ///     Kafka cluster.
+        ///
```
edenhill: ditos

src/Confluent.Kafka/Producer.cs
```diff
@@ -425,42 +640,123 @@ public void Dispose()
...
+        ///     Query the Kafka cluster for low (oldest/beginning) and high (newest/end)
+        ///     offsets for the specified topic/partition.
+        ///
+        ///     [UNSTABLE-API] - The API associated with this functionality is subject to change.
+        /// </summary>
+        /// <param name="topicPartition">
+        ///     The topic/partition of interest.
+        /// </param>
+        /// <param name="timeout">
+        ///     The maximum period of time the call should block.
```
edenhill: may block

src/Confluent.Kafka/Producer.cs
```diff
@@ -425,42 +640,123 @@ public void Dispose()
...
+        /// </param>
+        /// <returns>
+        ///     The requested WatermarkOffsets.
```
edenhill: no documentation on raised exceptions  on failure?

src/Confluent.Kafka/Producer.cs
```diff
@@ -425,42 +640,123 @@ public void Dispose()
...
+        /// </returns>
         public WatermarkOffsets QueryWatermarkOffsets(TopicPartition topicPartition, TimeSpan timeout)
             => kafkaHandle.QueryWatermarkOffsets(topicPartition.Topic, topicPartition.Partition, timeout.TotalMillisecondsAsInt());
 
+        /// <summary>
+        ///     Query the Kafka cluster for low (oldest/beginning) and high (newest/end)
+        ///     offsets for the specified topic/partition.
+        ///
+        ///     [UNSTABLE-API] - The API associated with this functionality is subject to change.
```
edenhill: indefinate blockage

src/Confluent.Kafka/Producer.cs
```diff
@@ -425,42 +640,123 @@ public void Dispose()
...
-
-
         private Metadata GetMetadata(bool allTopics, string topic, int millisecondsTimeout)
             => kafkaHandle.GetMetadata(allTopics, topic == null ? null : getKafkaTopicHandle(topic), millisecondsTimeout);
 
+        /// <summary>
+        ///     Query the cluster for metadata
+        ///
+        ///     - allTopics = true - request all topics from cluster
+        ///     - allTopics = false, topic = null - request only locally known topics (topic_new():ed topics or otherwise locally referenced once, such as consumed topics)
```
edenhill: topic_new() doesn't have any relevance in the dotnet client

src/Confluent.Kafka/Producer.cs
```diff
@@ -635,93 +957,364 @@ public class Producer<TKey, TValue> : ISerializingProducer<TKey, TValue>, IDispo
...
+        /// </param>
+        /// <param name="key">
+        ///     the message key (possibly null if allowed by the key serializer).
+        /// </param>
+        /// <param name="val">
+        ///     the message value (possibly null if allowed by the value serializer).
+        /// </param>
+        /// <param name="blockIfQueueFull">
+        ///     Whether or not to block if the send queue is full.
+        ///     If false, and exception will be thrown in the event.
```
edenhill: an exception. But which one? (QUEUE_FULL)

src/Confluent.Kafka/Producer.cs
```diff
@@ -635,93 +957,364 @@ public class Producer<TKey, TValue> : ISerializingProducer<TKey, TValue>, IDispo
...
+        ///      
+        ///     Note: blockIfQueueFull is set to true, background polling is
```
edenhill: warning

src/Confluent.Kafka/Producer.cs
```diff
@@ -635,93 +957,364 @@ public class Producer<TKey, TValue> : ISerializingProducer<TKey, TValue>, IDispo
...
+        /// </remarks>
         public Task<Message<TKey, TValue>> ProduceAsync(string topic, TKey key, TValue val, int partition, bool blockIfQueueFull)
             => serializingProducer.ProduceAsync(topic, key, val, partition, blockIfQueueFull);
 
+        /// <summary>
+        ///     Asynchronously send a single message to the broker.
+        ///     Refer to <see cref="ProduceAsync(string, TKey, TValue, int, bool)" /> for more information.
+        /// </summary>
+        /// <remarks>
+        ///     Blocks if the send queue is full. Note: if background polling is disabled and Poll is
```
edenhill: warning

src/Confluent.Kafka/Producer.cs
```diff
@@ -635,93 +957,364 @@ public class Producer<TKey, TValue> : ISerializingProducer<TKey, TValue>, IDispo
...
+        /// </remarks>
         public void ProduceAsync(string topic, TKey key, TValue val, int partition, bool blockIfQueueFull, IDeliveryHandler<TKey, TValue> deliveryHandler)
             => serializingProducer.ProduceAsync(topic, key, val, partition, blockIfQueueFull, deliveryHandler);
 
+        /// <summary>
+        ///     Asynchronously send a single message to the broker (order of delivery reports strictly guarenteed).
+        ///     See <see cref="ProduceAsync(string, TKey, TValue, int, bool, IDeliveryHandler{TKey, TValue})"/> for more information.
+        /// </summary>
+        /// <remarks>
+        ///     Blocks if the send queue is full. Note: if background polling is disabled and Poll is
```
edenhill: warning

src/Confluent.Kafka/Producer.cs
```diff
@@ -635,93 +957,364 @@ public class Producer<TKey, TValue> : ISerializingProducer<TKey, TValue>, IDispo
...
         public void ProduceAsync(string topic, TKey key, TValue val, bool blockIfQueueFull, IDeliveryHandler<TKey, TValue> deliveryHandler)
             => serializingProducer.ProduceAsync(topic, key, val, blockIfQueueFull, deliveryHandler);
 
 
+        /// <summary>
+        ///     Raised when there is information that should be logged.
+        /// </summary>
+        /// <remarks>
+        ///     You can specify the log level with the 'log_level' configuration
+        ///     property. You also probably want to specify one or more debug
```
edenhill: dito

src/Confluent.Kafka/Producer.cs
```diff
@@ -635,93 +957,364 @@ public class Producer<TKey, TValue> : ISerializingProducer<TKey, TValue>, IDispo
...
+        ///     contexts using the 'debug' configuration property.
+        ///
+        ///     This event can potentially be raised on any thread.
+        /// </remarks>
         public event EventHandler<LogMessage> OnLog;
 
+        /// <summary>
+        ///     Raised on librdkafka stats events.
```
edenhill: describe the parameter (one json document as defined by https://github.com/edenhill/librdkafka/wiki/Statistics)

src/Confluent.Kafka/CommittedOffsets.cs
```diff
@@ -21,6 +21,10 @@
 
 namespace Confluent.Kafka
 {
+    /// <summary>
+    ///     Encapsulates information provided to a Consumer's OnOffsetsCommitted
+    ///     event.
```
mhowlett: yeah.. i was trying to be quick, but now you put words into my mouth, added.

src/Confluent.Kafka/Loggers.cs
```diff
@@ -21,8 +21,14 @@
 
 namespace Confluent.Kafka
 {
+    /// <summary>
+    ///     OnLog callback event handler implementations.
```
mhowlett: righto, i put the warning here, though it's most appropriate on the actual OnLog events. It's noted on them already, but i'll make it an explicit warning.

delegate is called right away, not enqueued.

src/Confluent.Kafka/Loggers.cs
```diff
@@ -21,8 +21,14 @@
...
+    /// </summary>
     public static class Loggers
     {
+        /// <summary>
+        ///     The method used to log messages by default.
+        /// </summary>
         public static void ConsoleLogger(object obj, LogMessage logInfo)
         {
             var now = DateTime.Now.ToString("yyyy-MM-dd HH:mm:ss.fff");
```
mhowlett: good point. i'll track with a github issue.

mhowlett: though TBH locale specific date times annoy me and the one used here is neutral. that said, open to changing it. not bothering with a github issue.

src/Confluent.Kafka/Message.cs
```diff
@@ -18,6 +18,9 @@
 
 namespace Confluent.Kafka
 {
+    /// <summary>
+    ///     Represetns a (deserialized) Kafka Message - what, where and when.
```
mhowlett: what: key, value
where: topic, partition, offset
when: timestamp

I'll get rid of that for now and describe more fully in a later iteration.

typo fixed.

src/Confluent.Kafka/Offset.cs
```diff
@@ -18,25 +18,64 @@
...
             Value = offset;
         }
 
+        /// <summary>
+        ///     Gets the long value corresponding to this offset.
+        /// </summary>
         public long Value { get; }
 
+        /// <summary>
+        ///     Gets whether or not this is one of the four special
```
mhowlett: you're a thinker..

src/Confluent.Kafka/Producer.cs
```diff
@@ -241,19 +257,19 @@ public void HandleDeliveryReport(Message message)
...
         /// <param name="manualPoll">
         ///     If true, does not start a dedicated polling thread to trigger events or receive delivery reports -
-        ///     you must call the Poll method periodically instead.
+        ///     you must call the Poll method periodically instead. Typically you should set this parameter to false.
         /// </param>
         /// <param name="disableDeliveryReports">
-        ///     If true, disables notification of delivery reports. Note: if set to true and you use a ProduceAsync variant that return
-        ///     a Task, the Tasks will never complete. Generally you should leave this parameter as false. Set it to true for "fire and
+        ///     If true, disables delivery report notification. Note: if set to true and you use a ProduceAsync variant that returns
+        ///     a Task, the Tasks will never complete. Typically you should set this parameter to false. Set it to true for "fire and
```
mhowlett: I'll mark as unstable (you've probably noticed i like doing that).

BTW: unrelated: I've got another idea as to how to factor the general client functionality currently exposed on producers and consumers (and marked as unstable). you're going to love it. This time I actually like the idea too.

mhowlett: oh, i remember. this isn't exposed on `Producer<TKey,TValue>`, only `Producer`. And the entire `Producer` class is marked as unstable - though at this point i'm in favour of keeping it. It exposes two advanced use case things, yet keeps them out of the interface every one is going to use in practice.

src/Confluent.Kafka/Producer.cs
```diff
@@ -293,124 +309,323 @@ public Producer(IEnumerable<KeyValuePair<string, object>> config, bool manualPol
...
+        ///     librdkafka configuration parameters (refer to https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md).
+        ///     Topic configuration parameters are specified via the "default.topic.config" sub-dictionary config parameter.
+        /// </param>
+        public Producer(IEnumerable<KeyValuePair<string, object>> config)
+            : this(config, false, false) {}
+
 
-        public void Poll(int millisecondsTimeout)
+        /// <summary>
+        ///     Poll for callback events. You will not typically need
```
mhowlett: my terminology might not be quite right but will leave for now. delegate isn't right here i don't think.

src/Confluent.Kafka/Producer.cs
```diff
@@ -293,124 +309,323 @@ public Producer(IEnumerable<KeyValuePair<string, object>> config, bool manualPol
...
+        ///     to call this method. Only call on producer instances 
+        ///     where background polling has been disabled.
```
mhowlett: i guess 'has not been enabled' is better, changed.

src/Confluent.Kafka/Producer.cs
```diff
@@ -293,124 +309,323 @@ public Producer(IEnumerable<KeyValuePair<string, object>> config, bool manualPol
...
             }
-            this.kafkaHandle.Poll((IntPtr)millisecondsTimeout);
+            return this.kafkaHandle.Poll((IntPtr)millisecondsTimeout);
         }
 
-        public void Poll(TimeSpan timeout)
+        /// <summary>
+        ///     Poll for callback events. You will not typically need
+        ///     to call this method. Only call on producer instances 
+        ///     where background polling has been disabled.
```
mhowlett: yep.

src/Confluent.Kafka/Producer.cs
```diff
@@ -293,124 +309,323 @@ public Producer(IEnumerable<KeyValuePair<string, object>> config, bool manualPol
...
+        /// <returns>
+        ///     Returns the number of events served.
+        /// </returns>
+        public int Poll()
             => Poll(-1);
 
 
         /// <summary>
-        ///     Raised on critical errors, e.g. connection failures or all brokers down.
+        ///     Raised on critical errors, e.g. connection failures or all
```
mhowlett: thanks for the wording (changed in all 4 places)

src/Confluent.Kafka/Producer.cs
```diff
@@ -293,124 +309,323 @@ public Producer(IEnumerable<KeyValuePair<string, object>> config, bool manualPol
...
 
         /// <summary>
-        ///     Raised when there is information that should be to be logged.
+        ///     Raised when there is information that should be logged.
         /// </summary>
         /// <remarks>
-        ///     Specify which log level with the log_level configuration property.
-        ///     can happen on any thread.
+        ///     You can specify the log level with the 'log_level' configuration
+        ///     property. You also probably want to specify one or more debug
```
mhowlett: I'll add your comment that logging should not typically be enabled. However, i did want it and did find it useful at one point, and the need to specify a debug context tripped me up, so keen to leave that in.

mhowlett: (changed on all OnLog events)

src/Confluent.Kafka/Producer.cs
```diff
@@ -293,124 +309,323 @@ public Producer(IEnumerable<KeyValuePair<string, object>> config, bool manualPol
...
+        ///     for non-null values, the offset into the key array of the
+        ///     sub-array to use as the message key.
+        ///     if <paramref name="key" />  is null, keyOffset must be 0.
+        /// </param>
+        /// <param name="keyLength">
+        ///     for non-null keys, the length of the sequence of bytes that
+        ///     constitutes the key.
+        ///     if <paramref name="key" />  is null, keyOffset must be 0.
+        /// </param>
+        /// <param name="val">
```
mhowlett: good point

src/Confluent.Kafka/Producer.cs
```diff
@@ -293,124 +309,323 @@ public Producer(IEnumerable<KeyValuePair<string, object>> config, bool manualPol
...
+        /// <remarks>
+        ///     If you require strict ordering of delivery reports to be maintained, 
+        ///     you should use a variant of ProduceAsync that takes an IDeliveryHandler
+        ///     parameter, not a variant that returns a Task&lt;Message&gt;
+        /// </remarks>
         public Task<Message> ProduceAsync(string topic, byte[] key, int keyOffset, int keyLength, byte[] val, int valOffset, int valLength, int partition, bool blockIfQueueFull)
             => Produce(topic, val, valOffset, valLength, key, keyOffset, keyLength, null, partition, blockIfQueueFull);
 
+        /// <summary>
+        ///     Asynchronously send a single message to the broker.
```
mhowlett: because tasks are completed on the thread pool. will note.

src/Confluent.Kafka/Producer.cs
mhowlett: yep

src/Confluent.Kafka/Producer.cs
mhowlett: yep

src/Confluent.Kafka/Producer.cs
mhowlett: yep

mhowlett: (fixed in 12 places)

src/Confluent.Kafka/Producer.cs
mhowlett: here and other places. leaving now due to time constraints.

src/Confluent.Kafka/Producer.cs
mhowlett: ok. this is actually noted in the timeout parameter. but it's probably worth having here (ditto other places)

src/Confluent.Kafka/Producer.cs
mhowlett: yep.

src/Confluent.Kafka/Producer.cs
mhowlett: noted, in both copies of this comment.

src/Confluent.Kafka/Producer.cs
mhowlett: yep (twice)

src/Confluent.Kafka/Producer.cs
mhowlett: yep (4 times)

src/Confluent.Kafka/Producer.cs
mhowlett: yep

src/Confluent.Kafka/Producer.cs
mhowlett: changed this to still note log_level and debug, but dissuade people from using it.

src/Confluent.Kafka/Producer.cs
mhowlett: fixed (4 times)

src/Confluent.Kafka/Producer.cs
```diff
@@ -323,7 +323,7 @@ public Producer(IEnumerable<KeyValuePair<string, object>> config)
         /// <summary>
         ///     Poll for callback events. You will not typically need
         ///     to call this method. Only call on producer instances 
-        ///     where background polling has been disabled.
+        ///     where background polling is not enabled.
```
edenhill: I meant that we should reference the constructor argument here, e.g.:
"where manual polling has been enabled".

src/Confluent.Kafka/Consumer.cs
```diff
@@ -807,11 +820,16 @@ public Consumer(IEnumerable<KeyValuePair<string, object>> config)
         ///     Raised when there is information that should be logged.
         /// </summary>
         /// <remarks>
-        ///     You can specify the log level with the 'log_level' configuration
-        ///     property. You also probably want to specify one or more debug
-        ///     contexts using the 'debug' configuration property.
-        ///
-        ///     This event can potentially be raised on any thread.
+        ///     Note: By default not many log messages are generated.
```
edenhill: This is subjective, remove.

src/Confluent.Kafka/Producer.cs
```diff
@@ -487,9 +495,11 @@ public Task<Message> ProduceAsync(string topic, byte[] key, int keyOffset, int k
         /// </param>
         /// <param name="blockIfQueueFull">
         ///     Whether or not to block if the send queue is full.
-        ///     If false, and exception will be thrown in the event.
+        ///     If false, a KafkaExcepion (with Error.Code == ErrorCode.Local_QueueFull) 
+        ///     will be thrown if an attempt is made to produce a message
+        ///     and the send queue is full.
```
edenhill: when

src/Confluent.Kafka/Producer.cs
```diff
@@ -487,9 +495,11 @@ public Task<Message> ProduceAsync(string topic, byte[] key, int keyOffset, int k
...
         /// 
-        ///     Note: blockIfQueueFull is set to true, background polling is 
+        ///     Warning: blockIfQueueFull is set to true, background polling is
```
edenhill: if blockIf..

src/Confluent.Kafka/Producer.cs
```diff
@@ -1008,7 +1021,7 @@ public string Name
         /// <remarks>
         ///     The partition the message is produced to is determined using the configured partitioner.
         ///     
-        ///     Blocks if the send queue is full. Note: if background polling is disabled and Poll is
+        ///     Blocks if the send queue is full. Warning: if background polling is disabled and Poll is
```
edenhill: if manual polling is enabled

src/Confluent.Kafka/Producer.cs
```diff
@@ -1032,9 +1045,11 @@ public string Name
         /// </param>
         /// <param name="blockIfQueueFull">
         ///     Whether or not to block if the send queue is full.
-        ///     If false, and exception will be thrown in the event.
+        ///     If false, a KafkaExcepion (with Error.Code == ErrorCode.Local_QueueFull) 
+        ///     will be thrown if an attempt is made to produce a message
+        ///     and the send queue is full.
         ///      
-        ///     Note: blockIfQueueFull is set to true, background polling is 
+        ///     Warning: blockIfQueueFull is set to true, background polling is
```
edenhill: if block...

src/Confluent.Kafka/Producer.cs
```diff
@@ -1126,16 +1144,22 @@ public void ProduceAsync(string topic, TKey key, TValue val, bool blockIfQueueFu
         ///     Raised when there is information that should be logged.
         /// </summary>
         /// <remarks>
-        ///     You can specify the log level with the 'log_level' configuration
-        ///     property. You also probably want to specify one or more debug
-        ///     contexts using the 'debug' configuration property.
+        ///     Note: By default not many log messages are generated.
```
edenhill: dito.

Also, there is really no point in setting the log_level, you could probably just remove any reference to it.
If you enable debugging the log_level is automatically increased to 7.
But none of this is very interesting to app developers, and the info is available in CONFIGURATION.md.
There are more important concepts to explain than this one.

src/Confluent.Kafka/Producer.cs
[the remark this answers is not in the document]
darshanlakhwani: can you please which is best nugget package for using Kafka in .net core and how i can read topic with multiple partition and how we distributed data among the topic in partition using .net core

PR #65, opened 2017-02-20
WatermarkOffsets .ToString()

mhowlett:


src/Confluent.Kafka/WatermarkOffsets.cs
```diff
@@ -28,5 +28,8 @@ public WatermarkOffsets(Offset low, Offset high)
 
         public Offset Low { get; }
         public Offset High { get; }
+
+        public override string ToString()
+            => $"[{Low}, {High}]";
```
edenhill: This looks more like a list with elements rather than a range.
Maybe "L-H" or "L..H" ?

mhowlett: or maybe explicitly spelling it out: `Low: <low> High: <high>`
not sure

ewencp: Seems like anything to indicate a range would be fine here -- otherwise this patch LGTM

mhowlett: [L, H] is in fact correct mathematical notation for a closed interval (which this is), though mathematicians are not the audience i guess.

What about:

`<low> -> <high>`

?

I was just thinking we're being somewhat inconsistent with `ToString`. `Metadata` gives a JSON representation of the object (which will generally be quite large). Everything else something custom. I think this is what we want.

edenhill: I'd say .. and - are the most common range separators in computer lingo.

mhowlett:
ok .. is fine by me.

PR #79, opened 2017-02-25
tweaks to docfx api doc generation

mhowlett:
@aayars - post release i guess.
 
This PR (I'm 90% sure) will ensure docfx makes API docs that are suitable to be placed in any path (links are all relative, and don't reference 'up' the tree). html files will be all under api/.. i.e. we'll have:

http://docs.confluent.io/[version]/clients/confluent-kafka-dotnet/api/Conflunet.Kafka.html

.css etc are one level up from api. it might be better to not have the 'api' component in the path, but i couldn't easily see how to make docfx do that.

### DIFF
--- a/doc/docfx.json
+++ b/doc/docfx.json
@@ -11,13 +11,13 @@
       ],
       "dest": "api"
     }
+    
   ],
   "build": {
     "content": [
       {
         "files": [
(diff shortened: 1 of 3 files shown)


aayars:
:+1:

Issue #75, opened 2017-02-24
delivery report timestamps are incorrect


mhowlett:
delivery report timestamps should reflect the time the message was produced but they are @ some time during the year 1970. The problem is the timestamp is being errantly dividing 1000 in the `Producer.DeliveryReportCallbackImpl` method. Timestamps actually produced to Kafka are correct.

mhowlett:
same problem on the consumer (`SafeKafkaHandle.ConsumerPoll`) - the unix millisecond time returned by kafka is interpreted as seconds.. and then this value interpreted as milliseconds. Remove / 1000 to fix.

mhowlett:
resolved in master

PR #80, opened 2017-02-27
fixes incorrect timestamps in delivery report + consumed messages

mhowlett:
fix for #75 

@edenhill


mhowlett:
it's not currently possible to set timestamps explicitly

mhowlett:
@edenhill good point.

now includes tests for every variant of produceasync and consumption (for default time values only - explicit timestamps are not supported yet).
 
I also changed fixed group id's to guids for improved test robustness (in the event of failure, groups may stay alive and interfere with subsequent runs).

PR #94, opened 2017-03-09
-> MSBuild

mhowlett:
@edenhill 

Corresponding to the recent release of VS 2017, msbuild is the thing to use cross platform.

I tested loading / building the solution file in VS 2017 as well as building using the CLI tools on MacOS.

I tested creating the library nuget package on windows using CLI tools and referencing this in various places (including from VS 2015).


examples/AdvancedProducer/Program.cs
```diff
@@ -20,7 +20,6 @@
 using System.IO;
 using System.Text;
 using System.Collections.Generic;
-using System.Threading.Tasks;
```
mhowlett: sneaking in a tiny bit of tidyup ...

edenhill:
Is this backwards compatible with VS 2013 (which seems pretty common)?

mhowlett:
You can certainly reference the Confluent.Kafka nuget packages from projects in VS 2013.

I don't know if they are back porting the tooling changes to work with VS 2013 (allowing the sln in this repo to be loaded), but i doubt it. I'm not even sure if they will be supported in VS 2015 (there were preview releases of the tooling for that, but i'm not sure if there is / will be anything released proper). 

I don't think we have much choice but to upgrade - it's going to get pretty awkward pretty quickly if we don't. We should ideally provide sln / project files for *examples* compatible older VS versions (with pretty high priority - it's an ease of getting going thing), but I want that to be a separate prioritized PR (it's kind of a separate chunk of work). In doing this, I think we will likely also want to switch the example projects over to reference the nuget package rather than the project in the path (to match how people would use them).

mhowlett:
Some clarifying notes ... 
1: The existing .sln isn't compatible with VS 2013 either.
2: The move to .csproj is required in order to be able to load the solution in VS 2017.
3: project.json was only ever an interim thing - we shouldn't pay any regard to this going forward.

edenhill:
How is this affecting our development on non-VS platforms?

mhowlett:
Unless i'm missing something, the latest dotnet cli tools are also not compatible with project.json (need to be migrated to csproj).

mhowlett:
i.e. csproj is necessary to allow dev on non-windows platforms as well.

edenhill:
.travis.yml needs to be fixed:
https://travis-ci.org/confluentinc/confluent-kafka-dotnet/jobs/209549064#L306

mhowlett:
i was ignoring that 'cause it's never worked. ok, i'll look into it.

edenhill:
Sure it did, for Linux. The OSX port failed due to mismatching runtimes.

We really need CI to work (for us) so please fix it if it doesn't work.

mhowlett:
yep i'm looking into it.

.travis.yml
```diff
@@ -8,18 +8,13 @@ matrix:
   include:
   - os: linux
     dist: trusty # Ubuntu 14.04
-    dotnet: 1.0.0-preview2-1-003177
+    dotnet: 1.0.1
     mono: none
     env: DOTNETCORE=1
     sudo: required
-  - os: osx
```
edenhill: Since many developers seem to be using OSX, do we really want to remove this?

.travis.yml
```diff
@@ -8,18 +8,13 @@ matrix:
...
-    osx_image: xcode7.3 # macOS 10.11
-    dotnet: 1.0.0-preview2-1-003177
-    mono: none
-    env: DOTNETCORE=1
 
 install:
-   - sed -i.bak 's/^.*\"net451\":.*$//g' src/Confluent.Kafka/project.json
+   - sed -i.bak 's/net451;//g' src/Confluent.Kafka/Confluent.Kafka.csproj
```
edenhill: nit: Perhaps provide more context to this replacement to avoid future inadvertent matches?

mhowlett: yep. another git amend coming for this. hopefully the last one makes the build work.

.travis.yml
mhowlett: I removed it 'cause I thought you meant this is what was breaking it. ok, i'll add it back and try and make it work, if I fail, i'll remove it again because i think it's better to get the green light and have this as an open todo.

i've got it building now anyway.

.travis.yml
mhowlett: (btw, good point here)

mhowlett:
w00t! building of all examples and tests + running unit tests on travis is finally working on both linux + macos.

the sticking point on linux was getting the library to build. it was behaving as if `dotnet restore` was not working, or maybe it was trying to build for net451 (even though this is not specified in the csproj file). I replicated the environment / scenario in docker locally, and the problem was not happening. I don't know what's wrong. However it doesn't matter... building and executing projects that target netcoreapp1.1 that reference the library work fine, and implicitly this means the library itself must be built and operational for netstandard1.3 (to the extent of unit test coverage). So skipping the explicit library build doesn't omit any verification that explicitly building it would. So I just don't build the library explicitly. This also apparently avoids the too many open files problem on macos.

Makefile
```diff
@@ -4,30 +4,33 @@
 # Example:
 #   make build
 
-
-DIRS=$(shell find . -name project.json -exec dirname {} \;)
 OS=$(shell uname -s)
-LINUX_FRAMEWORK=netcoreapp1.0
-DEFAULT_FRAMEWORK=$(LINUX_FRAMEWORK)
 
+EXAMPLE_DIRS=$(shell find ./example -name *.csproj -exec dirname {} \;)
```
edenhill: escape wildcard: `-name '*.csproj'`

Makefile
```diff
@@ -4,30 +4,33 @@
...
+TEST_DIRS==$(shell find ./test -name *.csproj -exec dirname {} \;)
+
+LINUX_FRAMEWORK=netcoreapp1.1
+DEFAULT_FRAMEWORK=$(LINUX_FRAMEWORK)
```
edenhill: nit: Might be usable to have `DEFAULT_FRAMEWORK?=..` to allow `make DEFAULT_FRAMEWORK=hehe examples`

appveyor.yml
```diff
@@ -4,20 +4,20 @@ configuration:
 - Release
 
 install:
-- ps: Start-FileDownload 'https://download.microsoft.com/download/A/3/8/A38489F3-9777-41DD-83F8-2CBDFAB2520C/packages/DotNetCore.1.0.0-SDK.Preview2-x64.exe'
-- cmd: DotNetCore.1.0.0-SDK.Preview2-x64.exe /quiet
+- ps: Start-FileDownload 'https://download.microsoft.com/download/8/F/9/8F9659B9-E628-4D1A-B6BF-C3004C8C954B/dotnet-1.1.1-sdk-win-x64.exe'
+- cmd: dotnet-1.1.1-sdk-win-x64.exe /quiet
```
edenhill: Why is AppVeyor on dotnet 1.1.1 but travis on 1.0.1?

mhowlett: they refer to the same thing. downloading this then doing `dotnet version` will give you 1.0.1. I don't know why this is - MSs versioning confuses me.

Makefile
mhowlett: ok, might as well, though these are a bit superfluous anyway.

Makefile
mhowlett: This comment actually alerted me to a quite important mistake - should be finding in `./examples` not `./example`. Also, fixing the quotes.

PR #98, opened 2017-03-14
Run unit tests under standard net framework

treziac:
The unit tests were only run under .net core, we should run them under netframework too.
We can only use net452+ due to xunit reference, better than nothing

VS2017 is buggy when multiple frameworks are used for now and tests won't be discovered anymore in the test explorer (because of multiple frameworks). They can still be run from CLI though so not an issue

I don't know how integration tests / benchmarking are run on confluent side so only included the unit tests

Also removed the sed in travis CI, it's not necessary now that we have the frawmeork specified when building/testing (and it would prevent the restore with the changes in Kafka.UnitTests)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,6 @@ matrix:
     env: DOTNETCORE=1
 
 install:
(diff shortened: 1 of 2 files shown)


edenhill:
LGTM

Thanks for this @treziac

mhowlett:
LGTM

Issue #96, opened 2017-03-10
support target framework 4.5


fangliuwh:
We are still at .net framework 4.5. Can you please add support of framework 4.5 besides 4.5.1 and .NETStandard 1.3? Thanks.

mhowlett:
.NET Framework 4.5 (and 4.5.1) are no longer supported by Microsoft so I'm not going to prioritize looking into this unless lots of other people ask for it. We'll look at PRs or any investigation people might do into the viability of this though :-). The decision is inherited from rdkafka-dotnet and I don't know what @ah-'s reasons were. By default I assume there were reasons (but it may be somewhat arbitrary).

ah-:
I don't quite remember, I think it was due to some bug and/or .net standard incompatibility in 4.5.0.

combrinckd:
The project I am working on are also still on .NET 4.5 and updating this is not an option for at least a couple of months (maybe next year). Currently I am using kafka-net (https://www.nuget.org/packages/kafka-net/ ) producer. It would have been great to also validate this to find the best performance and stability on throughput. I need to be able to push at least 3000 to 4000 messages per second.

treziac:
Had also this problem when working first with rdkafka but could update my packages, made a quick PR as it doesn't to introduce any issue
I assume it's not an uncommon case - in our company, introduce kafka in a legacy project before to move to new projects in .net core or other languages

@combrinckd : i would advise you to check that your production server run under net framework 4.5.2 (or 4.6+) and not 4.5, as it's not supported anymore by microsoft. You don't have to change your project at all, just the framework installed on the servers. There are very rare breaking change doing this upgrade that you can find here : https://msdn.microsoft.com/en-us/library/dn458358(v=vs.110).aspx
If possible update your projects, it's not that long ;)

mhowlett:
ok - there seems to be sufficient demand for this, and @treziac worked out all the details and submitted a PR.. so it's in!

treziac:
As a note, we may want to add a net45 and net451 reference and not just net45 and maybe even other moniker
(everything which is noted as obsolete in an upper .NET Framework version should target a new version. I don't think there will be any breaking changes in future but yet. There should be no obsolete warnings when targeting any framework)

Will do this later, no urge as long as it's only obsolete and not removed :)

PR #99, opened 2017-03-14
Change net451 to net45

treziac:
Solve #96 
Just had to re implement the Marshal generic functions from https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Runtime/InteropServices/Marshal.cs

### DIFF
--- a/src/Confluent.Kafka/Internal/Util.cs
+++ b/src/Confluent.Kafka/Internal/Util.cs
@@ -16,6 +16,7 @@
 
 using System;
 using System.Text;
+using SystemMarshal = System.Runtime.InteropServices.Marshal;
 
 
 namespace Confluent.Kafka.Internal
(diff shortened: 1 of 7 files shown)


src/Confluent.Kafka/Internal/Util.cs
```diff
@@ -33,14 +34,41 @@ public static string PtrToStringUTF8(IntPtr strPtr)
...
                 var strBuffer = new byte[length];
                 System.Runtime.InteropServices.Marshal.Copy(strPtr, strBuffer, 0, length);
                 return Encoding.UTF8.GetString(strBuffer);
             }
+
+            public static T PtrToStructure<T>(IntPtr ptr)
+            {
+#if NET45
+                return (T)SystemMarshal.PtrToStructure(ptr, typeof(T));
+#else
```
mhowlett: The only potential concern I have with this PR is that this would be less efficient than the generic version and it's used every time `ConsumerPoll` handles a new message, which is frequently. I think it should be a very minimal hit though - `typeof(T)` doesn't allocate a new `Type` class on every invocation, so the GC isn't being pressured. Also, `Message` is a class, so there is no unboxing going on.

mhowlett:
Ok, LGTM - it seems 4.5 is popular enough to prioritize over an probably insignificant perf hit.

thanks for sorting this out @treziac

PR #102, opened 2017-03-15
StringSerializer and StringDeserializer handle null input by returning null

bjornicus:


bjornicus:
Great, I've deleted the TODO and remarks, and added a similar doc comment to Deserialize to match the one that existed on Serialize.   I don't see any .nuspec or anything to update, so I assume the NuGet package will get updated automatically by the CI build, or you all have some other release process for package updates?

src/Confluent.Kafka/Serialization/StringDeserializer.cs
```diff
@@ -30,6 +27,12 @@ public StringDeserializer(Encoding encoding)
             this.encoding = encoding;
         }
 
+        /// <param name="data">
+        ///     The data to deserialize
+        /// </param>
+        /// <returns>
+        ///     <paramref name="data" />deserialized to a string (or null if data was null).
```
mhowlett: space after `/>`

src/Confluent.Kafka/Serialization/StringSerializer.cs
```diff
@@ -36,10 +36,6 @@ public StringSerializer(Encoding encoding)
         /// <returns>
```
mhowlett: remove TODO at top of this fiel

src/Confluent.Kafka/Serialization/StringSerializer.cs
```diff
@@ -36,10 +36,6 @@ public StringSerializer(Encoding encoding)
...
         ///     <paramref name="val" /> encoded in a byte array.
```
mhowlett: perhaps add ' (or null if `<param name="val" />` is null)'

bjornicus: param or paramref?

bjornicus: ok, reading the docs and looks like paramref is the correct choice.

mhowlett: ahh yes, `paramref`

mhowlett:
Sorry - only just thought of this - I think we should include this functionality in an integration test. It would slot easily into `test/Confluent.Kafka.IntegrationTests/Tests/SimpleProduceConsume.cs` - produce another message with a null string value after the one already produced then consume it and check it's null. 

Feel free, if not, I'll get around to it :-).

bjornicus:
ok, I added something to the integration test, but am having trouble running the integration tests, so I'm going to lean on the CI system to run it :(

(For some reason my Visual Studio isn't finding any of the tests, despite the xunit test runner nuget being installed, I was able to run the unit tests with the command line runner, but the integration tests don't seem to produce a net452 build, and the netcoreapp1.1 build doesn't seem to include what the command line runner needs 😞 )

mhowlett:
Can you make line 76 this, then i'll merge:

`Assert.Equal(testString, msg.Value == null ? null : Encoding.UTF8.GetString(msg.Value, 0, msg.Value.Length));`

Thanks for the perseverance :-)... 

Integration tests aren't currently run automatically - and thanks for the feedback on your experience running them (noted - need to sort that out better).

mhowlett:
evidently I have permission to push to your master branch - made the above change and will merge after checks finish.

thanks...

PR #105, opened 2017-03-19
Add tests for IntSerializer and IntDeserializer

wangyanjun:
Add tests for IntSerializer and IntDeserializer


mhowlett:
thanks for the contribution @wangyanjun - there actually is already a good set of int serialisation / deserialization tests, but they're in IntSerde.cs not Serialization/Int.cs. Feel free to do a PR to make that file name change instead.

Or, if you're extra keen to help, how about a long serializer / deserializer + unit tests? that'd be awesome :-)

wangyanjun:
OK. I'd like to do this.

PR #106, opened 2017-03-20
Add long serializer / deserializer and related unit tests

wangyanjun:
Add long serializer / deserializer and related unit tests, refs this project and java client's implements


src/Confluent.Kafka/Serialization/LongDeserializer.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// Deserialize data to Int64
+        /// </summary>
+        /// <param name="data">input data</param>
+        /// <returns>deserialize result</returns>
+        public long Deserialize(byte[] data)
+        {
+            if (data == null)
+            {
+                // in accordance with java client.
+                return default(long);
```
mhowlett: The java client serialization uses `Long`, which is the `Object` form of `long`, which can be null.

By contrast, the numeric serialization in confluent-kafka-dotnet doesn't allow for null values. In some scenarios this would be useful, but I think it deserves a different set of serializers because it's a weaker type than required in many situations.

returning `default(long)` is confusing here IMO. Better would be to throw an `ArgumentException` stating that null values are not allowed.

test/Confluent.Kafka.UnitTests/Serialization/Long.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            var data = new LongSerializer().Serialize(23L);
+            Assert.Equal(23, data[7]);
+            Assert.Equal(0, data[0]);
+        }
+
+        [Fact]
+        public void DeserializeArgNull()
+        {
+            Assert.Equal(0L, new LongDeserializer().Deserialize(null));
```
mhowlett: change test to check for ArgumentException

src/Confluent.Kafka/Serialization/LongDeserializer.cs
edenhill: Allowing null values to be sent is required for compacted topics.

mhowlett: ok, i wasn't aware of this - but required only in that if you want to be able to tombstone messages you need to send be able to send `null`. This a pretty important use case of course, though the case where you never want to allow `null`s is still very prevalent.

But what to do though?
1. change value type serializer/deserializers to take `Nullable<ValueType>`
2. change value type serializer/deserializers to take `Object`
3. provide both variants or only one?

opinions from anyone on the above welcome. 

I currently think we should provide both variants, use `Nullable<ValueType>`'s and call things `NullableIntSerializer` etc.

edenhill: Since tombstones for compacted topics is an integral part of Kafka I think all serializers should allow null values and all deserializers to support passing null, without having to have two different variants of everything.

Having just Object will loose type-safe, right?

So I guess that leaves us with Nullable

mhowlett: yeah, `object` looses type safety, I think that option is out. `Nullable` is unfortunate for reasons mentioned elsewhere.

I still (currently) think it is preferable to provide both because I think it's beneficial to have enforcement of no `null` values in cases where this reflects the data / usage (which I think will be the case in many scenarios). Generally, stronger typing brings more clarity to what's going on - I quickly understand more about what's going on simply by noting which serializer is being used.

edenhill: How is the Java client handling this case?

ewencp: Maintaining 2 versions is probably confusing to users. Not absolutely to have something like NullableLongSerializer, but every time you give the user more choices that's more things you need to explain and document, and more things that can trip them up.

@edenhill I'm not sure I necessarily agree with your stance that all serdes should support null. They are critical, but in practice the place that most people use these primitive type serializers (at least in my experience) is for keys, where you don't have the tombstone issue. And in cases where they are using them for the values, perhaps we should be encouraging them to use a real serialization library that can support complex types and allow you to evolve the format over time...

treziac: Could we write a CompactProducer and CompactConsumer (or a config to activate in the producer at creation)
Even with null serialization, it does not mean the same thing to produce a null value (say for a string) and produce a null value on compact topic.
We could just add a method ProduceTombstone and in Consumer, add a OnTrombstone (which would only be called if config is activated)

edenhill: Those are good ideas @treziac, but we also want to be careful not to deviate from the Java API too much.

treziac: It's more due to language (generics can't be primitive in java so always nullable, that was easier for them and they they didn't have the choice ) and C# (where people except to find primitive serializer)

Also (but that's my point of view) I don't like the fact that it's the same method for tombstone and simple null serialization (hence the thread here : https://confluentcommunity.slack.com/archives/C499EFQS0/p1489662514821342)
People at kafka obviously don't like it either (https://cwiki.apache.org/confluence/display/KAFKA/KIP-82+-+Add+Record+Headers and others KIP)

If we want to stick with the java producer, we should force the typed consumer/producer with reference type` where TKey : class where TValue : class` so we have always null available, but we should benchmark it carefully for primitive types (int&long must be some of the most used serializer with String and Avro)

wangyanjun: Basically agree with @treziac, but i tend to not import fundamental API differences, because many new comings will reference huge amount java's demos. 
Can we just make it configurable for serialization? if not configured, work under strict mode, because pass null value maybe very rare. 
Each serializer treat null value differently according to the configuration, output a log while occurred.

test/Confluent.Kafka.UnitTests/Serialization/Long.cs
wangyanjun: OK. 
Should we create a new issue to trace problem of serialization?

src/Confluent.Kafka/Serialization/LongDeserializer.cs
mhowlett: @ewencp - I actually think it might be less confusing to have the two types. But to your second point, I have to admit I was completely asleep yesterday and imagining tomb-stoning was by done by setting the key to `null` (which of course is comically wrong). Now i'm thinking straight, I have a definite bias towards not supporting `null` values in numeric serializers as they won't be as commonly used as I was errantly thinking.

@treziac and repliers - in general agreement here - the API is not necessarily the best (though i imagine the people coming up with it were probably making tradeoffs). And we shouldn't deviate.

@wangyanjun 's PR LGTM except for some comment formatting / english things which I'm happy to change myself. I'll try to do that inline on github or push to your branch, unsure what's going to work.

test/Confluent.Kafka.UnitTests/Serialization/Long.cs
```diff
@@ -0,0 +1,72 @@
...
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace Confluent.Kafka.UnitTests
+{
+    public class LongTests
+    {
+        [Theory]
+        [MemberData(nameof(TestData))]
```
mhowlett: nice

PR #110, opened 2017-03-22
Added serializer and deserializer for long data type

MrGlenThomas:


mhowlett:
see #106. When I first saw this I thought you might be creating controversy by allowing for nullable values, but instead you're reinforcing my existing view point by not. thanks :-)

MrGlenThomas:
I hadn't seen that other pull request.

mhowlett:
thanks for the effort anyway..

PR #113, opened 2017-03-22
[WIP] Producev and datetime

treziac:
Fix producev binding, producing timestamp now works correctly

I only added one API (for easy testing)
Adding a parameter on all the produce api generates really many method.
Should we do like the java producer and have a ProducerRecord<TKey, TValue>?
Then we would have only the producerRecord and the blockIfQueueFull

Producev and produce doesn't return the same thing, had to duplicate the code. I think the goal is to only use producev in future (from what i saw in commentary but not sure. @edenhill ?)


edenhill:
That's right @treziac, producev() is the future.

mhowlett:
This is awesome - would be so happy to get this one in. 

We discussed this previously and arrived at using overloads. I think it's manageable for this many parameters (but not any more) and it's the most straightforward to use from a user perspective (no need to create another object).

Feel free to argue against if you strongly object

src/Confluent.Kafka/Impl/SafeTopicHandle.cs
```diff
@@ -88,28 +88,83 @@ internal long Produce(byte[] val, int valOffset, int valLength, byte[] key, int
...
+            }
+
+            try
+            {
+                // TODO: when refactor complete, reassess the below note.
+                // Note: since the message queue threshold limit also includes delivery reports, it is important that another
+                // thread of the application calls poll() for a blocking produce() to ever unblock.
+
+                return LibRdKafka.producev(
+                    kafkaHandle.DangerousGetHandle(),
```
treziac: I don't like it but that was the simplest to implement first
@edenhill  should we put it directly in the SafeKafkaHandle and call it with string rk/topic rather than rk/rkt?

test/Confluent.Kafka.IntegrationTests/Tests/Producer_ProduceAsync_DeliveryHandler.cs
```diff
@@ -50,9 +51,8 @@ public void HandleDeliveryReport(Message dr)
                 Assert.Equal(Topic, dr.Topic);
                 Assert.True(dr.Offset >= 0);
                 Assert.Equal(TimestampType.CreateTime, dr.Timestamp.Type);
-                Assert.True(Math.Abs((DateTime.UtcNow - dr.Timestamp.DateTime).TotalMinutes) < 1.0);
 
-                if (Count < 5)
+                if (Count != 5)
                 {
                     Assert.Equal(TestKey, dr.Key);
```
treziac: I don't seem to receive the handler in order. They are fired i different thread (note that i didn't run the bootstrap.sh and had a 3 partition topic) and even locking didn't solve the problem
Doing this test on a 1 partition topic doesn't seem to be right as we test partitioning

test/Confluent.Kafka.IntegrationTests/Tests/Producer_ProduceAsync_DeliveryHandler.cs
```diff
@@ -63,6 +63,15 @@ public void HandleDeliveryReport(Message dr)
                     Assert.Equal(new byte[] { 7 }, dr.Value);
                 }
 
+                if (Count == 6)
+                {
+                    Assert.True(Math.Abs((Now - dr.Timestamp.DateTime).TotalMilliseconds) < 1.0);
```
treziac: I didn't check this 1.0 : is it [-0.5:0.5] or [-1.0;1.0] or [0:1.0]? Not sure we care but should be in doc (especially the fact it's not the same datetime if datetime is more precise than ms)

Also need a test with local datetime

src/Confluent.Kafka/Impl/SafeTopicHandle.cs
edenhill: If we can scrap the local rkt cache and only producev() that'd be great!

treziac: should we completely remove the rd_kafka_produce binding?
(the doc on librdkafka barely mention producev so don't know if it has been proved enough)

edenhill: I think we should.

treziac: Will go for it then :)
On producev, does providing -1 as partition is the same as not giving any partition?
Is there any impact in doing it?

If we want to stick the the librdkafka implementation, we will have to make a DllImport for every possible parameter list we may want to invoke (with/out partition, with/out timestamp, with/out opaque...)
It's ok to do this, just want to be sure it's necessary.
Using __arg_list could really help the implementation but i don't like to use an undocumented feature...

treziac:
I completly removed produce for showing what it may look like.
Might be better to keep both, and perhaps just a boolean to choose which implementation we want to use at producer creation (and also for #115 : do we want to keep it working with librkdaka which doesn't provide producev?)

Note that i couldn't remove SafeKafkaTopicHandle because of getmetadata. It misses a method to call it with in char * rather than an rkt for get metadata on a topic only @edenhill

treziac:
Regarding overloads, what happens if in the future if we add a bool or int ? We won't be able to make an overload which works with only this parameter, or have to mess with parameter order. It's still ok with two optional parameters (partition and blockifqueuefull) but grow exponentially (adding datetime add 4 methods, adding an other parameter will add 20 more methods...)

I would prefer to go to two methods:
```
ProduceAsync(string topic, TKey key, TValue value) //keep the most used one
ProduceAsync(ProduceRecord<TVKey, TValue> produceRecord)
```

And add a structure
```
public struct ProduceRecord<TKey, TValue>
    {
        public ProduceRecord(string topic, TKey key, TValue value)
        {
            Topic = topic;
            Key = key;
            Value = value;
            Partition = Producer.RD_KAFKA_PARTITION_UA;
            Datetime = Timestamp.NO_PRODUCE_DATETIME;
            BlockIfQueueFull = true;
        }

        public string Topic { get; }
        public TKey Key { get; }
        public TValue Value { get; }
        public int Partition { get; set; }
        public DateTime Datetime { get; set; }
        public bool BlockIfQueueFull { get; set; }
    }
```

producing then looks like:


```
var deliveryReport = producer.ProduceAsync(topicName, key, val);
var deliveryReport2 = producer.ProduceAsync(new ProduceRecord<string, string>(topicName, key, val) {Partition = 1, BlockIfQueueFull = true });
```

This make:
- less methods, so easier to maintain
- closer to java API
- more complicated for user only if he doesn't use the simple method (so not a problem for a user discovering the lib)
- no problem if we add parameter. Just add a field in the struct. The struct might me big, but we pass all the parameter anyway so I assume it's ok
- break previous API (optionally, we could just mark them obsolete for the next release)

mhowlett:
Ok. Headers are coming along very soon as well which will make all combinations of parameters unworkable. So I agree with you, I think we should have a `ProducerRecord` based method and a small number (possibly only one, the one you suggest) of commonly used overloads. We should mark any uncommonly used ones as depreciated.

@edenhill @ewencp - if I recall correctly, you already agree with the above but feel free to comment if not.

treziac:
Close in favor of #121

PR #117, opened 2017-03-23
[WIP - Experimental] Change ProduceAsync methods: use Produce record and reduce API

treziac:
Look to change the api following this view:

- Producer is the entrypoint to librdkafka, and should be used in two cases: low level producing (memory management with ArraySegement, etc.) and be a single producer instance to call GetSerializingProducer when the app use multiple topic with multiple types. If we use this, we have to know what we do, so no need to overload.
- Producer<TKey, TValue> should be used when only one typed producer is needed
- Reduce at maximum the API regarding produce.

Note this will be breaking API. In future, if librdkafka add parameter to producev, to have backward assembly compatibility we will have to add overload to Producer.Produce and add the field on ProducerRecord.

Serialized producer:
Change the api regarding ProduceAsync: only two overload, the simple one (topic, key, value) and one with a ProducerRecord<TKey, TValue> struct

Change the datetime to timestamp. DateTime still usable in ProducerRecord

I also wonder about renaming the ProduceAsync(..., IDeliveryHandler) to Produce. Async postfix whould be for awaitable function - here we await a parameter so not sure about this...

Also, it might worth testing adding all overloads on Producer and calling the corresponding rd_kafka_producev (without the extra parameter, only the ones needed), same in SafeKafkaHandle (all possible overload and call to corresponding binding with variable args)

If not accepted, some thing are still worth taking. Didn't do atomic commits, my bad. Directly branched from #113


mhowlett:
lots of great stuff here @treziac. 

If you can you break this PR up a bit and submit a smaller number of more focussed PRs that would mean I get to review quicker (i know this is more work at your end...). I will still be happy to review this big PR, but it'll take me longer to get to + get the changes merged.

treziac:
Closing in favor of #118 and #113. Will post an other PR for the long/DateTime type once 118 is merged or refused.

PR #118, opened 2017-03-24
Rework and reduce ProduceAsync API with ProduceRecord struct

treziac:
#117 with only commits regarding the API rework. no change on timestamp or on producev

All ProduceAsync function (ISerializingProducer, Producer, Producer<K,V> and SerializingProducer<K,V> are replaced with two ProduceAsync func:
- one which only take topic, key, val
- one which take a ProduceRecord where we can add parameters later
- (*2 because deliveryHandler change the return type)
- I put all the old APIs as obsolete and error at compilation. Clients will need to upgrade code when upgrading, so replacing Confluent.Kafka.dll won't generate a runtime error
(we could just put an obsolete warning in a first time though, but if we go for this, the sooner we get rid of old api the better. It's still showed in intellisense so makes eveything confuse) Would hapily remove them completly now :)

Regarding ProducerRecord

- I could have left an empty constructor only with all properties settable, but I think we should force the topic / key / value to be set.
- BlockIfQueueFull is not really part of the message, but it seemed better to include it in
- DateTime is here for maintaining compatibility, I want to change it to a DateTime / UnixTimestamp later (see https://github.com/confluentinc/confluent-kafka-dotnet/pull/117/files#diff-df6efad58be109a8fb4f7205d30f38f3R71)

Didn't do benchmark nor integration test (no kafka on this machine)
Will be on holiday for one week so I likely won't add other PR from #117 till then

### DIFF
--- a/src/Confluent.Kafka/Producer.cs
+++ b/src/Confluent.Kafka/Producer.cs
@@ -194,6 +194,150 @@ namespace Confluent.Kafka
                 SetResult(message);
             }
         }
+        
+        /// <summary>
+        ///     Detailed messaged
+        /// </summary>
+        public struct ProduceRecord
+        {
+            /// <summary>
+            ///     Initialize a message which can be personalized via properties.
+            /// </summary>
+            /// <param name="topic">
+            ///     The target topic.
+            /// </param>
+            /// <param name="key">
+            ///     null, or a byte array that contains the message key.
+            /// </param>
+            /// <param name="keyOffset">
+            ///     for non-null values, the offset into the key array of the
+            ///     sub-array to use as the message key.
+            ///     if <paramref name="key" />  is null, keyOffset must be 0.
+            /// </param>
+            /// <param name="keyLength">
+            ///     for non-null keys, the length of the sequence of bytes that
+            ///     constitutes the key.
+            ///     if <paramref name="key" />  is null, keyLength must be 0.
+            /// </param>
+            /// <param name="val">
+            ///     null, or a byte array that contains the message value.
+            /// </param>
+            /// <param name="valOffset">
+            ///     for non-null values, the offset into the val array of the
+            ///     sub-array to use as the message value.
+            ///     if <paramref name="val" /> is null, valOffset must be 0.
+            /// </param>
+            /// <param name="valLength">
+            ///     for non-null values, the length of the sequence of bytes that
+            ///     constitutes the value.
+            ///     if <paramref name="val" /> is null, valLength must be 0.
+            /// </param>
+            public ProduceRecord(string topic,
+                byte[] val, int valOffset, int valLength,
+                byte[] key, int keyOffset, int keyLength)
+            {
+                Topic = topic;
+                Key = key;
+                KeyLength = keyLength;
+                KeyOffset = keyOffset;
+                Value = val;
+                ValueLength = valLength;
+                ValueOffset = valOffset;
+                Partition = RD_KAFKA_PARTITION_UA;
+                BlockIfQueueFull = true;
+            }
+
+            /// <summary>
+            ///     Initialize a message which can be personalized via properties.
+            /// </summary>
+            /// <param name="topic">
+            ///     The target topic.
+            /// </param>
+            /// <param name="key">
+            ///     null, or a byte array that contains the message key.
+            /// </param>
+            /// <param name="value">
+            ///     null, or a byte array that contains the message value.
+            /// </param>
+            public ProduceRecord(string topic, byte[] value, byte[] key)
+            {
+                Topic = topic;
+                Key = key;
+                KeyLength = key?.Length ?? 0;
+                KeyOffset = 0;
+                Value = value;
+                ValueLength = value?.Length ?? 0;
+                ValueOffset = 0;
+                Partition = RD_KAFKA_PARTITION_UA;
+                BlockIfQueueFull = true;
+            }
+
+            /// <summary>
+            ///     The target topic.
+            /// </summary>
+            public string Topic { get; }
+
+            /// <summary>
+            ///     null, or a byte array that contains the message key.
+            /// </summary>
+            public byte[] Key { get; }
+
+            /// <summary>
+            ///     for non-null keys, the length of the sequence of bytes that
+            ///     constitutes the key.
+            ///     if <see cref="Key" />  is null, KeyOffset must be 0.
+            /// </summary>
+            public int KeyOffset { get; }
+
+            /// <summary>
+            ///     for non-null keys, the length of the sequence of bytes that
+            ///     constitutes the key.
+            ///     if <see cref="Key" />  is null, KeyLength must be 0.
+            /// </summary>
+            public int KeyLength { get; }
+
+            /// <summary>
+            ///     null, or a byte array that contains the message value.
+            /// </summary>
+            public byte[] Value { get; }
+
+            /// <summary>
+            ///     for non-null values, the length of the sequence of bytes that
+            ///     constitutes the value.
+            ///     if <see cref="Value" />  is null, ValueOffset must be 0.
+            /// </summary>
+            public int ValueOffset { get; }
+
+            /// <summary>
+            ///     for non-null values, the length of the sequence of bytes that
+            ///     constitutes the value.
+            ///     if <see cref="Value" />  is null, ValueLength must be 0.
+            /// </summary>
+            public int ValueLength { get; }
+
+            /// <summary>
+            ///     The target partition (if -1, this is determined by the partitioner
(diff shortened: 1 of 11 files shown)


src/Confluent.Kafka/Producer.cs
```diff
@@ -194,6 +194,150 @@ public void HandleDeliveryReport(Message message)
                 SetResult(message);
             }
         }
+        
+        /// <summary>
+        ///     Detailed messaged
+        /// </summary>
+        public struct ProduceRecord
```
mhowlett: nested classes are typically frowned upon I believe (I think this should be moved out and into a new file).

treziac: I didn't want the user to have both ProduceRecord<K,V> and ProduceRecord to popup in intelissense, but no problem for me to move it

mhowlett: interesting point. ok, I'll think about this a bit more before having an opinion.

mhowlett:
great - just what I was hoping for :-) this is a good chunk of work to consider for a PR.

what do you think about only encapsulating the 'what' - i.e. key, value, timestamp (and later header), and leave the other things - the 'where' (topic, partition) and the 'how' (blockIfQueueFull) as parameters. I think i was previously thinking that's how i'd end up wanting to do things - if anything is going to change in the future, it's almost certainly going to be adding things to the 'what', so I think this should be future proof. the structure is more appealing to me, though it's possible I could change my mind towards your thinking (and the thinking of the authors of the java client).

I was also trying to work out before how to consolidate data structures across producer / consumer, but I couldn't see how that would work nicely (`Message` is not appropriate here). Anyway, that decision has already been made...

treziac:
I was unsure for BlockIfQueueFull - at first I did put it out of ProduceRecord and had 8 API instead of 4.
It's a choice between sanitize the API and make it fit wit kafka protocol. Keeping it like this seemed more practical for usage, but seem ok for me.

For topic/partition, I would keep it in ProduceRecord - partition and topic are part of what for me, as it's what is send to brokers - not like blockIfQueueFull or deliveryhandler, which are only how we do the things on our side (I would also consider compression and acks as a what for example if we could customize it on a Message, as it impact what is effectively stored on broker side - actually, I see a strong relation between a record and a kafka protocol ProduceMessage).

Also didn't notice at first, it's ProducerRecord and not ProduceRecord on Java Producer, we may want to have the same name

mhowlett:
I've switched over to thinking you (and others) are looking at things the right way (topic and partition are part of the what). Furthering this, I just spoke with @hachikuji and he mentioned that there is a potential use for headers related to routing. Here, header information would be somewhat like topic and partition.

treziac:
What is your opinion about depreciate / obsolete / remove the other ProduceAsync methods?
(I actually don't think there are many people who just upgrade their Confluent.Kafka dll without recompiling their app, so we should either remove it or juste depreciate)

mhowlett:
Methods we want to remove should be marked as obsolete (I believe the way you are doing it is what we want) and we will remove them entirely in a later release. We don't need to do this on the non generic `Producer` class though (we can just remove methods) because the entire class is currently marked as unstable.

The `ProducerRecord` name is inconsistent with `Message`. I don't have a good opinion on what to call this right now. Maybe `ProducerRecord` is ok. Some other options: `ProducerArgs`, `ProducerMessage`. `ProduceAsync` provides a `Message` on completion, so differentiating this class from `Message` with the keyword `Producer` seems a bit funny. `ProducerArgs` doesn't seem bad to me right now.

Other options instead of having this extra class: 
1. going back to having default args (i don't think this is a good option). 
2. could we stick with overloads but maybe there isn't a great need to have all the combinations? Have just a few common ones and one that includes everything. If someone wants to specify a non-common option, they have to specify everything. This might actually always be less verbose than constructing a `ProducerRecord` inline, which takes quite a few extra characters particularly when types are specified.

the second option is definitely more likely an acceptable option on the non-generic `Producer`, which we basically only want to expose so that people have an interface as close as possible to the raw API + as a launch pad for hanging serializing producers off. The thing we care about here is providing an efficient, fully featured raw API. doesn't have to be convenient - people are expected to provide an abstraction on top of it.

Anyway, assuming we have a `ProducerRecord` how to construct it is an interesting question, there are a number of options. The way you have it seems a decent compromise and my favorite right now. If we make it completely immutable, we just move the overload problem from `Producer` to this constructor. Also, it seems good to enforce that topic, key and value are always set (so constructor parameters + immutable as you have them). C# allows us to specify a `ProducerRecord` inline the way you have it, and separating how the required args and optional args are specified isn't bad.

src/Confluent.Kafka/Producer.cs
```diff
@@ -432,145 +576,130 @@ public int Poll()
...
+        ///     for this request.
+        /// </returns>
+        /// <remarks>
+        ///     If you require strict ordering of delivery reports to be maintained, 
+        ///     you should use a variant of ProduceAsync that takes an IDeliveryHandler
+        ///     parameter, not a variant that returns a Task&lt;Message&gt; because 
+        ///     Tasks are completed on arbitrary thread pool threads and can 
+        ///     be executed out of order.
+        /// </remarks>
+        public Task<Message> ProduceAsync(ProduceRecord produceRecord)
```
mhowlett: on the non-generic producer, i'm pretty sure I prefer not having a `ProduceRecord` overload. We can probably just get rid of a lot of the methods and keep the all inclusive one.

treziac: Ok with this.

alexvaluyskiy:
I think we could simplify the proposed API.
I don't think that we should add `ProduceRecord` class for low-level `Producer`, only for `ISerializingProducer`, because it is a high-level concept
I've changed the naming from `ProduceRecord` to `ProducerRecord`
```C#
public class ProducerRecord<TKey, TValue>
{
	public ProducerRecord(string topic, TKey key, TValue value, int partition, DateTime? timestamp) {}

	public string Topic { get; }
	public TKey Key { get; }
	public TValue Value { get; }
	public int Partition { get; }
	public DateTime? Timestamp { get; }
}
```
All properties of the message are immutable (the same as in Java)
And I've removed `blockIfQueueFull` from the `ProducerRecord` because semantically it's not the part of a message
In the future we could add the headers here.

The API will look like this
```C#
public interface ISerializingProducer<TKey, TValue>
{
	...
	Task<Message<TKey, TValue>> ProduceAsync(ProducerRecord<TKey, TValue> record);
	Task<Message<TKey, TValue>> ProduceAsync(ProducerRecord<TKey, TValue> record, bool blockIfQueueFull);
	Task<Message<TKey, TValue>> ProduceAsync(string topic, TKey key, TValue val);

	void ProduceAsync(ProducerRecord<TKey, TValue> record, IDeliveryHandler<TKey, TValue> deliveryHandler);
	void ProduceAsync(ProducerRecord<TKey, TValue> record, bool blockIfQueueFull, IDeliveryHandler<TKey, TValue> deliveryHandler);
	void ProduceAsync(string topic, TKey key, TValue val, IDeliveryHandler<TKey, TValue> deliveryHandler);
}
```

Here is my implementation: https://github.com/alexvaluyskiy/confluent-kafka-dotnet/commits/simplerproducer

mhowlett:
thanks for the input @alexvaluyskiy - i'll hopefully get back to this PR sooner rather than later.

Tungsten78:
Providing the event timestamp via `Producer<>.ProduceAsync()` is currently not exposed via any public contract methods.  This happens to be a side-effect of this PR.

Any idea when we might get this?

Tungsten78:
+1 on @alexvaluyskiy's submission for a simpler API that is more in-line with the Java implementation.

In my usage, I am needing to specify the event time but not the partition.  `ProducerRecord<>` would benefit from having `Partition` as a nullable - this would prevent having to embed `RD_KAFKA_PARTITION_UA`'s constant in the caller.

mhowlett:
@Tungsten78 - this is about the highest priority for this client after avro/sr. we'll have a producer record, and some other commonly used overloads.

mhowlett:
thanks for your input here @treziac and others - closing this in favor of another PR currently in progress. We'll be reusing the `Message` class. Also getting rid of `blockIfQueueFull` as a parameter (make it a config dictionary property).

Tungsten78:
@mhowlett - for reference what is the PR in which this work will continue?   Will the scope include making the `ProducerRecord` timestamp settable?

mhowlett:
@Tungsten78 - there is no PR yet (but i'm nearly finished it in a local branch). Yes, you'll be able to set a custom timestamp.

gcmbr:
Custom TimeStamps will be a great addition.  Thank you for the enhancement.

PR #120, opened 2017-03-24
Use unix timestamp rather than DateTime in Timestamp structure

treziac:
Timestamp is only created from librdkafka values (message notification and message consumed) so it's better to use the timestamp to define it.

DateTime is misleading given its kind (to datetime with different kind are not equal even if they have the same timestamp) and that it can't always be converted to a unix timestamp (precision is below the millisecond)
This PR also make the conversion more consistent (the static method DateTime -> timestamp now always round down to the latest millisecond - I like API to be 100% predictable)

### DIFF
--- a/src/Confluent.Kafka/Timestamp.cs
+++ b/src/Confluent.Kafka/Timestamp.cs
@@ -84,8 +84,15 @@ namespace Confluent.Kafka
         /// <summary>
         ///     Gets the Utc DateTime corresponding to the <see cref="UnixTimestampMs"/>.
         /// </summary>
+        [Obsolete("Prefer DateTimeOffset to avoid DateTimeKind issue, this may be removed in a future release")]
         public DateTime DateTime
-            => UnixTimestampMsToDateTime(UnixTimestampMs);
+            => DateTimeOffset.UtcDateTime;
+
+        /// <summary>
+        ///     Gets the DateTimeOffset corresponding to the <see cref="UnixTimestampMs"/>.
+        /// </summary>
+        public DateTimeOffset DateTimeOffset
+            => new DateTimeOffset(UnixTimestampMsToDateTime(UnixTimestampMs));
 
         public override bool Equals(object obj)
         {
@@ -128,7 +135,7 @@ namespace Confluent.Kafka
         ///     The milliseconds unix timestamp to convert.
         /// </param>
         /// <returns>
-        ///     The DateTime value associated with <paramref name="unixMillisecondsTimestamp"/>
+        ///     The DateTime value associated with <paramref name="unixMillisecondsTimestamp"/> with Utc Kind.
         /// </returns>
         public static DateTime UnixTimestampMsToDateTime(long unixMillisecondsTimestamp)
             => UnixTimeEpoch + TimeSpan.FromMilliseconds(unixMillisecondsTimestamp);
--- a/src/Confluent.Kafka/Timestamp.cs
+++ b/src/Confluent.Kafka/Timestamp.cs
@@ -32,7 +32,7 @@ namespace Confluent.Kafka
         /// </summary>
         public static readonly DateTime UnixTimeEpoch = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
 
-        private const long UnixEpochMilliseconds = 62135596800000; // = UnixEpochUtcDateTime.TotalMiliseconds
+        private const long UnixEpochMilliseconds = 62135596800000; // = UnixTimeEpoch.TotalMiliseconds
 
         /// <summary>
         ///     Initializes a new instance of the Timestamp structure.
@@ -84,7 +84,8 @@ namespace Confluent.Kafka
         /// <summary>
         ///     Gets the Utc DateTime corresponding to the <see cref="UnixTimestampMs"/>.
         /// </summary>
-        public DateTime DateTime => UnixTimestampMsToDateTime(UnixTimestampMs);
+        public DateTime DateTime
+            => UnixTimestampMsToDateTime(UnixTimestampMs);
 
         public override bool Equals(object obj)
         {
@@ -98,7 +99,7 @@ namespace Confluent.Kafka
         }
 
         public override int GetHashCode()
-            => Type.GetHashCode()*251 + UnixTimestampMs.GetHashCode();  // x by prime number is quick and gives decent distribution.
+            => Type.GetHashCode() * 251 + UnixTimestampMs.GetHashCode();  // x by prime number is quick and gives decent distribution.
 
         public static bool operator ==(Timestamp a, Timestamp b)
             => a.Equals(b);
@@ -115,15 +116,10 @@ namespace Confluent.Kafka
         /// </param>
         /// <returns>
         ///     The milliseconds unix timestamp corresponding to <paramref name="dateTime"/>
-        ///     rounded to the previous millisecond if dateTime precision is below millisecond
+        ///     rounded down to the previous millisecond.
         /// </returns>
         public static long DateTimeToUnixTimestampMs(DateTime dateTime)
-        {
-            checked
-            {
-                return dateTime.ToUniversalTime().Ticks / TimeSpan.TicksPerMillisecond - UnixEpochMilliseconds;
-            }
-        }
+            => dateTime.ToUniversalTime().Ticks / TimeSpan.TicksPerMillisecond - UnixEpochMilliseconds;
 
         /// <summary>
         ///     Convert a milliseconds unix timestamp to a DateTime value.
--- a/src/Confluent.Kafka/Timestamp.cs
+++ b/src/Confluent.Kafka/Timestamp.cs
@@ -31,6 +31,8 @@ namespace Confluent.Kafka
         /// </summary>
         public static readonly DateTime UnixEpochUtcDateTime = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
 
+        private const long UnixEpochMilliseconds = 62135596800000; // = UnixEpochUtcDateTime.TotalMiliseconds
+
         /// <summary>
         ///     Initializes a new instance of the Timestamp structure.
         /// </summary>
@@ -85,12 +87,13 @@ namespace Confluent.Kafka
         /// </param>
         /// <returns>
         ///     The milliseconds unix timestamp corresponding to <paramref name="dateTime"/>
+        ///     rounded to the previous millisecond if dateTime precision is below millisecond
         /// </returns>
         public static long DateTimeToUnixTimestampMs(DateTime dateTime)
         {
             checked
             {
-                return (long)(dateTime.ToUniversalTime() - UnixEpochUtcDateTime).TotalMilliseconds;
+                return dateTime.ToUniversalTime().Ticks / TimeSpan.TicksPerMillisecond - UnixEpochMilliseconds;
             }
         }
 
--- a/src/Confluent.Kafka/Timestamp.cs
+++ b/src/Confluent.Kafka/Timestamp.cs
@@ -27,9 +27,10 @@ namespace Confluent.Kafka
     public struct Timestamp
     {
         /// <summary>
-        ///     Unix epoch as UTC DateTime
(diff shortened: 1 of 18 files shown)


mhowlett:
I just had a quick look at this and my initial reaction is yes, I like it a lot - it's a much better factoring. Will do a closer review a bit later.

src/Confluent.Kafka/Timestamp.cs
```diff
@@ -27,6 +27,11 @@ namespace Confluent.Kafka
     public struct Timestamp
     {
         /// <summary>
+        ///     Unix epoch as UTC DateTime
+        /// </summary>
+        public static readonly DateTime UnixEpochUtcDateTime = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
```
mhowlett: I think my favourite name for this is just `UnixTimeEpoch`.

src/Confluent.Kafka/Timestamp.cs
```diff
@@ -27,6 +27,11 @@ namespace Confluent.Kafka
     public struct Timestamp
     {
         /// <summary>
+        ///     Unix epoch as UTC DateTime
```
mhowlett: Maybe: "Unix time is defined as the number of seconds past this UTC time, excluding leap seconds."

src/Confluent.Kafka/Timestamp.cs
```diff
@@ -85,12 +87,13 @@ public override int GetHashCode()
...
         /// <returns>
         ///     The milliseconds unix timestamp corresponding to <paramref name="dateTime"/>
+        ///     rounded to the previous millisecond if dateTime precision is below millisecond
         /// </returns>
         public static long DateTimeToUnixTimestampMs(DateTime dateTime)
         {
             checked
             {
-                return (long)(dateTime.ToUniversalTime() - UnixEpochUtcDateTime).TotalMilliseconds;
+                return dateTime.ToUniversalTime().Ticks / TimeSpan.TicksPerMillisecond - UnixEpochMilliseconds;
```
mhowlett: now you have me wondering if we should be rounding to the nearest millisecond or rounding down. I'm currently favouring rounding down (as I had it before, and as you continue going forward) because I think this type of binning is easier to think about and I prefer it if times never get adjusted into the future. Somewhat arbitrary as I don't expect this level of precision is really ever required in practice.

src/Confluent.Kafka/Timestamp.cs
```diff
@@ -85,12 +87,13 @@ public override int GetHashCode()
         /// </param>
         /// <returns>
         ///     The milliseconds unix timestamp corresponding to <paramref name="dateTime"/>
+        ///     rounded to the previous millisecond if dateTime precision is below millisecond
```
mhowlett: i'd delete 'if dateTime precision is below millisecond'

src/Confluent.Kafka/Timestamp.cs
treziac: Rounding down is how microsoft do it in DateTimeOffset.ToUnixTimestamp (we could actually retake their function, but not available on net45 and would require an intermediate DateTimeOffset)

Same than you, I never except to have below millisecond precision (except when building from a DateTime.Now as it's often not rounded tick) and don't want to project in future.

I'm wondering if we should keep the DateTime constructor or simply remove it (depreciate first). It should never be used and is misleading with the datetime comparaison

test/Confluent.Kafka.UnitTests/Timestamp.cs
```diff
@@ -64,5 +64,30 @@ public void Conversion()
             Assert.Equal(ts, ts2);
             Assert.Equal(ts2.Kind, DateTimeKind.Utc);
         }
+
+        [Fact]
+        public void Rounding()
+        {
```
mhowlett: you could add a test to subtract a tick from a millisecond aligned datetime and see that it rounds down (optional - only if you feel like it).

src/Confluent.Kafka/Timestamp.cs
treziac: UinxTimeEpoch or UnixTimestampEpoch? (we may want to be consistant with other methods)
Also wanted to specify it is Utc as it's a Datetime, and Utc vs Local vs Unspecified is misleading

mhowlett: `UnixTimeEpoch` is more correct i believe - this is a thing related to unix time, not a specific timestamp. The word epoch already covers that it is a specific time. I don't think there is any naming inconsistency.

I think it's better to exclude Utc from the name as this is part of the definition of what unix time is. Including it in the name to me might imply there is some other way this could have been reasonably specified (which there is not). Also, I don't think there is any likely confusion as this is in the summary commend and it's also a property on the DateTime instance. If I thought there was likely a scenario in which this was likely to trip people up, i'd have been in favor of including this in the name, but I think that's not the case (its possible I'm wrong on that).

src/Confluent.Kafka/Timestamp.cs
mhowlett: > I'm wondering if we should keep the DateTime constructor or simply remove it (depreciate first). It should never be used and is misleading with the datetime comparaison

can you argue further why it should never be used and why it's misleading :-) - i can't see this atm.

src/Confluent.Kafka/Timestamp.cs
```diff
@@ -36,16 +36,38 @@ public struct Timestamp
...
+        /// </param>
+        public Timestamp(long unixTimestampMs, TimestampType type)
+        {
+            Type = type;
+            UnixTimestampMs = unixTimestampMs;
+        }
+
+        /// <summary>
+        ///     Initializes a new instance of the Timestamp structure.
+        ///     Prefer the (long, TimestampType) overload if possible.
```
mhowlett: I'd remove the line 'Prefer the (long, TimestampType) overload if possible.', (or if I'm wrong and missing something here, then explain why the other overload is preferred as well).

src/Confluent.Kafka/Timestamp.cs
```diff
@@ -54,9 +76,14 @@ public Timestamp(DateTime dateTime, TimestampType type)
...
-        ///     Gets the timestamp value.
+        ///     Get the Unix millisecond timestamp.
+        /// </summary>
+        public long UnixTimestampMs { get; }
+
+        /// <summary>
+        ///     Gets the Utc DateTime corresponding to the <see cref="UnixTimestampMs"/>.
         /// </summary>
-        public DateTime DateTime { get; }
+        public DateTime DateTime => UnixTimestampMsToDateTime(UnixTimestampMs);
```
treziac: Ok
Oh, given this, do we accept C#7 ? Do we have a convention of when using this notation? (say less than 1 line / x characters / 1 instruction ?)

src/Confluent.Kafka/Timestamp.cs
treziac: That was because the DateTime in constructor will be likely different to the DateTime property (Utc conversion or rounding) Even with the remarks, it doesn't seem right to me that constructor use a different object than the one given in parameter.

I even wondered about deprecating this method (which we never call with DateTime now, Timestamp class should represent a librdkafka Timestamp, which is implemented from unix ms timestamp) Didn't remove it to not break api (yes, sometime I care about this ;) )

test/Confluent.Kafka.UnitTests/Confluent.Kafka.UnitTests.csproj
```diff
@@ -17,4 +17,8 @@
     <PackageReference Include="xunit.runner.visualstudio" Version="2.2.0" />
   </ItemGroup>
 
+  <ItemGroup>
```
mhowlett: Don't think we want this:
http://stackoverflow.com/questions/18614342/what-is-service-include-in-a-csproj-file-for

test/Confluent.Kafka.IntegrationTests/Confluent.Kafka.IntegrationTests.csproj
```diff
@@ -24,4 +24,8 @@
     <PackageReference Include="Newtonsoft.Json" Version="9.0.1" />
   </ItemGroup>
 
+  <ItemGroup>
```
mhowlett: intentionally here?

src/Confluent.Kafka/Timestamp.cs
mhowlett: i've been doing it for methods without braces.

Haven't looked into details of C# 7, but we're happy to require the latest released tools for development. We'd will need very good reason to stop compatibility with netstandard1.3 or net45 for usage of the library/nuget package however.

src/Confluent.Kafka/Timestamp.cs
mhowlett: Ok - i see where you're coming from. I think having the `DateTime` constructor will be convenient when people are specifying timestamps to `ProduceAsync`. I'd be happy to depreciate the `DateTime` property - I don't have a strong opinion on that, but maybe removing it is slightly better.

mhowlett: actually, since there was a bug that made all Timstamp values incorrect in the released version, I think the DateTime property could just be removed (i.e. skip depreciating it).

test/Confluent.Kafka.UnitTests/Confluent.Kafka.UnitTests.csproj
treziac: Forgot to remove it, It does just add it by itself in VS2017 when using test explorer (kind of guid) Looked for a way to tell the project is a unittest (like in the stack overflow link) but no luck so far on .netcoreapp

src/Confluent.Kafka/Timestamp.cs
treziac: To be more clear : 
- Timestamp struct are created only from librdkafka return, so remove the constructor with datetime is ok.
- Message (returned by consumer and delivery handler) contains TimeStamp, so I would keep the datetime property for easy usage
- ProduceRecord contains (will contain, I couldn't do both in atomic PR) a DateTime and UnixTimestampMs set property

What I don't like about dateTime in .NET is this:
if(DateTime.Now.AddMinutes(1) > timestamp.DateTime)
return unexpected value when kind is not the same (depend on timezone) I know it's user side to be sure he uses correct kind on both side, but DateTimeOffset resolve this (and it has DateTime / DateTimeUtc property). We  could do the same with a LocalDateTime, UtcDateTimeUtc property here. Microsoft actually advise using DateTimeOffset for utc datetime


https://msdn.microsoft.com/library/bb384267(v=vs.110).aspx
> These uses for DateTimeOffset values are much more common than those for DateTime values. As a result, DateTimeOffset should be considered the default date and time type for application development.

mhowlett: > Message (returned by consumer and delivery handler) contains TimeStamp, so I would keep the datetime property for easy usage

the released version contains a / 1000 bug (in two places) which means timestamp values are always wrong (fixed in master, but nothing released yet) - this means no one is relying on this yet, so I think we can get away with just changing it... however obsolete is fine too, I wouldn't bother changing it.

Perhaps we should have a constructor that takes `DateTimeOffset` as well? (feel free to add, but I'll still merge without).

mhowlett:
some changes: 

1. I removed the `LocalDateTime` and `DateTimeOffset` properties as `DateTimeOffset` apparently inserts some local timezone magic, which I think should be avoided here.
2. I added some missing comments.

I've pushed the changes to my (mhowlett) timezone branch. I can merge this, or you can give me write permission to your repo and I can push to yours. The reason i suggest the latter is I don't know if merging from my branch will affect attribution.

treziac:
Ok for me :+1: You can merge from your branch, commit attribution should stay ok

mhowlett:
closing - PR #138 (which is about to be merged) extends this PR with some tidy-up.

mhowlett:
this did affect attribution (because of squash before merge) - feel free to give me write access to your confluent-kafka-dotnet repo - sometimes it will be a quicker workflow for me to just push commits rather than explain everything in words...

treziac:
Done, thanks!

PR #121, opened 2017-03-24
Change produce call from rd_kafka_produce to rd_kafka_producev

treziac:
Part of #113 rebased and with only producev concern

rd_kafka_producev is the new way to send mesage. It also allow sending custom timestamp.
This will give TypeInitializeException if version of kafka not supporting producev is used

Didn't benchmark it
Will allow to use timestamp in ProduceRecord once #118 is included

I'm not sure we should completly remove producev. Perhaps add a boolean (like BlockIfQueueFull) to say which function we want to use? (produce would be the legacy api, automatically used if librdkafkaversion is < 0.9.4)


mhowlett:
It'll be fine to require librdkafka >= 0.9.4. It'll be rare that anyone uses anything other than the referenced librdkafka.redist nuget package.

after a quick look through this is looking good. will review in depth a bit later.

treziac:
Note I won't have computer for one week, so feel free to add any commits to the PR

I didn't test a lot, just that I could effectively send/receive message.


Le 25 mars 2017 2:54 AM, Matt Howlett <notifications@github.com> a écrit :

It'll be fine to require librdkafka >= 0.9.4. It'll be rare that anyone uses anything other than the referenced librdkafka.redist nuget package.

after a quick look through this is looking good. will review in depth a bit later.

treziac:
Some extra note:
I didn't touch the Proucer.Produce signature (with nullable DateTime) as it's in conflict with the new produceAsync api, this PR only modify the internal call to librdkafka in SafeKafkaHandle.

We can't implement variable argument in C# in the same way that in C (there was an undocument __arglist before but it's not available for the moment in netstandard, and undocumented so not really want to use it.
There is two choices then: have multiple binding with multiple argumentList (with or without timestamp, with or without opaque, etc) or only one method providing default values (firs is buggy for now on partition so will keep the one method option - see https://github.com/edenhill/librdkafka/issues/1153 )

mhowlett:
> I didn't touch the Proucer.Produce signature (with nullable DateTime) as it's in conflict with the new produceAsync api, this PR only modify the internal call to librdkafka in SafeKafkaHandle.

cool. i have a funny feeling the `Producer.ProduceAsync` signature is going to take some time to think over / argue about :-). Lets get the foundations in first.

> We can't implement variable argument in C# in the same way that in C (there was an undocument __arglist before but it's not available for the moment in netstandard, and undocumented so not really want to use it.

yep, I was definitely anticipating having binding(s) with a fixed number of args.

> There is two choices then: have multiple binding with multiple argumentList (with or without timestamp, with or without opaque, etc) or only one method providing default values (firs is buggy for now on partition so will keep the one method option - see https://github.com/edenhill/librdkafka/issues/1153 )

That sounds good. I think there is a decent chance we'll eventually want to have multiple variants for performance reasons as marshalling is relatively expensive (though we should measure), however a single method for now will be fine.

treziac:
> cool. i have a funny feeling the Producer.ProduceAsync signature is going to take some time to think over / argue about :-). Lets get the foundations in first.

Agreed 🍿

src/Confluent.Kafka/Impl/LibRdKafka.cs
```diff
@@ -709,24 +731,22 @@ private class NativeMethods
...
-                IntPtr rkt,
-                int partition,
-                IntPtr msgflags,
-                IntPtr val, UIntPtr len,
-                IntPtr key, UIntPtr keylen,
-                long timestamp,
-                IntPtr msg_opaque);
+            internal static extern ErrorCode rd_kafka_producev(
+                IntPtr rk,
+                ProduceVarTag topicType, string topic,
```
mhowlett: i'm worrying about the marshaling of topic name here.
with the old produce method (where topic handles were required instead of topic name) the new topic method has the topic name parameter decorated with the `[MarshalAs(UnmanagedType.LPStr)]` attribute.
But is even that exactly what we want? it results in the topic name being encoded as 'ANSI' - whatever that means - seems a bit dodgy to me.
I'm wondering if ideally what we want is to manually encode as UTF8 into a byte array (and terminate with 0) and pass that.
This would also apply to other methods that marshal strings.
For this PR, decorating with `[MarshalAs(UnmanagedType.LPStr)]` is probably enough (questioning string marshaling is probably a separate PR).

src/Confluent.Kafka/Impl/LibRdKafka.cs
```diff
@@ -709,24 +731,22 @@ private class NativeMethods
...
                     IntPtr rk, IntPtr partitions);
 
+            // note: producev signature is rd_kafka_producev(rk, ...)
+            // __arglist is an undocumented keyword which deal with it
+            // but doesn't work on standardlib (and undocumented so don't use)
+            // stay simple with one binding for now, but it will be worth benchmarking
+            // the overload with no timestamp, opaque, partition, etc
             [DllImport(DllName, CallingConvention = CallingConvention.Cdecl)]
-            internal static extern IntPtr rd_kafka_produce(
-                    IntPtr rkt,
```
mhowlett: it seems a pity to require a memory allocation + string serialization for every produce call now with producev :-( - @edenhill would be possible to make the `rd_kafka_producev` method alternatively take a topic handle? i guess this is super low priority - the overall perf hit will be negligible (though i think it's probably measurable).

mhowlett: actually, it would be possible to do a lookup into a dictionary of pre-serialised byte arrays instead of one of topic handles (though, come to think of it, that is not especially cheap with the string key, so we won't). anyway, my point is no need for producev to ever accept topic handles, please ignore my above comment.

treziac: There already is a rkt overload in producev. I didn't use it cause I thought we wanted get rid of the rkt (from comments on getTopicHandle) but we will need it anyway if we want to pass a custom topic on a topic. Will update the PR

mhowlett: no - no need to change.

mhowlett:
I think this just needs rebasing onto master (new Timestamp constructor) then it will be good to go

treziac:
Anything missing? This may need to be properly benchmarked (this was ok in my tests from what I remember but will have to redo it and post results)

Also, we may want to be able to create topic by ourself by passing a config in an other method. Currently with this api, there is no way to create a new topic dynamically with a custom conf. I didnd't investigate yet if it was possible without dealing with the topic handle created (so that it' automatically removed when client handle is disposed, without having to dispose the topic handle ourself) 
Should not be a problem to keep the handle anyway as it was done before (and as it is done for metadata) but if we could let all this to librdkafka the better)

src/Confluent.Kafka/Impl/LibRdKafka.cs
```diff
@@ -766,24 +790,21 @@ private class NativeMethods
...
-                IntPtr rkt,
-                int partition,
-                IntPtr msgflags,
-                IntPtr val, UIntPtr len,
-                IntPtr key, UIntPtr keylen,
-                long timestamp,
-                IntPtr msg_opaque);
+            internal static extern ErrorCode rd_kafka_producev(
+                IntPtr rk,
+                ProduceVarTag topicType, [MarshalAs(UnmanagedType.LPStr)] string topic,
```
mhowlett: it would be better if topic name was encoded as UTF-8, however we can leave this for a later PR (same comment applies elsewhere in the lib).

mhowlett:
just rebased and fixed merge issues. I haven't checked the implementation super carefully but I think this is probably almost good to go. @treziac, at this point you can feel free to give this another review. I'm also interested to run the Producer performance test to see if much has changed by moving to `producev`. Thinking the difference will be negligible, but this might actually be slightly faster.

treziac:
I didn't check since we did the change in runasynchronously (but it shouldn't impact), will recheck properly on monday

mhowlett:
I ran the benchmarks and can't differentiate performance difference with the previous implementation from the noise. 

Code still LGTM.

Issue #127, opened 2017-03-29
Consumer methods using ICollection parameters instead of IEnumerable


MrGlenThomas:
I have just noticed that most of the methods of the Consumer<TKey, TValue> class take ICollection parameters. I think it would make a lot of cases simpler to use IEnumerable instead, otherwise having to convert IEnumerables (e.g. arrays) to ICollection types and the internal implementation should not require ICollection...

```
public void SubscribeTopics(Consumer<long, string> consumer, IEnumerable<string> topics)
{
    consumer.Subscribe(topics.ToList());
}
```

mhowlett:
Sounds good to me (though maybe there is something I'm missing). This decision was inherited from rdkafka-dotnet - I've not considered it in any depth.

`ICollection` inherits from `IEnumerable` so we don't need to worry about breaking the API.

treziac:
Some API were changed from ICollection to IEnumerable but all weren't modified so yes, we will have to change this

mhowlett:
closing this as there is PR #128 for it now.

Issue #101, opened 2017-03-14
Travis CI failing with 'too many open files' on MacOS


mhowlett:
Occurs consistently but at seemingly random times whilst `dotnet restore` is executing.
Looks like this is actually a known issue: https://github.com/dotnet/cli/issues/809
This was working. Wondering if whether or not this issue occurs is dependent on time of day (i.e. load on server) - when I got it working, it was late on a Friday evening PST time.
Two things to try:
 - `ulimit` (though does this require root privileges, which we don't have?)
 - `dotnet restore --disable-parallel`

lou1306:
In my case, specifying `--disable-parallel` was enough to work around the issue.

PR #128, opened 2017-03-30
Changed ICollection parameters to IEnumerable

MrGlenThomas:
As described in #127. Using ICollection forces user to convert any not ICollection type before passing as a parameter. IEnumerable provides compatibility with more collection types.


test/Confluent.Kafka.UnitTests/Confluent.Kafka.UnitTests.csproj
```diff
@@ -17,4 +17,8 @@
     <PackageReference Include="xunit.runner.visualstudio" Version="2.2.0" />
   </ItemGroup>
 
+  <ItemGroup>
```
treziac: Remove this, will mark as Unit test project in an other PR

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -258,9 +258,9 @@ internal WatermarkOffsets GetWatermarkOffsets(string topic, int partition)
             return new WatermarkOffsets(low, high);
         }
 
-        internal void Subscribe(ICollection<string> topics)
+        internal void Subscribe(IEnumerable<string> topics)
         {
-            IntPtr list = LibRdKafka.topic_partition_list_new((IntPtr) topics.Count);
+            IntPtr list = LibRdKafka.topic_partition_list_new((IntPtr) topics.Count());
```
treziac: Quick though on it I didn't mention first: I tend to prefer using ICollection when using Count (linq count is not good on pure IEnumerable, it may just hang forever if it's an infinite enumerable for example)

Though we are enumerating just after, so it is ok in this case

MrGlenThomas:
It looks like the CI build failed to publish a nuget package? Can it be run again?

treziac:
Related to #101 , it's okay I think,
There is no ci published from my knowledge (note I'm not from confluent)

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
mhowlett: `Count()` is less efficient in general of course, though after a quick scan through I don't see anywhere this is going to be of practical concern. This PR is LGTMing, but i'm going to hold off merging 'till I have a bit more time to review carefully.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -422,12 +422,12 @@ public Task<CommittedOffsets> CommitAsync()
         ///     Hence, this method commits an offset of <paramref name="message" />.Offset + 1.
         /// </remarks>
         public Task<CommittedOffsets> CommitAsync(Message<TKey, TValue> message)
-            => consumer.CommitAsync(new List<TopicPartitionOffset> { new TopicPartitionOffset(message.TopicPartition, message.Offset + 1) });
+            => consumer.CommitAsync(new[] { new TopicPartitionOffset(message.TopicPartition, message.Offset + 1) });
```
mhowlett: this change is a good observation

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
mhowlett: In reply to my own comment above, the LINQ `Count()` implementation is evidently fast if the collection is an `ICollection`: http://stackoverflow.com/questions/168901/howto-count-the-items-from-a-ienumerablet-without-iterating

... but now I'm wondering under what circumstances anyone would ever want to use a data structure that doesn't implement `ICollection` in these methods? @treziac has a good point - using `IEnumerable` opens up the possibility of an infinite loop. Seems unlikely any one would use it in this way, but if they're passing in something that isn't an `ICollection`, well, I think i would be inclined to worry about it a little bit.

So now i'm thinking that rather than changing from `ICollection` -> `IEnumerable`, we should do an audit of everywhere we're using `IEnumerable` and possibly think about changing to `ICollection`. 

Not that this is a big deal. Chances of anyone running into it are minimal.

treziac: IEnumerable is the default return of Linq. This avoid a call to ToList or ToArray

mhowlett: oh yeah, good point.

MrGlenThomas: I understand the reason for ICollection, but exposes the internal implementation of the method in a very minor way by showing that the collection requires to be counted. Having thought about it for a while, I'm not sure what is best..

I suppose that inside the method, to get the count you could do something like
`var partitionCollection = partitions as ICollection ?? partitions.ToArray();`

This would allow the method to support IEnumerable, but automatically handle deferred execution types. BUT, it seems like a lot of extra effort/maintenance just for the simple benefit of supporting IEnumerable.

MrGlenThomas: Also, I think ICollection<T> is not good because it requires a collection type that supports Add(T item), Clear(), Contains(T item), bool Remove(T item), which seems overkill. Personally I like to used fix-size collection types in most scenarios.

treziac: The infinite problem I stated was only "in general". And if someone provide an infinite enumerable in the parameters, I would say it's his problem ;) 

IEnumerable is the standard for collection (new List<T>(IEnumerable<T>) for example. If you try to create a List from an infinite IEnumerable, it will just hang. So standard way =D

mhowlett: yeah - you just needed to remind me `IEnumerable` is prevalent with LINQ. Anyway, this PR should be good without changes, hopefully merge on the weekend.

mhowlett:
LGTM

PR #129, opened 2017-03-30
Change ISerializer and IDeserializer to include topic

treziac:
This will eventually come with the schema registry support. 
Add this now so that we can already bind to schema registry (with other open sources project or custom ones)

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -95,8 +95,8 @@ namespace Confluent.Kafka
                     e.Topic,
                     e.Partition,
                     e.Offset,
-                    KeyDeserializer.Deserialize(e.Key),
-                    ValueDeserializer.Deserialize(e.Value),
+                    KeyDeserializer.Deserialize(e.Topic, e.Key),
+                    ValueDeserializer.Deserialize(e.Topic, e.Value),
                     e.Timestamp,
                     e.Error
                 )
--- a/src/Confluent.Kafka/Producer.cs
+++ b/src/Confluent.Kafka/Producer.cs
@@ -836,8 +836,8 @@ namespace Confluent.Kafka
         private Task<Message<TKey, TValue>> Produce(string topic, TKey key, TValue val, DateTime? timestamp, int partition, bool blockIfQueueFull)
         {
             var handler = new TypedTaskDeliveryHandlerShim(key, val);
-            var keyBytes = KeySerializer?.Serialize(key);
-            var valBytes = ValueSerializer?.Serialize(val);
+            var keyBytes = KeySerializer?.Serialize(topic, key);
+            var valBytes = ValueSerializer?.Serialize(topic, val);
             producer.Produce(topic, valBytes, 0, valBytes == null ? 0 : valBytes.Length, keyBytes, 0, keyBytes == null ? 0 : keyBytes.Length, timestamp, partition, blockIfQueueFull, handler);
             return handler.Task;
         }
@@ -889,8 +889,8 @@ namespace Confluent.Kafka
         private void Produce(string topic, TKey key, TValue val, DateTime? timestamp, int partition, bool blockIfQueueFull, IDeliveryHandler<TKey, TValue> deliveryHandler)
         {
             var handler = new TypedDeliveryHandlerShim(key, val, deliveryHandler);
-            var keyBytes = KeySerializer?.Serialize(key);
-            var valBytes = ValueSerializer?.Serialize(val);
+            var keyBytes = KeySerializer?.Serialize(topic, key);
+            var valBytes = ValueSerializer?.Serialize(topic, val);
             producer.Produce(topic, valBytes, 0, valBytes == null ? 0 : valBytes.Length, keyBytes, 0, keyBytes == null ? 0 : keyBytes.Length, timestamp, partition, blockIfQueueFull, handler);
         }
 
--- a/src/Confluent.Kafka/Serialization/Extensions/Message.cs
+++ b/src/Confluent.Kafka/Serialization/Extensions/Message.cs
@@ -42,8 +42,8 @@ namespace Confluent.Kafka.Serialization
                 message.Topic,
                 message.Partition,
                 message.Offset,
-                keyDeserializer.Deserialize(message.Key),
-                valueDeserializer.Deserialize(message.Value),
+                keyDeserializer.Deserialize(message.Topic, message.Key),
+                valueDeserializer.Deserialize(message.Topic, message.Value),
                 message.Timestamp,
                 message.Error
             );
--- a/src/Confluent.Kafka/Serialization/IDeserializer.cs
+++ b/src/Confluent.Kafka/Serialization/IDeserializer.cs
@@ -26,6 +26,9 @@ namespace Confluent.Kafka.Serialization
         ///     Deserialize a byte array to an instance of
         ///     type T.
         /// </summary>
+        /// <param name="topic">
+        ///     The topic associated wih the data.
+        /// </param>
         /// <param name="data">
         ///     The serialized representation of an instance
         ///     of type T to deserialize.
@@ -33,6 +36,6 @@ namespace Confluent.Kafka.Serialization
         /// <returns>
         ///     The deserialized value.
         /// </returns>
-        T Deserialize(byte[] data);
+        T Deserialize(string topic, byte[] data);
     }
 }
--- a/src/Confluent.Kafka/Serialization/IDeserializer.cs
+++ b/src/Confluent.Kafka/Serialization/IDeserializer.cs
@@ -14,13 +14,15 @@
 //
 // Refer to LICENSE for more information.
 
+using System;
+
 namespace Confluent.Kafka.Serialization
 {
     /// <summary>
     ///     Implement this interface to define a deserializer 
     ///     for a particular type T.
     /// </summary>
-    public interface IDeserializer<T>
+    public interface IDeserializer<T> : IDisposable
     {
         /// <summary>
         ///     Deserialize a byte array to an instance of
--- a/src/Confluent.Kafka/Serialization/ISerializer.cs
+++ b/src/Confluent.Kafka/Serialization/ISerializer.cs
@@ -25,12 +25,15 @@ namespace Confluent.Kafka.Serialization
         /// <summary>
         ///     Serialize an instance of type T to a byte array.
         /// </summary>
+        /// <param name="topic">
+        ///     The topic associated wih the data.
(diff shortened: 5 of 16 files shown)


mhowlett:
awesome - lots of people are looking for schema registry integration so work towards this particularly great.

after a quick scan through this is LGTMing - i'll get to a careful review soon.

treziac:
However this is breaking change in API, and nothing we can depreciate properly.

How do you guys plan to version the project? 
We are in 0.9.x but timestamp are on kafka 0.10 and we support them, so it doesn't seem directly related.
If we keep following the librdkafka versioning, this doesn't respect simver usually used on nuget (librkdakfka does, but we don't necesseraly). And we may want to produce change on api even without librdkafka release

Do you plan to follow librdkafka versioning or go for simver? 

For now as API is still unstable, we can keep the 0.x, but this would make more sense for a c# developper who doesn't know librdkafka to have a simver versioning (and easier for breaking changes)

treziac:
@mhowlett 
Did you have the chance to review this? It will really help to add Avro serialization simply

mhowlett:
Can you explain the reasoning behind choosing to use explicit interface implementations?

I think breaking changes could be avoided using a second interface and either a runtime check or method overloads, though I don't think it's worth all the hassle. The changes will impact only people who have written their own serdes, and the magnitude of annoyance should be relatively low.

We're also going to want the (de)serializer interfaces to inherit from `IDisposable` i think, and i'm thinking it probably makes sense to do this now as well (have all the breaking changes at once). I'm thinking that serdes passed to a producer or consumer should be owned by the producer or consumer, and disposed automatically when the producer or consumer is disposed.

Schema registry connected serdes will require configuration. I think I'm probably favor having this as a constructor argument of type `IEnumerable<KeyValuePair<string, object>> config` on the relevant serdes (in practice, serde config will be amalgamated with other config). Alternatively, the API could be terser if the producer or consumer passed the global config through in their constructors (requiring a `config` method or something on the serde interfaces), however this is a bit more magical.

treziac:
> Can you explain the reasoning behind choosing to use explicit interface implementations?

To not explose methods which are here only for interface compliance on instances. For example, using a StringSerializer instance, you can't write
`serializer.Serialize("my string", "my topic")` but only `serializer.Serialize("my string") `as topic is not used and is only here to comply with the interface (so we can still do (`(ISerializer<string)serializer).Serialize("my string", "mytopic") `or use with a ISerializer) Of course it will be explicit on class using topic (like AvroSerializer)


> I think breaking changes could be avoided using a second interface and either a runtime check or method overloads, though I don't think it's worth all the hassle. The changes will impact only people who have written their own serdes, and the magnitude of annoyance should be relatively low.

If we make a ISerializerWithTopic interface, what would be the signature for Producer<K,V> constructor?
If ISerializerWithTopic implements ISerializer, in case where topic is needed the Serialize(V data) will be incorrect as topic is needed. If we don't implement it, we will have to make constructor which accept all combination plus the switch on runtime on the interface implemented in Produce method (if KeySerializer is ISerializer.. else if is ISerializerWithTopic...). The java producer only have the interface with topic, so that's your choice (perhaps we should benchmark it)

qed-:
I'm not sure if it's strictly necessary for an API change to support AVRO and the schema registry.

For the producer case you already know the topic when you create the producer, so the serializer can be created at this point which will register the schema and retrieve the ID.

The consumer will receive the schema registry magic byte followed by the schema ID at the beginning of each message, which it can use to resolve the schema using the registry.

I have some prototype code that does this and seems to work OK.

treziac:
I agree it's possible to do this without the topic (you can create a serializer with the topic in the GetSerializingProducer) but I would rather have a single SerializingProducer<K,V> which produce to multiple topics than multiple SerializingProducer where each one is binded to a single topic (or we could create a GetSerializingProduce<K,V>(string topic, ISerializer<K>, ISerializer<V>) where produce wouldn't take the topic, it's also an API option, though I don't really like it)

But note there is not only the confluent avro serializer. People may chose protobuf or homemade serializer/deserializer, already implemented in java for example (where desers works with topics), and for deserialization sometimes you have to depends on topic (and with current API and the subscribe/assign mechanism, you can't do the same as in producer without impact performance)

Also the goal for me is to be as generic as possible and not bind to confluent schema registry for me (note I'm not a confluent employee) so that it can be used with any kafka deployment and not just with confluent platform.

treziac:
Rebased on master to avoid conflict on comments

treziac:
Given your remarks for serdes and conf @mhowlett, how do you see the creation of the serializer inside the producer (with config?)

producer.GetSerializingProducer<Avro, Avro>(config) ?
How would config work with other serializers?

I would keep with the current implementation, ie `producer.GetSerializingProducer<Avro, Avro>(avroKeySeralizer, avroValueSerializer)` where `avroKeySerializer` and `avroValueSerializer` are created before (we need different here to distinct from key and value), and add an helper for avro `producer.GetAvroSerializingProducer()` which would take the conf directly from producer. This should be in a separate nuget for me (Btw, you already published a Confluent.Kafka.Avro on nuget but I cant find the corresponding repo on github)

treziac:
Added the IDisposable interface. Didn't introduce the code in producer/consumer, this should be a separate PR

I also used explicit implementation here as the existing serializer don't use disposable ressources (the Dispose and De-Serialize(topic, data) methods only exist on the interface scope)

mhowlett:
@qed- re: producer - I think what you suggest is a valid option to consider because it aligns with common usage scenarios and it's simple, however I think consistency with the `ProduceAsync` method which has topic as a required parameter is more important (i.e. in agreement with @treziac).

re: consumer - yeah, unnecessary in this case, but including topic in the interface may be important in other scenarios (as @treziac also points out). Also, it follows the java deserializer interface.

treziac:
I could still try to implement it with both interface and benchmark it, this won't cost much and if we can prevent API breaking changes it would be better - will try to do it this we, don't merge yet

mhowlett:
TBH I'm actually a bit uncomfortable expanding the purpose of the serializers as we're doing (because classes should remain as focussed as possible - the proposal is conflating ideas), even though it mirrors how the java client works. Maybe the reason you're using explicit interface implementation is that you are too. I don't think this is a good idea though, because our primary usage of the serializers (their purpose) is to use the (de)serialization methods that take a topic and value, not just a value - so this shouldn't be hidden. 

I'm currently thinking about something along the lines of keeping the serializers as they are, and creating a two new interfaces `ITopicValue(De)Serializer` (note: really bad names - need something better - `IExtended(De)Serializer`? ). We could have overloads of the producer/consumer constructors that accept these as well as `I(De)Serializer` (i.e. 4 overloads in total each). If a constructor that takes a basic serializer is used, some boilerplate wrapper is constructed in the constructor that uses the serializer but ignores the passed in topic. this is then used in the `ProduceAsync` method. Alternatively, `IExtended(De)Serializer` could derive from `I(De)Serializer` and a runtime check could be done in the constructor to see what we have. This avoids the constructor overloads but isworse though IMO because the API doesn't make it's capability clear - too magical.

note: half formed thoughts... haven't thought it through fully... time pressure to think about other things.. just noting thinking so far.

treziac:
IDeserializerWithTopic? All seem ok for me, pick the one you want

I see many ways if we don't want to break API (so don't modify the I(De)Serializer interface): 

- Have the two interfaces possible as argument (this makes 4 constructor), and have properties `KeySerializer`, `KeyWithTopicSerializer`, `ValueSerializer`, `ValueWithTopicSerializer`. Then, 
```keyBytes = KeySerializer?.Serialize( key) ?? KeyWithTopicSerializer?.Serializer(topic, key);```

- Have the 4 overloads, but only have the KeyWithTopic(De)Serializer properties in Producer/consumer. Just create a proxy which implements the (topic, data) method using the (data) from I(De)Serializer. But this means one more call on the stack each time we serialize / deserialize

Derive from IDeserializer won't work for me: if we need IDeserializerWithTopic, we can't do anything with the method without topic except throwing, so it's not natural when impementing - would better hide it in our implementation

Note: well, will be hard without modifying the ISerializingProducer interface...

**Comment by mhowlett** [comment]
Quick thought: We might want to depreciate the key/value(de)serializer properties - they aren't especially useful and are constraining.

**Comment by mhowlett** [comment]
and good point - if we use interfaces/inheritance to provide static type checking on the constructor parameters, we'd probably want to have an additional interface that didn't have any methods (that the two cases 'implement'). Don't like this. 

another quick thought: the API decision we're trying to make here is related to the main API decision that is needed for custom partitioners - they are both configuration. i'm getting to thinking that through as well.

**Comment by mhowlett** [comment]
continuing the above: I'm thinking that we're going to want to allow configuration of partitioners and possibly serializers like so:

```
            var config = new Dictionary<string, object>
            {
                { "group.id", "simple-csharp-consumer" },
                { "bootstrap.servers", brokerList },
                { "partitioner", new MyPartitioner() },
                { "value.deserializer", new StringSerializer(Encoding.UTF8) },
            };
```
sidenote: @edenhill wonders whether the implementation of custom partitioners is going to be easy (noting that other languages make it difficult) - that is a separate question, but at this point I'm guessing it's straightforward and the rdkafka-dotnet implementation is fine.

I've always liked serializers in the constructor argument list. I like that they match up with the type parameters (are not specified separately) and that we get static type checking. The last point is not a good one though, considering we put up with no type checking on other parameters in the config dictionary.

I still like them there. However, it's noteworthy that if they were in the config dictionary, it would be easier to allow for the two different serializer types as there is no static type checking. Maybe the current problem is what will make me change my mind on this. We could depreciate specifying them in the constructor easily enough.

Another option we could go with is to only having one serializer type that knows about topics (as you originally suggest).

If serializers can be specified both as constructor arguments and in the config dictionary there is the question of what happens if both are specified.

Another alternative for the config dictionary, is to have string properties for both the name of classes + filesystem location to load the library from (if not statically referenced). This is consistent with Java in that the class names are specified as a string, but because there isn't really the equivalent of a classpath in C#, I like specifying the actual instances better. 

Another issue is if people are going to want an easy way to get properties specified in a config file (probably yes) - this is not something we've addressed explicitly yet, but something I was thinking about by deciding to use a plain old Dictionary for config. If we allow objects that can't be de-serialized from string easily, this might make that more awkward. 

Actually ... maybe if a string value is specified in a serializer config property its interpreted as the name of a class... if it's not, then the actual class... interesting.

Anyway, I don't know what's best. Whichever way I can think of has tradeoffs. Sorry to be delaying on the decision here... but i think it's worth mulling it over.

btw: I think the decision for custom partitioners is an easier one. This is a little used feature, so it doesn't matter if it's a bit awkward.

mhowlett:
closing in favor of #205

PR #130, opened 2017-03-30
Add  serialializer/deserializer and related unit tests for float and byte[]

wangyanjun:
Add  serialializer/deserializer and related unit tests for float and byte[].  
Same names with java client's types. 
FloatSerializer, FloatDeserializer, ByteArraySerializer, ByteArrayDeserializer

### DIFF
--- a/src/Confluent.Kafka/Serialization/FloatDeserializer.cs
+++ b/src/Confluent.Kafka/Serialization/FloatDeserializer.cs
@@ -44,18 +44,24 @@ namespace Confluent.Kafka.Serialization
                 throw new ArgumentException($"Size of {nameof(data)} received by {nameof(FloatDeserializer)} is not 4");
             }
 
-            float result = default(float);
             // network byte order -> big endian -> most significant byte in the smallest address.
-            unsafe
+            if (BitConverter.IsLittleEndian)
             {
-                byte* p = (byte*)(&result);
-                *p++ = data[3];
-                *p++ = data[2];
-                *p++ = data[1];
-                *p++ = data[0];
+                unsafe
+                {
+                    float result = default(float);
+                    byte* p = (byte*)(&result);
+                    *p++ = data[3];
+                    *p++ = data[2];
+                    *p++ = data[1];
+                    *p++ = data[0];
+                    return result;
+                }
             }
-
-            return result;
+            else
+            {
+                return BitConverter.ToSingle(data);
+            }
         }
     }
 }
--- a/src/Confluent.Kafka/Serialization/FloatDeserializer.cs
+++ b/src/Confluent.Kafka/Serialization/FloatDeserializer.cs
@@ -1,0 +1,61 @@
+﻿// Copyright 2016-2017 Confluent Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Refer to LICENSE for more information.
+
+using System;
+
+namespace Confluent.Kafka.Serialization
+{
+    /// <summary>
+    ///     A deserializer for big endian encoded (network byte ordered) System.Single values.
+    /// </summary>
+    public class FloatDeserializer : IDeserializer<float>
+    {
+        /// <summary>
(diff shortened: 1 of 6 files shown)


mhowlett:
thanks! I think we probably do want a `ByteArraySerializer` (though i need to think about that a bit more - you can also use the non-serializing `Producer` directly, though you might not always want to). I'm not sure that serializing floats will be common enough to warrant including this in the main library - they aren't useful as keys and values will more often than not be complex types. The java library doesn't include float serdes as far as i can tell. I will think about that.

wangyanjun:
Yes. float as key is not useful, as value is also not useful, but more than as key.
Java client added FloatSerializer/FloatDeserializer 17 days ago.
https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/serialization/FloatSerializer.java
Should we keep pace with it?

mhowlett:
The contributors there are Kafka Streams people, so I guess that's what's driving the need here. Ok, I can see that would be useful, currently of the opinion it should be included.

src/Confluent.Kafka/Serialization/FloatDeserializer.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            if (data.Length != 4)
+            {
+                throw new ArgumentException($"Size of {nameof(data)} received by {nameof(FloatDeserializer)} is not 4");
+            }
+
+            float result = default(float);
+            // network byte order -> big endian -> most significant byte in the smallest address.
+            unsafe
+            {
```
treziac: Note that corefx just changed their implementation of BitConverter (don't know when it will be released though), using directly IL (https://github.com/dotnet/corefx/blob/master/src/System.Runtime.CompilerServices.Unsafe/src/System.Runtime.CompilerServices.Unsafe.il and https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/BitConverter.cs#L126)

I wonder if it's better to use BitConverter rather than doing it ourself, especially now that we are cross patform (nothing in C# spec say that int/float it are little endian) The same goes for the other serialization class (we have time to see it coming though, and I previously benchmarked in an other project it was way quicker to use your way)

test/Confluent.Kafka.UnitTests/Serialization/ByteArray.cs
```diff
@@ -0,0 +1,18 @@
...
+    public class ByteArrayTests
+    {
+        [Theory]
+        [InlineData(default(byte[]))]
+        [InlineData(new byte[0])]
+        [InlineData(new byte[] { 1 })]
+        [InlineData(new byte[] { 1, 2, 3, 4, 5 })]
+        public void CanReconstruct(byte[] values)
+        {
+            Assert.Equal(values, new ByteArrayDeserializer().Deserialize(new ByteArraySerializer().Serialize(values)));
```
treziac: add null check (should return null)

wangyanjun: OK.

src/Confluent.Kafka/Serialization/FloatDeserializer.cs
```diff
@@ -0,0 +1,61 @@
...
+        ///     A byte array containing the serialized System.Single value (big endian encoding)
+        /// </param>
+        /// <returns>
+        ///     The deserialized System.Single value.
+        /// </returns>
+        public float Deserialize(byte[] data)
+        {
+            if (data == null)
+            {
+                throw new ArgumentException($"Arg [{nameof(data)}] is null");
```
wangyanjun: Yes. ArgumentNullException is more intuitive.

src/Confluent.Kafka/Serialization/FloatDeserializer.cs
mhowlett: yes, this needs careful checking! I think I would be happiest to see BitConverter used to check whether we are big or little endian, then the deserialization spelled out in unsafe blocks (one for the little endian case, one for big endian) using pointers as you have it. Same for serialization. On the wire, we should always be big endian.

mhowlett:
It would be great if someone could make the case for having `ByteArraySerializer / Deserializer`. I think I remember thinking it would be useful, but I can no longer remember why. Also, we could consider having `ArraySegment<byte>?` instead of byte[] (after all `byte[]` is available on the non-serde variants). I was against using `ArraySegment` on the basic Producer/Consumer but it could make sense here.

treziac:
ByteArraySerializer was useful when we stated that we would keep only the ISerializingProducer and not expose the Producer. I think it's still usefull so that people who want serialize byte array data can do it with the serializing producer (for homogeneity with the rest of their apps)

Concerning ArraySegment, it's also useful to avoid array copy. I don't know what was the advantages comparing to the current implementation with GCHandle and UnsafeAddrOfPinnedArrayElement. But it sure would also make sense for the same reason than byte array (avoid byte copy on the serializing producer)

mhowlett:
continued in #265

PR #134, opened 2017-04-01
Remove events from SerializingProducer

treziac:
Events should come from Producer only. If we create multiple serializedProducer, we don't want to duplicates the log/stats/error (we have only one producer instance and can't know from which Serializingproducer the error came)

Also the constructors made a subscription on Producer.OnLog, thus not logging anything in console when no callbacks were made

Producer<K,V> keeps the events as it create a Producer instance, but events are now just mapped to the Producer instance rather than making an intermediate subscription (same reasons as before)


mhowlett:
This PR is LGTM. The only question I have is whether or not to remove now or depreciate now and remove later. Technically, this is an API change so we should depreciate first, however I think this is a relatively little used feature and expected impact is only minor if someone is using it (not to mention it's fixing a mistake in the API). For that reason, my view is we should just make the change. This question is probably going to come up a bit, so before merging I will consult with some other dev's here to get a better idea of Confluent's general stance on how pedantic we should be in these situations.

mhowlett:
Thanks for picking up on this @treziac - good observation. To answer my open question above, we should generally be very pedantic about marking things as depreciated. In this case, `ISerializingProducer` is only exposed via `Producer`, which is marked as unstable in it's entirety, so I think `Obsolete` attributes are not necessary.

Issue #72, opened 2017-02-23
System.DllNotFoundException: Unable to load DLL 'librdkafka' (Exception from HRESULT: 0x8007007E)


manpreet845:
System.DllNotFoundException: Unable to load DLL 'librdkafka' (Exception from HRESULT: 0x8007007E)

I just wanted to note that running an application with this library, requires Visual C++ Redistributable Packages for Visual Studio 2013.
I couldn't find any Documentation for this requirement, just letting you know since the exception is misleading.

mhowlett:
thanks a lot for pointing that out. I thought we were statically linking these - @edenhill ?

edenhill:
@manpreet845 which Confluent.Kafka and librdkafka.redist nuget package versions are you on?

manpreet845:
@edenhill i was on 0.9.2-preview2 updated to 0.9.4-preview4 and had the same problem.

<package id="Confluent.Kafka" version="0.9.4-preview4" targetFramework="net461" />
<package id="librdkafka.redist" version="0.9.4-RC1-preview4" targetFramework="net461" />
<package id="RdKafka.Internal.librdkafka" version="0.9.2-ci-28" targetFramework="net461" />

mhowlett:
please try referencing nuget package Confluent.Kafka 0.9.4-preview8. we think we've probably solved this, but haven't verified.

manpreet845:
@mhowlett I tried Confluent.Kafka 0.9.4-preview8, but you still need the c++ runtime environment installed

mhowlett:
So you had the problem on 0.9.4-preview2 as well? that means the RdKafka.Internal.librdkafka has the same requirement (that package wasn't created by us). 

Are you on 64 or 32 bit? What OS specifically?

we will try to replicate.

manpreet845:
Sorry for the confusion I did mean the depency RdKafka.Internal.librdkafka, we have machines that are provisioned with puppet and we only install what we need. Our stack is mainly c# so this error wasn't obvious to debug. I have since installed the c++ run time on those machines, so its not an issue for us anymore.
Just wanted to let you know incase some one else  ran into the same issue

mhowlett:
[...]

PR #146, opened 2017-04-14
Fix loading of librdkafka dependencies in .NET Framework 

### LINKED ISSUE #115: Warn if librdkafka version lower than expected
the version of the librdkafka binary that gets loaded is generally enforced via a nuget package dependency, however there are some scenarios where the version loaded could be different than expected (e.g. if the library is swapped out with a different build in the nuget packages directory or if using a custom build that references and older version of librdkafka.redist). 

this has lead to some difficult to track down problems - it'd be great if there was some mechanism to warn the user if an older version of librdkafka is being loaded.

more discussion at the end of #90

### LINKED ISSUE #145: System.TypeInitializationException: The type initializer for 'Confluent.Kafka.Impl.LibRdKafka' threw an exception. ---> System.DllNotFoundException: Unable to load DLL 'librdkafka'
When using an ASP.NET MVC web api project I cannot get the binaries to be in the correct location. 

When publishing it puts them in {PROJECT_HOME}/bin/x64 and x86 as expected but the only way to get it to work is if I move those libraries into {PROJECT_HOME}/bin folder. 

Is this expected?

Obviously a quick fix for this is to move the required binaries during the publish, but I had assumed it would work with bin/x86 and bin/x64 folders that we needed.

See error below.
`System.TypeInitializationException: The type initializer for 'Confluent.Kafka.Impl.LibRdKafka' threw an exception. ---> System.DllNotFoundException: Unable to load DLL 'librdkafka': The specified module could not be found. (Exception from HRESULT: 0x8007007E)`

treziac:
Fix #115 : make 0.9.4 as min version (for producev even if it's not used for now)
We may want to have fallbacks, but for now this will help to not have people targeting Internal.Rdkafka 0.9.1 I think only 0.9.2 is a minimum for now (with flush). Perhaps we could add an attribute on each method in NativeMethods to say what is the minimal supported version and what version are deprecated (for example due to a bug) We could then parse them all when starting given the version loaded and provide some logs on this. This would be an other enhancement though

Fix #145 #72 : now load with LoadLibraryEx with a custom flag so that dependencies are searched in the same folder (this was not the case for msvcr120.dll) Also throw a more meaningful exception

### DIFF
--- a/src/Confluent.Kafka/Impl/LibRdKafka.cs
+++ b/src/Confluent.Kafka/Impl/LibRdKafka.cs
@@ -23,6 +23,7 @@ using System.Runtime.InteropServices;
 using Confluent.Kafka.Internal;
 #if NET45
 using System.Reflection;
+using System.ComponentModel;
 #endif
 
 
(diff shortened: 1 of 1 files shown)


mhowlett:
Because we're referencing librdkafka.redist, I think it's fine to even require the specific version we reference (which will be is 0.9.5 for the imminent release), but 0.9.4 is all we need.

Other changes look good in theory. If they work on all the systems I test on, I'll assume they are a step forward and merge.

thanks!

treziac:
I only tested it on my windows 10 and appveyor (which is also windows server 2016) - but as we don't officialy support mono and it's only for net45, I didn't do much more. The flag should be supported on all windows platform given [msdn documentation](https://msdn.microsoft.com/en-us/library/windows/desktop/ms684179(v=vs.85).aspx).

mhowlett:
Ok, LGTM.

Issue #133, opened 2017-04-01
Correct librdkafka.redist nupkg


treziac:
librdkafka.redist nuget is not correctroy structured
@edenhill  (This is more a librdkafka issue but I think this is more relevant to put it here)

- Content types are incorrect, this cause nuget2 (VS2013) to not unpack the nupkg
- build props is incorrectly loaded for new csproj (it was working for project.json) Took the solution from here : https://github.com/dotnet/cli/issues/4848#issuecomment-265320368
- it misses the package folder we find everywhere else (with metdatata psmdcp file, don't really know where it's used though)

Don't know how to do it with CoApp but here is how to create manually a correct librdkafka.redist to replace existing package

- Unzip https://github.com/treziac/confluent-kafka-dotnet/blob/nuget-librdkafka.redist/librdkafka.redist.zip to get the proper hierarchy and modify version from nuspec
- run "nuget pack librdkafka.nuspec" (tested with nuget 3.5&4 CLI)

The resulted package can be found here:
https://github.com/treziac/confluent-kafka-dotnet/blob/nuget-librdkafka.redist/librdkafka.redist.0.9.4.nupkg

Could test it successfully on VS2013, 2015, 2017 and dotnet cli (windows/linux)

mhowlett:
@treziac - awesome - so happy you've looked into this for us.

treziac:
My bad, it seems there is still a problem with net45 (unit tests don't work on dotnet cli), will investigate a bit more

edenhill:
I played a bit with CoApp, there's a template for it here:
https://github.com/edenhill/librdkafka/blob/master/win32/librdkafka.autopkg.template

But it seems that MS is abandoning CoApp for [vcpkg](https://github.com/Microsoft/vcpkg) which is most likely fantastic but it wont help our users very much since it is too new and not really integrated into anything else, from what I can tell.

edenhill:
I'm planning to roll a new librdkafka.redist packages soon and I would like to include your fixes, @treziac.
What's the latest status, are all known issues resolved with your 0.9.4 patch1?
Can I base my package off that?

treziac:
I did some more tests. 
I found a bug : when using a .net45+ project (standard, not core) with packagereference (new option in vs2017) enable instead of package.config, it does not work.
[...]

PR #147, opened 2017-04-17
Fix librdkafka loading

treziac:
Fix the case where # was present in path (for Uri) (see @cklaw issue in #133 )

Also in the case of @cklaw with the previous PR #146 , it was not possible to load librdkafka manually with the exception thrown.
I still think it's a good thing to throw if we can't load librkdkafka, but also check first if the user has loaded the lib himself (which should not be common, but I had this case and it seems other may have also)

This will also allow people to easily change the version of librdkafka used without having to hack with the content of nupkg (for example to test fix on librdkafka)

It doesn't seem to be feasible in netcoreapp, will have to do more tests on this (calling LoadLibrary on an other librdkafka seems to still use the librdkafka provided in librdkafka.redist)

As I'm not 100% sure GetModuleHandle, LoadLibraryEx or EscapeCadeBase won't throw on some specific platform using net45 (mono, PCL?) I added a try catch with silent catch (this was the case before)

### DIFF
--- a/src/Confluent.Kafka/Impl/LibRdKafka.cs
+++ b/src/Confluent.Kafka/Impl/LibRdKafka.cs
@@ -37,7 +37,7 @@ namespace Confluent.Kafka.Impl
 #if NET45
         [Flags]
         public enum LoadLibraryFlags : uint
-        {     
+        {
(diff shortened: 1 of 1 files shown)


mhowlett:
thanks for this - probably some work to do around this still, but LGTM for now.

Issue #136, opened 2017-04-03
Integration test issues in VS2017


treziac:
I have problems with some integration tests.

**OnPartitionAssignedNotSet**
`Assert.Equal(consumer.Assignment.Count, 1)` fail
I test again a distant kafka. I assume there is not enough time between the Subscription and the poll to get the assignment. Setting a higher timespan (10s) solve this (or making successive poll until the assignment is set, in practice 4 to 6) Also, there is test missing regarding OnPartitionAssigned/Revoked

**OnLog**
OnLog hangs forever sometimes on Loggers.ConsoleLogger. Setting Console.Error to TextWriter.Null resolve the issue, so I assume there is a deadlock somewhere (it occurs when the first consumer destroy is called on librdkafka handle, only if producer were created before. Will try to produce a bunch of reproductible tests.

treziac:
As I'm trying to work on travis for work, I tried to integrate the integration tests. My first idea was to install kafka/zookeeper via the confluent docker images. It actually works well for linux, and I have again this bug on OnPartitionAssignedNotSet

https://travis-ci.org/treziac/confluent-kafka-dotnet/jobs/218813482#L1052
I will have a look later on why it doesn't work on osx.

The branch is here :
https://github.com/treziac/confluent-kafka-dotnet/tree/travis_integration_tests

It won't be easy to do the same for windows as appveyor only support windows container (I don't think confluent plan to release/support docker images for confluent platform on windows container, though it may be possible to run kafka directly) Other solution is just to bind to a distant kafka, though not ideal, or directly download/install/run zookeeper and kafka (but it would be better to test against a kafka installed on linux)

treziac:
For osx, opened an issue on travis-ci here:
https://github.com/travis-ci/travis-ci/issues/7615

Issue #95, opened 2017-03-10
VS 2017 Project system - Project loads fails


enzian:
I have VS2017 installed and we have a base library (netstandard1.5) that references the confluent kafka package. Now when every this project is loaded, it will fail:
![image](https://cloud.githubusercontent.com/assets/5496234/23787679/6e968dac-0573-11e7-9fac-99bebf36259a.png)
The base library also shows a strange behavior - There are ominous x64 and x86 folder in the solution:
![image](https://cloud.githubusercontent.com/assets/5496234/23787764/d18581ca-0573-11e7-8e6e-31d1f4e902e2.png)
As far as I can tell, this is caused by the `contnetFiles` in the NuGet package!

One workaround here is to delete the `obj` folder everytime befor the solution containing the problematic projects is opened - but that's really not a permanent solution!

treziac:
Here is what grpc do on their side to load unmanaged lib with platform support:
https://github.com/grpc/grpc/blob/5098508d2d41a116113f7e333c516cd9ef34a943/src/csharp/Grpc.Core/Internal/UnmanagedLibrary.cs
(basically, do not use DLLImport but GetProcAddress (for windows and equivalent for other) after loading the lib, and include librdkafka_x86.dll and librdkafka_x64.dll in bin rather than in separate folder on windows, so will also need an update on librdkafka.redist for packaging)
Should also make it easier to resolve #91 

Will look at it this we

mhowlett:
I'm no longer seeing this issue when I load the `Confluent.Kafka.sln` file in VS 2017 (I was before). I'm guessing this is probably a result of moving to msbuild - the .csproj files I made by hand differ meaningfully from those auto generated by the migration tool. Also, I don't see the issue when creating a new project that references the 0.9.4 nuget package (in visual studio or via the cli, .net framework or .net core).

I recall seeing some warnings that looked related when using `dotnet pack` on a project that referenced the 0.9.4 nuget package, though I'm struggling to reproduce them now.

I'll leave this open for now, but won't do any more related testing until we are packaging and testing for the next release.

treziac:
I had this solve by modifying my librdkafka.redist nupkg, putting the props in build folder in a new build/net and modify the prop accordingly (it's only a problem on standard net framework) It solves the problem as long as we don't target net framework in apps using the lib

Didn't give a try on the link I gave before

edenhill:
[...]

PR #148, opened 2017-04-17
[WIP] Integration tests in CI

treziac:
See #136 

Add integration tests in travis

This download and install zookeeper / librdkafka from confluent-oss package

As this add almost 8 minutes to the build, it's perhaps not good to run it always. We might also try to cache the zookeeper/kafka directory to accelerate the installation.

An other solution would be to test against some remote cluster which would always be up on confluent side (this would also allow to test in appveyor also) but not an easy go

Concerning appveyor, we could install kafka and zookeeper as well, but it's not officialy supported by confluent platform.

Also set ulimit for osx to correct  dotnet restore on osx

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,7 @@ matrix:
     env: DOTNETCORE=1
 
 before_install:
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then ulimit -n 1024 ; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install --with-openssl curl ; fi
+  - if [ "$TRAVIS_OS_NAME" == "osx" ]; then ulimit -n 1024 ; fi
 
 install:
    - dotnet restore
@@ -29,4 +28,5 @@ install:
 script:
    - make build
    - make unit_test
+   - make install_kafka
    - make integration_test
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: csharp
 env:
   global:
     - CONFIGURATION=Release
+    - INSTALL_KAFKA=True
 
 matrix:
   include:
@@ -18,9 +19,14 @@ matrix:
     mono: none
     env: DOTNETCORE=1
 
+before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then ulimit -n 1024 ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install --with-openssl curl ; fi
+
 install:
    - dotnet restore
 
 script:
    - make build
-   - make test
+   - make unit_test
+   - make integration_test
--- a/Makefile
+++ b/Makefile
@@ -12,11 +12,15 @@ TEST_DIRS==$(shell find ./test -name '*.csproj' -exec dirname {} \;)
 LINUX_FRAMEWORK=netcoreapp1.1
 DEFAULT_FRAMEWORK?=$(LINUX_FRAMEWORK)
 
+CONFLUENT_VERSION=3.2
+CONFLUENT_OSS=confluent-oss-3.2.0-2.11
+CONFLUENT_DIR=confluent-3.2.0
+
 all:
 	@echo "Usage:   make <dotnet-command>"
 	@echo "Example: make build - runs 'dotnet build' for all projects"
 
-.PHONY: test
+.PHONY: unit_test integration_test
 
 build:
 	# Assuming .NET Core on Linux (net451 will not work).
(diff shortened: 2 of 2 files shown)


mhowlett:
cool! one thought: it might be worth running the VerifiableClient as well.

treziac:
Not sure on how to test it though.

Add integration test using the verifiable producer/consumer (and parsing the result, for example failing if we don't have "Stopping: All messages accounted for: 100000 delivered + 0 failed >= 100000" after some given time?)

mhowlett:
I guess I was thinking just some simple test, something like the below (like you say). The point of the verifiable client is to give the clients a good work out as part of our system tests (which are run frequently), but I was thinking there is value in knowing right away if a PR is doing something obviously bad that will make the system tests fail. I haven't thought this through in any depth - I don't mind if not included, just an idea.

```
if [ `dotnet run --producer --broker-list localhost:9092 --topic test.net --acks all --property log.thread.name=true --max-messages 10 --throughput 10 | grep -q 'producer_send_error'` ]
then
    echo "failed"
    exit 1
else
    echo "success"
fi
```

edenhill:
The VerifiableClient is for running the Kafka client kafkatests which take about an hour and require 10 vagrant instances, I dont think we want to do that in CI.

.travis.yml
```diff
@@ -3,6 +3,7 @@ language: csharp
 env:
   global:
     - CONFIGURATION=Release
+    - INSTALL_KAFKA=True
```
mhowlett: is this required?

Makefile
```diff
@@ -9,28 +9,41 @@ OS=$(shell uname -s)
 EXAMPLE_DIRS=$(shell find ./examples -name '*.csproj' -exec dirname {} \;)
 TEST_DIRS==$(shell find ./test -name '*.csproj' -exec dirname {} \;)
 
-LINUX_FRAMEWORK=netcoreapp1.1
-DEFAULT_FRAMEWORK?=$(LINUX_FRAMEWORK)
+#framework can be change to net45 or similar
+FRAMEWORK?=netcoreapp1.1
+
+CONFLUENT_VERSION?=3.2
```
mhowlett: took me so long to get to this you can bump the version to 3.2.1 :-)

mhowlett:
this is looking good to me (few v.minor comments). It's labelled WIP - is there anything else you were thinking of adding?

We can forget about adding anything to do with the VerifiableClient in this PR, though I still think it's a good idea to quickly verify that it at least runs here.

Makefile
treziac: Yes, i couldn't figure a proper way with oss-version and the way the url is generated... Perhaps add some comments

.travis.yml
treziac: Not anymore, forgot to remove it

treziac:
WIP as I wanted to add appveyor (as the package is generated comes from appveyor, it's this one we want to test, especially under windows...)
Do you guys think it can be tested against a remote cluster on your side?

mhowlett:
I think it's fine for this PR to only make changes for travis - it's a good step forward.

Right now I can't think of a good way of setting up / getting access to a kafka cluster from AppVeyor...

treziac:
Is there any reason the bin/windows script are not working on windows? (w10 x64) I couldn't find where the files came from, they are not the same in confluentinc/kafka github repo. The one from the official kafka package works correctly though. Would happily correct this if I could find them. Goal is to run kafka on windows for integration test on appveyor (using the confluent-oss rather than apache tar gz)
@ewencp
I would really feel more confident making integration tests works on appveyor too :)

treziac:
Oops, forgot my tests would mess up this PR, will continue them on an other branch

mhowlett:
I'm still in favour of incremental PR to get integration tests working on travis, then tackle windows in a later PR. quick wins :-)

treziac:
I think it should be ok like this then. There is a dummy conflict which I cannot resolve on my side, but it should be ok here. Do you want me to open a new PR with rebase or do you do it yourself?

treziac:
Should be ok like this - will add VerifiableClient in an other PR, renamed to make clear it is only for integration test on travis. 

Any news about the windows script invalid on confluent package? Would be nice to correct this for next confluent release, would happily do this if I have proper indications

mhowlett:
Our packaging scripts are patching the .bat files (to accommodate config -> etc directory name change). I've filed an issue that they aren't working and pointed here. Note that we don't officially support Kafka on Windows.

treziac:
I couldn't find where the bat comes from (it's not the one from kafka, it's not the one on confluentinc/kafka repo either). As they don't work as-si (they aren't even runnable bacause of misspel or not out-of-the-box command, specifically line 117 in kafka-run-call.bat from confluent-oss-3.2.1-2.11 package and confuent-3.2.1-2.11 is

```bash
rem Log4j settings
IF ["%KAFKA_LOG4J_OPTS%"] EQU [""] (
	if exists %~dp0../../etc/kafka/tools-log4j.properties (
		set KAFKA_LOG4J_OPTS=-Dlog4j.configuration=file:%~dp0../../etc/kafka/tools-log4j.properties
	) else (
		set KAFKA_LOG4J_OPTS=-Dlog4j.configuration=file:%BASE_DIR%/config/tools-log4j.properties
	)
) ELSE (
  rem create logs directory
  IF not exist %LOG_DIR% (
      mkdir %LOG_DIR%
  )
)
```
Making exist instead of exists solve the issue


where in the corresponding repo https://github.com/confluentinc/kafka/blob/v3.2.1/bin/windows/kafka-run-class.bat#L117 we find
```shell
rem Log4j settings
IF ["%KAFKA_LOG4J_OPTS%"] EQU [""] (
	set KAFKA_LOG4J_OPTS=-Dlog4j.configuration=file:%BASE_DIR%/config/tools-log4j.properties
) ELSE (
  rem create logs directory
  IF not exist %LOG_DIR% (
      mkdir %LOG_DIR%
  )
)
```

So I have no clue where the bat comes from to check it or do a PR

treziac:
There are still some problems but I would rather see them when happening to better judge them:
- Kafka may not be started after ten seconds. Maybe looking at the logs file with a 30s timeout
- Some tests are still failing sometimes: https://travis-ci.org/treziac/confluent-kafka-dotnet/jobs/237930662 (see #95 ), would rather correct them when integration run both on appveyor and travis. (with the verifiable client too)

mhowlett:
thanks for identifying the problem! I'll try to get that in today.

packaging is a private repo.

treziac:
I didn't tried to run the batch - there was other errors and quickly switched to kafka package on windows. Would happily correct and add appveyor integration tests on the repo if ewencp or other are ok to give me read access ;)

treziac:
Also, didn't checked why but travis fail to build strongname, and strangely does not report broken build. Will investigate. It's working fine on linux (it has no real impact as we only update the package built on appveyor to have the net45 framework)

PR #149, opened 2017-04-18
 API docs completed

mhowlett:
@edenhill 
feel free to be pedantic.


src/Confluent.Kafka/LogMessage.cs
```diff
@@ -24,17 +24,48 @@ namespace Confluent.Kafka
     /// </summary>
     public class LogMessage
     {
+        /// <summary>
+        ///     Instantiates a new LogMessage class instance.
+        /// </summary> 
+        /// <param name="name">
+        ///     The librdkakfa client name.
```
edenhill: "client instance name"

src/Confluent.Kafka/LogMessage.cs
```diff
@@ -24,17 +24,48 @@ namespace Confluent.Kafka
...
+        /// </param>
+        /// <param name="level">
+        ///     The log level (levels correspond to syslog(3)).
```
edenhill: ", lower is worse" or a variation thereof

src/Confluent.Kafka/LogMessage.cs
```diff
@@ -24,17 +24,48 @@ namespace Confluent.Kafka
...
         {
+            // TODO: Define an enumeration for log level.
             Name = name;
             Level = level;
             Facility = facility;
             Message = message;
         }
 
+        /// <summary>
+        ///     Gets the librdkafka client name.
```
edenhill: "client instance name"

src/Confluent.Kafka/LogMessage.cs
```diff
@@ -24,17 +24,48 @@ namespace Confluent.Kafka
...
+        /// </summary>
         public string Name { get; }
+
+        /// <summary>
+        ///     Gets the log level (levels correspond to syslog(3)).
```
edenhill: dito

src/Confluent.Kafka/PartitionMetadata.cs
```diff
@@ -27,6 +27,24 @@ namespace Confluent.Kafka
     /// </summary>
     public class PartitionMetadata
     {
+        /// <summary>
```
edenhill: Should constructors for these sort of types really be exposed to the users?

src/Confluent.Kafka/PartitionMetadata.cs
```diff
@@ -27,6 +27,24 @@ namespace Confluent.Kafka
...
+        ///     The id of the partition this metadata relates to.
+        /// </param>
+        /// <param name="leader">
+        ///     The id of the broker that is the leader for the partition.
+        /// </param>
+        /// <param name="replicas">
+        ///     The ids of all brokers that contain replicas of the partition.
+        /// </param>
+        /// <param name="inSyncReplicas">
+        ///     The ids of all brokers that contain in-sync replicas of the partition.
```
edenhill: Make a note that this value is cached by the broker and can thus be outdated and not be trusted.

src/Confluent.Kafka/Producer.cs
```diff
@@ -124,10 +124,10 @@ private SafeTopicHandle getKafkaTopicHandle(string topic)
 
         private static readonly LibRdKafka.DeliveryReportDelegate DeliveryReportCallback = DeliveryReportCallbackImpl;
 
-        /// <param name="opaque">
+        /// <remarks>
         ///     note: this property is set to that defined in rd_kafka_conf
```
edenhill: is note: needed when used with the remarks tag?

src/Confluent.Kafka/Serialization/NullDeserializer.cs
```diff
@@ -16,8 +16,20 @@
 
 namespace Confluent.Kafka.Serialization
 {
+    /// <summary>
+    ///     A 'dummy' deserializer for use with values that must be null.
+    /// </summary>
     public class NullDeserializer : IDeserializer<Null>
     {
+        /// <summary>
+        ///     'deserializes' any byte array to a null value.
```
edenhill: capital d

src/Confluent.Kafka/Serialization/NullDeserializer.cs
```diff
@@ -16,8 +16,20 @@
 
 namespace Confluent.Kafka.Serialization
 {
+    /// <summary>
+    ///     A 'dummy' deserializer for use with values that must be null.
```
edenhill: dont quote dummy, it really is a dummy

edenhill: does "that must be null" imply that it will fail if not null? Clarify in doc

src/Confluent.Kafka/Serialization/NullDeserializer.cs
```diff
@@ -16,8 +16,20 @@
...
+    /// </summary>
     public class NullDeserializer : IDeserializer<Null>
     {
+        /// <summary>
+        ///     'deserializes' any byte array to a null value.
+        /// </summary>
+        /// <param name="data">
+        ///     The data to deserialize (should be null)
```
edenhill: should, must?

src/Confluent.Kafka/Serialization/NullSerializer.cs
```diff
@@ -16,8 +16,17 @@
 
 namespace Confluent.Kafka.Serialization
 {
+    /// <summary>
+    ///     A 'dummy' serializer for use with values that must be null.
```
edenhill: dito

src/Confluent.Kafka/Serialization/StringDeserializer.cs
```diff
@@ -18,20 +18,32 @@
 
 namespace Confluent.Kafka.Serialization
 {
+    /// <summary>
+    ///     A deserializer for string values.
```
edenhill: mention encoding. UTF8?

src/Confluent.Kafka/Serialization/StringSerializer.cs
```diff
@@ -18,15 +18,27 @@
 
 namespace Confluent.Kafka.Serialization
 {
+    /// <summary>
+    ///     A serializer for string values.
```
edenhill: dito encoding

src/Confluent.Kafka/TopicMetadata.cs
```diff
@@ -28,17 +28,46 @@ namespace Confluent.Kafka
     /// </summary>
     public class TopicMetadata
     {
+        /// <summary>
```
edenhill: dito: dont expose constructor?

src/Confluent.Kafka/TopicPartition.cs
```diff
@@ -59,16 +68,52 @@ public override bool Equals(object obj)
...
+        /// </summary>
+        /// <returns>
+        ///     An integer that specifies a hash value for this TopicPartition.
+        /// </returns>
         public override int GetHashCode()
+            // x by prime number is quick and gives decent distribution.
             => Partition.GetHashCode()*251 + Topic.GetHashCode();
 
+        /// <summary>
+        ///     Tests whether TopicPartition instance a is equal to Offset instance b.
```
edenhill: -Offset +TopicPartition

src/Confluent.Kafka/TopicPartition.cs
```diff
@@ -59,16 +68,52 @@ public override bool Equals(object obj)
...
+        ///     The second TopicPartition instance to compare.
+        /// </param>
+        /// <returns>
+        ///     true if TopicPartition instances a and b are equal. false otherwise.
+        /// </returns>
         public static bool operator ==(TopicPartition a, TopicPartition b)
             => a.Equals(b);
 
+        /// <summary>
+        ///     Tests whether TopicPartition instance a is not equal to Offset instance b.
```
edenhill: -offset +..dito

src/Confluent.Kafka/TopicPartitionOffset.cs
```diff
@@ -86,16 +95,52 @@ public override bool Equals(object obj)
...
+        /// </summary>
+        /// <returns>
+        ///     An integer that specifies a hash value for this TopicPartitionOffset.
+        /// </returns>
+        public override int GetHashCode()  
+            // x by prime number is quick and gives decent distribution.
             => (Partition.GetHashCode()*251 + Topic.GetHashCode())*251 + Offset.GetHashCode();
 
+        /// <summary>
+        ///     Tests whether TopicPartitionOffset instance a is equal to Offset instance b.
```
edenhill: -offset dito

src/Confluent.Kafka/TopicPartitionOffset.cs
```diff
@@ -86,16 +95,52 @@ public override bool Equals(object obj)
...
+        ///     The second TopicPartitionOffset instance to compare.
+        /// </param>
+        /// <returns>
+        ///     true if TopicPartitionOffset instances a and b are equal. false otherwise.
+        /// </returns>
         public static bool operator ==(TopicPartitionOffset a, TopicPartitionOffset b)
             => a.Equals(b);
 
+        /// <summary>
+        ///     Tests whether TopicPartitionOffset instance a is not equal to Offset instance b.
```
edenhill: dito

src/Confluent.Kafka/TopicPartitionOffsetError.cs
```diff
@@ -116,15 +125,52 @@ public override bool Equals(object obj)
...
+        /// <summary>
+        ///     Returns a hash code for this TopicPartitionOffsetError.
+        /// </summary>
+        /// <returns>
+        ///     An integer that specifies a hash value for this TopicPartitionOffsetError.
+        /// </returns>
         public override int GetHashCode()
-            => ((Partition.GetHashCode()*251 + Topic.GetHashCode())*251 + Offset.GetHashCode())*251 + Error.GetHashCode(); // x by prime number is quick and gives decent distribution.
+            // x by prime number is quick and gives decent distribution.
+            => ((Partition.GetHashCode()*251 + Topic.GetHashCode())*251 + Offset.GetHashCode())*251 + Error.GetHashCode();
```
edenhill: is all this multiplication giving proper distribution?

src/Confluent.Kafka/TopicPartitionOffsetError.cs
```diff
@@ -116,15 +125,52 @@ public override bool Equals(object obj)
...
+        ///     The second TopicPartitionOffsetError instance to compare.
+        /// </param>
+        /// <returns>
+        ///     true if TopicPartitionOffsetError instances a and b are equal. false otherwise.
+        /// </returns>
         public static bool operator ==(TopicPartitionOffsetError a, TopicPartitionOffsetError b)
             => a.Equals(b);
 
+        /// <summary>
+        ///     Tests whether TopicPartitionOffsetError instance a is not equal to Offset instance b.
```
edenhill: dito

src/Confluent.Kafka/LogMessage.cs
mhowlett: good idea. we should have made an enum for these, bit late now.

src/Confluent.Kafka/PartitionMetadata.cs
mhowlett: Minor improvement, but good point - it would be better if they weren't. I'll make an issue for this (should be a separate PR). It's an API change, but we can probably get away with just making the changes as the metadata methods are marked as unstable (even if these classes aren't).

src/Confluent.Kafka/Producer.cs
mhowlett: no. good point, i do this in more than one place (changed a few). also i forgot to reference the name 'opaque' having changed from param to remarks. the reason for changing to remarks is a warning is generated if you document one param and not the others (even on a private method).

src/Confluent.Kafka/Serialization/NullDeserializer.cs
mhowlett: it's clarified on the `Serialize` method. I copied that to the class summary.

src/Confluent.Kafka/Serialization/NullDeserializer.cs
mhowlett: should was more correct because there was previously no check that the input data was `null`. However I just added this check - I think it's a valuable check that will catch real world errors. This is technically an API change (change in semantics), but I propose just making it.

src/Confluent.Kafka/Serialization/StringDeserializer.cs
mhowlett: there is no specific encoding, it is specified in the constructor (documented)

src/Confluent.Kafka/TopicPartitionOffsetError.cs
mhowlett: I don't think it matters much if the hash function is poor for this class, only that it is consistent with `Equals`. Anyway, I took the implementation from somewhere that claimed it was good (though it's possible I misinterpreted. I can no longer find the reference).

src/Confluent.Kafka/LogMessage.cs
```diff
@@ -28,10 +28,10 @@ public class LogMessage
         ///     Instantiates a new LogMessage class instance.
         /// </summary> 
         /// <param name="name">
-        ///     The librdkakfa client name.
+        ///     The librdkakfa client instance name.
         /// </param>
         /// <param name="level">
-        ///     The log level (levels correspond to syslog(3)).
+        ///     The log level (levels correspond to syslog(3)), lower is worse.
```
edenhill: sorry, but "severe" is probably the correct word :)
(feel free to ignore)

src/Confluent.Kafka/Serialization/NullDeserializer.cs
```diff
@@ -17,21 +17,26 @@
...
+        ///     The data to deserialize (must be null).
         /// </param>
         /// <returns>
         ///     null
         /// </returns>
         public Null Deserialize(byte[] data)
         {
+            if (data != null)
+            {
+                throw new System.ArgumentException("NullDeserializer may only be used to deserialize data that is null.");
```
edenhill: I like it, but we might want to double check this behaviour with the Java counterpart?

src/Confluent.Kafka/TopicPartitionOffsetError.cs
edenhill: I don't know math so I can't argue either way :)

PR #163, opened 2017-04-27
Remove redundant awaits

andreycha:
Removed redundant `await`'s and `async` modifiers:
- There is no need to await call results, if they are not used in the continuation. Methods will be awaited by the caller. Also removing `async` modifier allows to gain some additional performance, since the compiler now won't generate state machines for these methods.
- Library code anyway shouldn't capture client's synchronization context ([details](https://channel9.msdn.com/Series/Three-Essential-Tips-for-Async/Async-library-methods-should-consider-using-Task-ConfigureAwait-false-)).

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -1041,8 +1041,8 @@ namespace Confluent.Kafka
         /// <summary>
         ///     Commit offsets for the current assignment.
         /// </summary>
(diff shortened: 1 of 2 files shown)


mhowlett:
LGTM

mhowlett:
thanks @andreycha !

andreycha:
@mhowlett you're welcome!

Issue #175, opened 2017-05-11
When a deserializer throws an exception, expose via OnConsumeError


mhowlett:
Exceptions are not used very much in the consumer API, so i'm thinking it would be a more consistent (and more convenient) if any exception thrown by a deserializer was surfaced via the `OnConsumeError` event.

It would be best if there was a new `ErrorCode` value for this. It would be good if librdkafka explicitly set aside a block of values reserved for custom use.

also, when a serializer throws an exception, producer probably should handle.

treziac:
Agreed on this

mhowlett:
closing - PR #179 addresses this.

PR #179, opened 2017-05-14
Expose deserialization errors via OnConsumeError

mhowlett:
see #175 for discussion
@treziac @edenhill

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -89,18 +89,8 @@ namespace Confluent.Kafka
 
             consumer = new Consumer(config);
 
-            // TODO: bypass this.consumer for this event to optimize perf.
-            consumer.OnMessage += (sender, e) => OnMessage?.Invoke(sender,
-                new Message<TKey, TValue> (
-                    e.Topic,
-                    e.Partition,
-                    e.Offset,
-                    KeyDeserializer.Deserialize(e.Key),
-                    ValueDeserializer.Deserialize(e.Value),
-                    e.Timestamp,
-                    e.Error
-                )
-            );
+            consumer.OnConsumeError += (sender, msg) 
+                => OnConsumeError?.Invoke(this, msg);
         }
 
         /// <summary>
@@ -139,7 +129,26 @@ namespace Confluent.Kafka
                 return false;
             }
 
-            message = msg.Deserialize(KeyDeserializer, ValueDeserializer);
+            try
+            {
+                message = msg.Deserialize(KeyDeserializer, ValueDeserializer);
+            }
+            catch (KafkaException ex)
+            {
+                var erroredMsg = new Message(
+                    msg.Topic,
+                    msg.Partition,
+                    msg.Offset,
+                    msg.Key,
+                    msg.Value,
+                    msg.Timestamp,
+                    ex.Error
+                );
+                OnConsumeError(this, erroredMsg);
+                message = null;
+                return false;
+            }
+
             return true;
         }
 
@@ -284,11 +293,7 @@ namespace Confluent.Kafka
         /// <remarks>
         ///     Executes on the same thread as every other Consumer event handler (except OnLog which may be called from an arbitrary thread).
         /// </remarks>
-        public event EventHandler<Message> OnConsumeError
-        {
-            add { consumer.OnConsumeError += value; }
-            remove { consumer.OnConsumeError -= value; }
-        }
+        public event EventHandler<Message> OnConsumeError;
 
         /// <summary>
         ///     Raised on critical errors, e.g. connection failures or all 
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -144,7 +144,7 @@ namespace Confluent.Kafka
                     msg.Timestamp,
                     ex.Error
                 );
-                OnConsumeError(this, erroredMsg);
+                OnConsumeError?.Invoke(this, erroredMsg);
                 message = null;
                 return false;
             }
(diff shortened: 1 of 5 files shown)


src/Confluent.Kafka/ErrorCode.cs
```diff
@@ -29,6 +29,26 @@ namespace Confluent.Kafka
     public enum ErrorCode
     {
         /// <summary>
+        ///     Value serialization error
+        /// </summary>
+        Local_ValueSerialization = -1004,
```
mhowlett: @edenhill - what do you think about the idea of reserving a block of librdkafka error return codes for arbitrary client use? alternatively these particular values could be defined by librdkafka even though it never produces them (it's reasonable other clients might also need them). 

the first idea has the disadvantage that different clients may define different values for the same error, which could potentially cause confusion. For that reason I'd prefer it if librdkafka defined them individually.

Maybe these error codes could be prefixed with `Client_` rather than `Local_` here (or something else). Maybe librdkafka could distinguish this type of error code (one that it never produces) using triple underscore.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -89,18 +89,8 @@ public class Consumer<TKey, TValue> : IDisposable
 
             consumer = new Consumer(config);
 
-            // TODO: bypass this.consumer for this event to optimize perf.
```
mhowlett: removed this because it is never called because the `Poll` method on the non-deserializing consumer is never called. It's hanging around from earlier times.

src/Confluent.Kafka/KafkaException.cs
```diff
@@ -40,6 +40,22 @@ public KafkaException(Error error)
...
+        ///     Initialize a new instance of KafkaException based on
+        ///     an existing Error instance and inner exception.
+        /// </summary>
+        /// <param name="error">
+        ///     The Kafka Error.
+        /// </param>
+        /// <param name="innerException">
+        ///     The exception instance that caused this exception.
+        /// </param>
+        public KafkaException(Error error, Exception innerException)
```
mhowlett: The InnerException is never used (because the exception is eventually squashed down into an `Error`), but I thought we might as well propagate it whilst it's easy to do so. In practice, not propagating this all the way to the `OnConsumeError` event is not a problem I think, as a custom Message string will be enough in practice in this scenario.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -139,7 +129,26 @@ public bool Consume(out Message<TKey, TValue> message, int millisecondsTimeout)
...
+                var erroredMsg = new Message(
+                    msg.Topic,
+                    msg.Partition,
+                    msg.Offset,
+                    msg.Key,
+                    msg.Value,
+                    msg.Timestamp,
+                    ex.Error
+                );
+                OnConsumeError(this, erroredMsg);
```
mhowlett: good catch. it was this originally, i changed it when experimenting and forgot to change it back..

src/Confluent.Kafka/Serialization/Extensions/Message.cs
```diff
@@ -38,14 +40,37 @@ public static class MessageExtensions
...
         /// </returns>
         public static Message<TKey, TValue> Deserialize<TKey, TValue>(this Message message, IDeserializer<TKey> keyDeserializer, IDeserializer<TValue> valueDeserializer)
-            => new Message<TKey, TValue> (
+        {
+            TKey key;
+            TValue val;
+
+            try
+            {
+                key = keyDeserializer.Deserialize(message.Key);
```
mhowlett: Yep but keeping this PR to the consumer which is most important as people may hit this by running the examples.

src/Confluent.Kafka/Serialization/Extensions/Message.cs
```diff
@@ -38,14 +40,37 @@ public static class MessageExtensions
...
+                throw new KafkaException(new Error(ErrorCode.Local_KeyDeserialization, ex.Message), ex);
+            }
+
+            try
+            {
+                val = valueDeserializer.Deserialize(message.Value);
+            }
+            catch (Exception ex)
+            {
+                throw new KafkaException(new Error(ErrorCode.Local_KeyDeserialization, ex.Message), ex);
```
mhowlett: whoops, yep.

src/Confluent.Kafka/Serialization/Extensions/Message.cs
treziac: Perhaps remove the related serialization errorCode from this PR then

mhowlett: makes sense

src/Confluent.Kafka/ErrorCode.cs
mhowlett: The extra error codes are specific to librdkafka in that they won't be used by anything other than a library that wraps librdkafka - I can't think of any better place for them to be defined outside of confluent-kafka-dotnet. 

I don't mind too much if they are defined only in confluent-kafka-dotnet (and repeated in other libraries if necessary).

I also like `Client_` better.

test/Confluent.Kafka.IntegrationTests/Tests/DeserializingConsumer_Poll_Error.cs
```diff
@@ -0,0 +1,145 @@
...
+                    errCnt += 1;
+                    Assert.Equal(ErrorCode.Local_KeyDeserialization, msg.Error.Code);
+                };
+
+                consumer.OnPartitionEOF += (_, partition)
+                    => done = true;
+
+                consumer.OnPartitionsAssigned += (_, partitions) =>
+                {
+                    Assert.Equal(1, partitions.Count);
```
mhowlett: yeah, we could have a better setup for integration tests and what you point out is probably worth doing as it's probably a common gotcha.

Issue #92, opened 2017-03-09
producer.Flush can hang using a simple example


LucVL:
While experimenting with the SimpleProducer/-Consumer examples, I found that I can systematically make the producer.Flush() call hang without doing anything obviously wrong.

In my GitHub repo asyncAwait branch [LucVL/rdkafka-dotnet](https://github.com/LucVL/rdkafka-dotnet/tree/asyncAwait) you can see the steps I took to isolate and reproduce the problem, and then workaround it in an unexpected way.

What I basically did was:
- Change the example to produce a single message and return.
- Put the `producer.ProduceAsync` call chained with a `ContinueWith` in a separate method returning the Task created by `ContinueWith` (`ProduceAndReportUsingContinueWithAsync`).
- await the returned task instead of relying on `producer.Flush()` to do the awaiting on tasks in progress.

So far everything continues to work as expected.

Then:
- Create a new method that does exactly the same thing, but using the async/await syntax (`ProduceAndReportUsingAwaitAsync`).
- Call that method instead of `ProduceAndReportUsingContinueWithAsync`
=> example hangs on the producer.Flush() call.

If the producer.Flush call is put in comments, the program hangs on leaving the using block instead.

The async/await syntax is much more readable and should in principle be equivalent to the `ContinueWith` style, except that the compiler can short-circuit the `ContinueWith` if the first task immediately returns in the `RanToCompletion` state.

I suspect that that is the case here since a simple workaround is putting `await Task.Yield()` after the `await producer.ProduceAsync(...)` call.

Putting the `ProduceAndReportUsingContinueWithAsync` code directly in the using block produces the same effect (and requires the same workaround).

mhowlett:
Your SimpleProducer variant is blocking in the Producer Dispose method. I haven't had a chance to take a proper look, but I'm currently guessing the answer is likely to be found in the below blog post and that we're doing things incorrectly in the library - thanks for pointing this out. 

http://blog.stephencleary.com/2012/07/dont-block-on-async-code.html

mhowlett:
[...]

PR #180, opened 2017-05-16
[WIP] Call SetResult on threadpool for Task ProduceAsync variants

mhowlett:
This is a quick fix for #92.

The performance hit is unfortunately very high - scheduling on the thread pool results in a 4x slow down compared to the `IDeliveryHandler` variants of `ProduceAsync` (as @ah- warned).

I'm wondering if there is a way of solving this on the poll loop side of things - some fancy dual-thread arrangement or something (but i'm a bit wary - it's less straightforward to analyse).

finally: the tests actually work even without the fix in place for some reason. I haven't tried to work out why. I've verified the fix is effective in a console app.

Considering this a Work-In-Progress - putting it out there as a start.

### DIFF
--- a/src/Confluent.Kafka/Confluent.Kafka.csproj
+++ b/src/Confluent.Kafka/Confluent.Kafka.csproj
@@ -13,7 +13,7 @@
     <Title>Confluent.Kafka</Title>
     <AssemblyName>Confluent.Kafka</AssemblyName>
     <VersionPrefix>0.9.5</VersionPrefix>
-    <TargetFrameworks>net45;netstandard1.3</TargetFrameworks>
+    <TargetFrameworks>net45;net46;netstandard1.3</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
--- a/src/Confluent.Kafka/Impl/LibRdKafka.cs
+++ b/src/Confluent.Kafka/Impl/LibRdKafka.cs
@@ -21,7 +21,7 @@ using System.IO;
 using System.Text;
 using System.Runtime.InteropServices;
 using Confluent.Kafka.Internal;
-#if NET45
+#if NET45 || NET46
 using System.Reflection;
 using System.ComponentModel;
 #endif
@@ -34,7 +34,7 @@ namespace Confluent.Kafka.Impl
         //min librdkafka version, to change when binding to new function are added
         const long minVersion = 0x000903ff;
 
-#if NET45
+#if NET45 || NET46
         [Flags]
         public enum LoadLibraryFlags : uint
         {
@@ -84,7 +84,7 @@ namespace Confluent.Kafka.Impl
 
         static LibRdKafka()
         {
-#if NET45
+#if NET45 || NET46
             try
             {
                 // In net45, we don't put the native dlls in the assembly directory
--- a/src/Confluent.Kafka/Producer.cs
+++ b/src/Confluent.Kafka/Producer.cs
@@ -182,11 +182,19 @@ namespace Confluent.Kafka
 
         private sealed class TaskDeliveryHandler : TaskCompletionSource<Message>, IDeliveryHandler
         {
+#if !NET45
+            public TaskDeliveryHandler() : base(TaskContinuationOptions.RunContinuationsAsynchronously)
+            { }
+#endif
             public bool MarshalData { get { return true; } }
 
             public void HandleDeliveryReport(Message message)
             {
+#if NET45
                 System.Threading.Tasks.Task.Run(() => SetResult(message));
+#else
+                SetResult(message);
+#endif
             }
         }
 
@@ -825,6 +833,9 @@ namespace Confluent.Kafka
         private class TypedTaskDeliveryHandlerShim : TaskCompletionSource<Message<TKey, TValue>>, IDeliveryHandler
         {
             public TypedTaskDeliveryHandlerShim(TKey key, TValue val)
+#if !NET45
+                : base(TaskContinuationOptions.RunContinuationsAsynchronously)
+#endif
             {
                 Key = key;
                 Value = val;
@@ -848,7 +859,11 @@ namespace Confluent.Kafka
                     message.Error
                 );
 
+#if NET45
                 System.Threading.Tasks.Task.Run(() => SetResult(mi));
+#else
+                SetResult(mi);
+#endif
             }
         }
 
--- a/src/Confluent.Kafka/Producer.cs
+++ b/src/Confluent.Kafka/Producer.cs
@@ -186,7 +186,7 @@ namespace Confluent.Kafka
 
             public void HandleDeliveryReport(Message message)
             {
-                SetResult(message);
+                System.Threading.Tasks.Task.Run(() => SetResult(message));
             }
         }
 
@@ -848,7 +848,7 @@ namespace Confluent.Kafka
                     message.Error
                 );
 
-                SetResult(mi);
+                System.Threading.Tasks.Task.Run(() => SetResult(mi));
             }
         }
 
--- a/test/Confluent.Kafka.IntegrationTests/Tests/DeserializingConsumer_Poll_Error.cs
+++ b/test/Confluent.Kafka.IntegrationTests/Tests/DeserializingConsumer_Poll_Error.cs
@@ -44,7 +44,7 @@ namespace Confluent.Kafka.IntegrationTests
             {
                 firstProduced = producer.ProduceAsync(topic, Encoding.UTF8.GetBytes("key"), null).Result.TopicPartitionOffset;
                 producer.ProduceAsync(topic, null, Encoding.UTF8.GetBytes("val"));
-                producer.Flush();
+                producer.Flush(30);
             }
 
             var consumerConfig = new Dictionary<string, object>
--- a/test/Confluent.Kafka.IntegrationTests/Tests/Producer_ProduceAsync_Await.cs
+++ b/test/Confluent.Kafka.IntegrationTests/Tests/Producer_ProduceAsync_Await.cs
@@ -1,0 +1,53 @@
+// Copyright 2016-2017 Confluent Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Refer to LICENSE for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Xunit;
+
+
+namespace Confluent.Kafka.IntegrationTests
+{
+    /// <summary>
+    ///     Ensures that awaiting ProduceAsync does not deadlock.
+    /// </summary>
+    public static partial class Tests
+    {
+        public static async Task Producer_ProduceAsync_Await_Task(Dictionary<string, object> config, string topic)
+        {
+            using (var producer = new Producer(config))
+            {
+                var dr = await producer.ProduceAsync(topic, new byte[] {42}, new byte[] {44});
+                Assert.Equal(ErrorCode.NoError, dr.Error.Code);
+                producer.Flush();
+            }
+        }
+
+        [Theory, MemberData(nameof(KafkaParameters))]
+        public static void Producer_ProduceAsync_Await(string bootstrapServers, string topic, string partitionedTopic)
+        {
+            var producerConfig = new Dictionary<string, object> 
+            { 
+                { "bootstrap.servers", bootstrapServers },
+                { "api.version.request", true }
+            };
+
+            var task = Producer_ProduceAsync_Await_Task(producerConfig, topic);
+            task.Wait();
+        }
+    }
+}
--- a/test/Confluent.Kafka.IntegrationTests/Tests/Producer_ProduceAsync_Await.cs
+++ b/test/Confluent.Kafka.IntegrationTests/Tests/Producer_ProduceAsync_Await.cs
@@ -33,7 +33,7 @@ namespace Confluent.Kafka.IntegrationTests
             {
                 var dr = await producer.ProduceAsync(topic, new byte[] {42}, new byte[] {44});
                 Assert.Equal(ErrorCode.NoError, dr.Error.Code);
-                producer.Flush();
+                producer.Flush(30);
             }
         }
 
--- a/test/Confluent.Kafka.IntegrationTests/Tests/SerializingProducer_ProduceAsync_Await.cs
+++ b/test/Confluent.Kafka.IntegrationTests/Tests/SerializingProducer_ProduceAsync_Await.cs
@@ -1,0 +1,55 @@
+// Copyright 2016-2017 Confluent Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Refer to LICENSE for more information.
+
+using System;
+using System.Text;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using Confluent.Kafka.Serialization;
+using Xunit;
+
+
+namespace Confluent.Kafka.IntegrationTests
+{
+    /// <summary>
+    ///     Ensures that awaiting ProduceAsync does not deadlock.
+    /// </summary>
+    public static partial class Tests
+    {
+        public static async Task SerializingProducer_ProduceAsync_Await_Task(Dictionary<string, object> config, string topic)
+        {
+            using (var producer = new Producer<Null, string>(config, null, new StringSerializer(Encoding.UTF8)))
+            {
+                var dr = await producer.ProduceAsync(topic, null, "test string");
+                Assert.Equal(ErrorCode.NoError, dr.Error.Code);
+                producer.Flush();
+            }
+        }
+
+        [Theory, MemberData(nameof(KafkaParameters))]
+        public static void SerializingProducer_ProduceAsync_Await(string bootstrapServers, string topic, string partitionedTopic)
+        {
+            var producerConfig = new Dictionary<string, object> 
+            { 
+                { "bootstrap.servers", bootstrapServers },
+                { "api.version.request", true }
+            };
+
+            var task = SerializingProducer_ProduceAsync_Await_Task(producerConfig, topic);
+            task.Wait();
+        }
+    }
+}
--- a/test/Confluent.Kafka.IntegrationTests/Tests/SerializingProducer_ProduceAsync_Await.cs
+++ b/test/Confluent.Kafka.IntegrationTests/Tests/SerializingProducer_ProduceAsync_Await.cs
@@ -35,7 +35,7 @@ namespace Confluent.Kafka.IntegrationTests
             {
                 var dr = await producer.ProduceAsync(topic, null, "test string");
                 Assert.Equal(ErrorCode.NoError, dr.Error.Code);
-                producer.Flush();
+                producer.Flush(30);
             }
         }
 


treziac:
I didn't benchmark yet, but this also work
```c#
#if NET45
            public TypedTaskDeliveryHandlerShim(TKey key, TValue val)
#else
            public TypedTaskDeliveryHandlerShim(TKey key, TValue val) : base(TaskCreationOptions.RunContinuationsAsynchronously)
#endif
```

available in net46+ and netstandard1.3

If it's quicker than the `Task.Run` solution, then we could either implement both with this switch and add net46 in the framework moniker (but then poor perf on net45, document this) or reimplement it in net45 if possible

treziac:
While doing some benchmark I got a segfault when running the benchmark on net45 (don't remember if I ever did it before) Will investigate and post more detailed explanations in an other issue later this day/week, but it seems complicated to reproduce in VS (was done in CLI)

Benchmark looks nice with RunContinuationsAsynchronously - same throughput on netcoreapp and net46 than current impl. I couldn't run benchmark with Task.Run(), my pool thread won't take those many task runned (I managed to run for 4M but not 5M - did not dug yet)

Here are all the solutions I see:

1) Keep it like it is now, and specify in the doc that await should not be used on ProduceAsync (or that the other methods involving Poll should not be run synchronously after the delivery.
2) Remove the Task ProduceAsync method (just keep the void ProduceAsync with IDelveryHandler)
3) Same as 2), but only for net45, keep those methods for net46+ and netcoreapp with RunContinuationsAsynchronously
4) Have the Task.Run in all case
5) Have Task.Run in net45 and RunContinuationsAsynchronously in net46 and netstandard and document accordingly
6) Have everything which interfere with Poll async (Flush, Dispose, QueryMetadata...), I think this should work (given the user know how to use await/async to avoid deadlock) but not sure it's the best approach
in our case :)
7) Re-implement a synchronization context to mock the behaviour of RunContinuationsAsynchronously, not even sure it's possible (it's an internal thing in Task) 
8) find an other way

My preference is in 3 or even 2 (remove tasked ProduceAsync)
 
---
The async method is not preserving order of delivery handle, which is counter-intuitive for kafka _(but yet the easiest method to use so the one most newcomer will use, then thinking it's badly made)_ The only usage I see for this method is
1) if we want to use ContinueWith on it
2) if we want to produce synchrounously
3) if we want to use it the way most async methods are meant: produce a message (or multiple), do some work, await it so we can optimize cpu /network and have a nice async app

For 2), to avoid synchronization context, we could create a ProduceSync method. It would use a single DeliveryHandler created in private in the Producer, and would return a Message or throw if there was an error

For 1), using tasked ProduceAsync first does not keep order, second has some impact on GC (there is as many objects created as there are message send, where with the IDeliveryHandler method we have only one instance) 
We can also create an overload without an IDeliveryHandler but with an Action<Mesage>. Then use an internal IDeliveryHandler (as previously) in producer to set the deliveryHandler and make it easy to use. 

For 3) ... well, that's a more advanced case where people will have to read the doc and use IDeliveryHandler properly (that's the only case where the current ProduceAsync would make sense)

Sure, it's a big change on the interface, and not the typical way to do it in c#. But typical way does not work with kafka (order and co), so should we keep it? If for simple usage we have simple methods which do the work, this seems ok for me

treziac:
Well, I found a candidate for 8, run StartPollTask in a custom TaskScheduler which creates it's own thread.
awaiting ProduceAsync will resume on the TaskScheduler.Default so not on the same thread
I'm still not 100% that it's a correct thing to do or what is exactly the workflow under this, and it's unorthodox for the less, but still an option

mhowlett:
`TaskCreationOptions.RunContinuationsAsynchronously` is a great find. I'd say we almost certainly want to be using this where supported.

another option: don't fix on net45, fix on net46 + netstandard1.3.

your option 5 makes the most sense to me, assuming there is nothing fundamentally wrong with the `Task.Run` approach (pretty sure it's ok). And it's simple to implement.

--- 

I think the `Task` based methods would make sense in say a web app where each request might produce only one message to Kafka but there may be hundreds of requests in flight simultaneously. You don't want to be blocking in the request handlers with a synchronous call + not requiring a callback makes the code nicer.

chiru1205:
Just want to post a question here. I'm using kafka for a web app. We have multiple requests that send messages to kafka simutaneoulsy. Every request to kafka is run in a separate thread pool thread there by not blocking the main request thread. Our load test failed because too many threads lead to thread pool starvation.  To  resolve this thread pool stravation, I ran the Kafka code synchronously instead of spinning a new thread for every send request. Here is a sample of my code that runs for every request
using(var producer = new Producer(config)){
producer.ProduceAsync(topic,null,val);
producer.Flush();
}
Code uses .NET 4.6.1 framework.
Keeping producer.Flush() is adding an average of 6 seconds to the request time. Removing producer.Flush() is adding only about 3 - 45 ms per request. Using Await is adding roughly 7 seconds sometimes up to 25 seconds. All I wanted is to make sure my message is delivered or not. How can I know of a delivery or failure without adding much time to my request. Do you guys suggest using IDeliveryHandler overload?

I don't know whats the point of using Flush for every request? Some kind of documentation is required on whats the advantage of using this flush method. I don't need to know how many messages are still pending to be delivered.  

Need your help. Thank you

treziac:
1) you should use a single Producer instance in your webapp registred as singleton. Confluent.Kafka is thread safe. Just be sure to call Poll in a single thread (this is done automatically with the default constructor) Also be sure to dispose when closing the app. Creating a Peoducer or Consumer instance is costly because it must contact and connect with brokers at initialization and you won't be able to batch messages together
2) I would advise leaving the framework deal with pool thread himself if you are not confident with it. Just use case and async/await properly
3) You can use continueWith. In your case:
- if you need to make sure the message is delivered before to continue your request, use await ProduceAsync or .Result
- if you just want to know if the request was successful, use ContinueWith on your ProduceAsync. You can then filter on ProduceAsync who failed by throwing (due to disconnected or other problem: the task won't be flagged completed)

treziac:
Also be sure to register to the Error event on the producer instance. In case of disconnection to broker, this is were you will get the errors.

chiru1205:
@treziac Thank you so much for your input. Much appreciated. Here is what I have done based on your recommendation. I have some more questions and please respond when you have time.

Await is still very bad because its adding an average of 7 seconds to the request. At any given time we have around 60000 or more simultaneous requests to send messages to the queue. I did create a singleton instance of the producer in my class as a static property. Here is some sample code I have used to ensure a singleton instance of producer all the time
Public class Test{
private static Producer<Null, string> _kafkaProducer;
private static readonly object Padlock = new object();
  public static Producer<Null, string> GetProducer()
        {
            lock (Padlock)
            {
                if (_kafkaProducer != null) return _kafkaProducer;
                _kafkaProducer = new Producer<Null, string>(GetProducerConfig(), null, new 
                                                                   StringSerializer(System.Text.Encoding.UTF8));
              //set up event handlers here
              
                return _kafkaProducer;
            }
        }

  public static void SendMessage(KafkaMessage message)
        {
            //var deliveryHandler = new KafkaMessageDeliveryHandler();
            GetProducer().ProduceAsync("TestAttempt", null, JsonConvert.SerializeObject(message));
       }
}

I use the SendMessage method for all my send requests. This method will internally call GetProducer which will inturn return the singleton instance.

I didn't quite understand using ContinueWith. All I'm looking for is whether my message is successfully sent and acknowledged by leader and followers. I don't need the offset of the message either as long as its saved to the queue on all servers.  

I did a benchmark test with 6 Million messages on my windows machine , kafka running locally. Here is the result. This is all done with the singleton class(thank you @treziac ) . 80 messages per millisecond is very good in my case.
image

One problem I have noticed through looking at the snapshot of  memory is for every message sent TypedTaskDeliveryHandlerShim objects are automatically created even though my producerasync doesn't have any IDeliveryHandler. So for 6 Million requests 6 million objects were created.
Here is a screenshot
image

I will wait for your valuable suggestions. Thank you so much one again.

treziac:
Shortly: 
- ProduceAsync will create an IDeliveryHandler if you don't use the IDeliveryHandler overload
```csharp
private Task<Message<TKey, TValue>> Produce(string topic, TKey key, TValue val, DateTime? timestamp, int partition, bool blockIfQueueFull)
{
    var handler = new TypedTaskDeliveryHandlerShim(key, val);
    var keyBytes = KeySerializer?.Serialize(key);
    var valBytes = ValueSerializer?.Serialize(val);
    producer.Produce(topic, valBytes, 0, valBytes == null ? 0 : valBytes.Length, keyBytes, 0, keyBytes == null ? 0 : keyBytes.Length, timestamp, partition, blockIfQueueFull, handler);
    return handler.Task;
}
```
that's why you have so many objects created (they will live as long as you don't get the message). Two solution:

- You should be able to create a singleton DeliveryHandler to avoid this to the max. For example (did not test, just quickly wrote it, not sure it will work properly on asp.net, will have to test this we):

```csharp
public class SimpleProducerDeliveryHandler : IDeliveryHandler<K, V>
        {
            public SimpleProducerDeliveryHandler ()
            {
            }

            public bool MarshalData { get { return false; } }

            public void HandleDeliveryReport(Message<K, V> deliveryReport)
            {
                        //your process here
            }
        }
+ set `delivery.report.only.error` to true in config.
```
Note that you won't find your key/value in the message with this, but value wont be copied for the lifetime of the delivery so huge improvement. If the singleton does not work, you can use this deliveryhandler for each request, this will still decrease the memory used a lot. You will still have one TypedDeliveryHandlerShim by call though

@mhowlett, perhaps we should provide a Produce which take a IDeliveryHandler instead of a IDeliveryHandler<TKey, TValue> to avoid creating TypedDeliveryHandlerShim every time.

If those TypedDeliveryHandlerShim is the bottomneck, use the untyped Producer

chiru1205:
@treziac . Thank you for the reply. I did benchmark with delivery.report.only.error to true. I noticed memory leak. TypedDeliveryHandlerShim objects are not GC'ed if the setting is set to true. However I think for the error call backs they will be GC'ed. Removing the setting and using a single instance of IDeliveryHandler did work. As you mentioned TypedDeliveryHandlerShim  objects are created for every send request.
image

No matter what, thread pool starvation is still there. I did a load test and noticed several network errors which tells me that these send requests are using up a lot of threads behind the scenes. I don't how to resolve this issue. One difference I noticed is with Singleton instances request times came down significantly. Just in the range I'm looking for. 

I have to resolve this using up of threads by the kafka client.  One final option I had is to rely on the Rest Proxy but I won't get high through put with that. Any suggestions?

treziac:
By default, Typed Producer will always start its own thread (Task.StartNew with LongRunning flag which typically result in a thread) to poll, that's why you should have only one producer.



I didn't test the delivery.report.only.error myself but I though it's what you may want if you don't care about standard delivery. You should NOT used the Task ProduceAsync then or you will indeed have memoryleaks, just the one with IDeliveryHandler. Also we have to make some improvement on our side to reduce the overhead (as we typically except callbacks with the current API)
Will make an APS.NET Core sample tomorrow I think to show the singleton usage.

mhowlett:
thanks very much @chiru1205 for pointing out the memory leak bug - I made an issue for it.

If you don't care about delivery reports under any circumstance, you can also create a `Producer` with `disableDeliveryReports=false`. But I think that's going to have the memory leak issues in the same scenarios as `delivery.report.only.error=true` (need to fix).

Now to the scenario where there are no memory leaks - my assumption was that the cost of creating and GCing a `TypedTaskDeliveryHandlerShim` is small compared to everything else that is going on to make the produce request and I preferred to go with something that is slightly less efficient than impacting the API for marginal gain. What is the cost of this really? that is my question. I suspect the cost to throughput is not that high, but if you make a good case we'll make changes to the API for efficiency reasons. Of course if we can optimise this behind the scenes as well we will (i just had a quick think and couldn't think how).

The Producer only uses one thread on the C# side of things behind the scenes. I'm not sure what your thread pool problems are.

chiru1205:
@mhowlett , Thank you for the reply. We can afford GC'ing not an issue at this point. However if the send message is huge lets say 2000 character string or more, TaskDeliveryObject has to maintain and give this information back, so objects size will be huge and it adds up. With memory leaks this will only lead to disaster. 
Coming to thread pool issue.I can't really figure out if its a thread pool issue or something else. Our load test failed with bunch of 500 errors and network errors .  Initially I thought it was threads because the call back to IDeliveryHandler is all executed in a separate thread? So for each send request I thought a new thread is borrowed from the pool to execute the call back. I have to execute a test tomorrow to test this theory.  One other thing I'm doing in my code is to flush every 100000 messages which I'm not sure is needed or not. I'm thinking to remove this code and only flush before the producer instance is disposed. I have taken @treziac suggestion and created a singleton producer instance. 
Here is what I thought so far on the threads
1. For each Instance of producer of consumer(long running)
2. One thread for every call back if IDeliveryHandler is used.
Correct me If I'm wrong.

mhowlett:
> Thank you for the reply. We can afford GC'ing not an issue at this point. However if the send message is huge lets say 2000 character string or more, TaskDeliveryObject has to maintain and give this information back, so objects size will be huge and it adds up. With memory leaks this will only lead to disaster.

Your best bet at optimising for your scenario at the moment - which shouldn't be too much work - is to use an `IDeliveryHandler` variant of `ProduceAsync` on the non-serializing `Producer` and do the serialization yourself. We'll consider our options for improving performance of `Producer<K,V>`. 

> So for each send request I thought a new thread is borrowed from the pool to execute the call back.

currently no. each callback is executed on the same thread (the background polling thread). This is probably going to change however, see #92, #180. 

You should only Flush directly before Dispose - there is only downside to `flush`ing regularly (a pause whilst you flush).

treziac:
Didn't have time to do my sampes finally :/

The thread issue was by using multiple producer (one thread each time you created a new instance), it should vanish with a singleton

treziac:
Coming back to the PR:
We have to pay attention with Task.Run() in .net45 as it will potentially use a thread from ThreadPool, which could block other thing. 
Also, Task.StartNew(LongRunning) used in StartCallbackTask is not assured to create a new thread which won't be used anywhere else (it does in the current implementation, but nothing prevent it to change in future) Perhaps we should use a new thread which can't be used by the thread pool to be sure there won't be any deadlock?

chiru1205:
@mhowlett , memory is stable when I used the non-serializing Producer.  I have noticed another memory leak issue this morning while testing a disconnect scenario. 
Steps to recreate the issue
1. Shut down Kafka Broker(in my case local).
2. Use a non-serializing producer with a Delivery Handler.
3. Send thousands of messages to see if Delivery Handler can handle the error logging. I received error messages as expected. But memory was pushed to the limit and finally got the Access violation exception. Here is a screenshot of the memory leak.
image

I think there should be some kind of mechanism to halt message send requests if broker is not connected or down.

mhowlett:
> memory is stable when I used the non-serializing Producer.

good to know.

> Send thousands of messages to see if Delivery Handler can handle the error logging. I received error messages as expected. But memory was pushed to the limit and finally got the Access violation exception. Here is a screenshot of the memory leak.

not good to know. I can't think of anything you might be doing wrong. very important to investigate/fix this one. Filing a separate bug for this #208. Please post any further info there.

thanks very much for reporting @chiru1205 - much appreciated.

treziac:
Did some benchmark on windows (windows 10) with a local zookeeper & broker (installed on windows)
Run benchmark with 3 tests on release for the three monikers.

Run in this order:
netcoreapp1.1
net46
net45
netcoreapp1.1
net46
net45

Results are formated this way :
**framework**
first run - second run (runs without the change to compare to actual behaviour)
The four lines are for Task, DeliveryHandler, Poll, Consume in this order

**netcoreapp1.1** (RunContinuationsAsynchronously)
```
326  309  300 - 332  309  298 - 348  320  306 (340  317  317 - 349  307  316 - 336  313  307)
634  638  630 - 616  617  641 (597  620  646 - 641  640  640)
134  137  137 - 126  126  127 (123  124  124 - 131  131  131)
137  136  136 - 126  126  127 (124  124  124 - 131  131  131)
```

**net46** (RunContinuationsAsynchronously)
```
266  259  262 - 268  263  268 (278  265  272 - 278  266  270)
544  510  510 - 541  521  518 (491  494  494 - 506  510  545)
149  149  149 - 130  130  131 (135  135  136 - 130  129  130)
146  149  149 - 131  131  131 (136  136  136 - 130  130  130)
```

**net45** (Task.Run)
```
268  266  264 - 272  272  265 (280  265  270 - 278  271  270)
488  492  489 - 511  515  515 (513  549  519 - 515  546  518)
126  126  126 - 142  141  142 (142  142  143 - 148  149  149)
127  127  127 - 142  142  142 (143  143  143 - 149  149  149)
```


- There doesn't seem to be any significant impact, either with Task.Run or RunContinuationsAsynchronously (for windows at least)

- DeliveryHandler (at least how they are done in benchmark) is way quicker

- DeliveryHandler is 20% quicker on netcoreapp compared to net framework

- Consuming is very slow, and strangely, consistent in one run but not between two runs from same scenario. This may come from my hardware (slow HDD)

Concerning the integration tests passing, if I remember correctly xunit use a special thread context, so this must come from this. If we really want to test it in CI, we will have to use a dotnet run rather than dotnet test I think and write a program to check

chiru1205:
@treziac would you mind adding those code samples to the examples folder for reference.

mhowlett:
awesome! this is exactly what I was thinking. I don't think it's important enough to test this in CI that we do something special to accommodate this.

LGTM 🍉☀️

chiru1205:
@treziac @mhowlett . Still having issues with our load test. Most of the wait time from our monitoring tool(New Relic) seems to be .NET CLR.
image

My Application is three layed. UI --> webService --> Framework. Framework has a singleton producer that sends requests. Service and Framework are on the same server and we run 8 instances of the webservice load balanced. So we are running 8 instances of producer all on separate boxes, I'm not sure what is causing this .NET CLR wait time. I want to eliminate thread issue(not sure). we run a load of 25K transactions per minute in our load test. This whole thing is puzzling.

mhowlett:
Are you waiting for a delivery report / Task before returning from the
request handler?
If so (and this is necessary), you may need to tune your config parameters
to optimize for latency (rather than throughput).

chiru1205:
Not doing any kind of awaits actually. I'm using a logger in my IDeliveryHandler and logging only error-ed messages.

mhowlett:
Have you tried disabling only the produce requests and comparing results?

The ProduceAsync method will block by default if librdkafka's internal queue fills up. you could try turning that off (`blockIfQueueFull` parameter on ProduceAsync) and seeing if you get any exceptions indicating the queue was full.

If the librdkafka's internal queue isn't full ProduceAsync won't block.

chiru1205:
> Have you tried disabling only the produce requests and comparing results?

Do you mean using {"delivery.report.only.error",true } ?

mhowlett:
no, I mean commenting out use of the producer - test the time taken to do everything but produce messages (what is the delta)

chiru1205:
I'll post an update tomorrow. In the mean time I'll play with options locally

mhowlett:
thanks - btw appreciate that you're giving the client a decent workout and all the feedback we're getting back on this.

treziac:
@chiru1205 can on you also enable statistics?
Set `statistics.interval.ms` to non 0 (60000 for example) and register to `OnStatistic`, then pos json here.
Don't think it will help, but if you have high latency between client and broker this could point us to something

Also agree with matt, try setting blockIfQueueFull to false. What is the code you are using now? (DeliveryHandler, Task...), there was multiple options discussed about so not sure what you are doing.
Thanks again for feedback!

treziac:
Also if you can do a CPU analysis on your project to see where exactly you take so much time, this would be great ;)

chiru1205:
@mhowlett  @treziac.  Currently I'm using a non serializing producer(single instance) with ProduceAsync and IDeliveryHandler. I get 500 internal server errors and 503 errors sometimes. 503 tells me (no threads available to carry out the incoming request). I see a lot of .NET CLR wait time not sure what it could be. I'm running a load test with out any producer requests to eliminate other code creating issues.
 
To be more clear on how things are done. Here is a step by step information
1. UI sends a request to call the webservice to get some information or post information
2. Webservice then calls our application layer
3. Each method in my application layers calls a static method(send message) in a class(resides in the application layer). Below is  how the class is structured. Note that single instance of DeliveryHandler and Single Instance of Producer are created as static variables.


public class TestProducer : ITestProducer 
    {
        private static readonly ILog Logger = LogManager.GetLogger(typeof(TestProducer));
        private static Producer _kafkaProducer;
        private static readonly object Padlock = new object();
        private static readonly IDeliveryHandler TestDeliveryHandler = new KafkaMessageDeliveryHandler();


        private static Producer GetProducer()
        {
            lock (Padlock)
            {


                if (_kafkaProducer != null) return _kafkaProducer;
                _kafkaProducer = new Producer(GetProducerConfig());
                Logger.Info($"Thread: {Thread.CurrentThread.ManagedThreadId},Getting a new Instance of Producer: {_kafkaProducer.GetHashCode()}");


                _kafkaProducer.OnError += (_, e) =>
                {

                };

                _kafkaProducer.OnLog += (sender, e) =>
                {


                };

                _kafkaProducer.OnStatistics += (_, json)
                        =>
                {

                    // Logger.Info($"Statistics: {json}");
                };
                return _kafkaProducer;
            }
        }

        public static void SendMessage(byte[] messageBytes)
        {
            

            GetProducer().ProduceAsync("Test", null,0,0,messageBytes,0,messageBytes?.Length ?? 0,true,TestDeliveryHandler);            
          
        }

}

I want to get to the point where I have to wait for the result of the  produce request but I'm not there yet. For now I don't actually do any kind of awaits or wait or result on the produceaysnc call . I leave it to the delivery handler to log errors when there is an error sending messages. 

Something is blocking the SendMessage and inturn blocking my webservice call and adding bunch of time to it. I thought SendMessage method with producerasync and IDelivery Handler is async becuase there is a polling thread behind the scenes which reports back to the handler on the class at some other time(when the response is received) and our call doesn't have to sit and wait for the result. 

I'll post the result of my load test later this evening. I want to see what my .NET CLR Wait time is without produce requests. However before the produce request code we passed LNP with flying colors. Our service calls are never above 200 ms. After the produce request code we are at 2 to 3 seconds.  And there is a pattern in the time jump too. When our load is up, errors pop-up(500, 503,403,404) and time jumps to 2-3 seconds. Its recovers in about a minute or two and time comes down to around 600 ms . 
image

I want to get to the bottom of this probably by next week sometime. 
I want to thank you guys for your help here. It means a lot.

treziac:
Thi lock seems strange - I would have put it after the return, then lock, rhen redo the check. But I don't think this is your problem. Can you also copy your KafkaMessageDeliveryHandler class? 

    ///     Methods of this interface (IDeliveryHandler) will be executed on the poll thread and will
    ///     block other operations - consider this when implementing.

Thanks!

chiru1205:
Here is my DeliveryHandler


 
        private class TestDeliveryHandler : IDeliveryHandler
        {
            private static int Count { get; set; }
            /// <summary>
            /// Gets whether or not to marshal key and value data from librdkafka when the delivery
            /// report is available. Usually this should return true. Return false for a small
            /// performance improvement if you don't need this information.
            /// </summary>
            public bool MarshalData => true;
            private static readonly object CountLock = new object();
            public void HandleDeliveryReport(Message message)
            {
                lock (CountLock)
                {
                    Count = Count + 1;
                }
                //ToDO: Handle error messages, Save offsets somewhere else
               if (message.Error.HasError)
                {
                   Logging.LogException()
                }
            }


            
            
        }

mhowlett:
yes, it would seem thread synchronization is likely your problem. see if you can think of a way to avoid the lock every time you access the producer (note that using a constructed producer is thread safe - you just need to avoid the race condition on construction). Also, maybe just count the number of errors? also take a look at https://msdn.microsoft.com/en-us/library/system.threading.interlocked.increment(v=vs.110).aspx

chiru1205:
> Thi lock seems strange - I would have put it after the return, then lock, rhen redo the check

@treziac , its a mistake. I totally missed that one. 

@mhowlett Like you mentioned I will rewrite the code and see if that works.

chiru1205:
@mhowlett @treziac . I was pulled into a different project this week. I'm still working on doing an LNP and post some results here soon. I took care of the thread sync issue.

chiru1205:
@mhowlett  @treziac . Here is an update from my load test today. I have basically updated my Producer wrapper like below and removed the lock statements and thread sync issues.

    public class AssessmentProducer : IAssessmentProducer
    {

        #region Class Methods
        
        private readonly Producer _kafkaProducer;
        private readonly IDeliveryHandler _kafkaDeliveryHandler;
       
        public AssessmentProducer()
        {
            _kafkaDeliveryHandler = new KafkaMessageDeliveryHandler();
            _kafkaProducer = new Producer(GetProducerConfig());
           _kafkaProducer.OnError += (_, e) =>
            { //TODO:

            };

            _kafkaProducer.OnLog += (sender, e) =>
            {
                //TODO:

            };

            _kafkaProducer.OnStatistics += (_, json)
                    =>
            {

                // Logger.Info($"Statistics: {json}");
            };

        }

        public void SendMessage(KafkaMessage message)
        {
            var messageDatainBytes = System.Text.Encoding.UTF8.GetBytes(JsonConvert.SerializeObject(message));
           _kafkaProducer.ProduceAsync("Test", null, messageDatainBytes, _kafkaDeliveryHandler);
        }
        
        /// <summary>
        /// Get Producer Configuration
        /// </summary>
        /// <returns></returns>
       private static Dictionary<string, object> GetProducerConfig() => new Dictionary<string, object>
        {
            {"bootstrap.servers", "localhost:9092"},
            {"message.max.bytes", 2000000}, //2MB message size
            {"queue.buffering.max.ms",100 },
            {"message.send.max.retries", 3 },
            {"client.id", Dns.GetHostName()},
            {
                "default.topic.config", new Dictionary<string, object>
                {
                    {"acks", "all"}, //updated to support leader & follower acknowledgments
                    {"message.timeout.ms", 3000}
                }
            }
        };


still noticed high response times and 22 minutes into the test all kafka brokers were down. 

Look at netstat on the brokers , I found too many TCP connections. Hard limit on cent os TCP connections is 16358, if that exceeds server won't accept any connections. Most of the connections seems to be in TIME_WAIT mode.
image

I think this is whats killing my good response times.

Our load is very high , in a 4 hour test we send close to 20 millions send requests.

Issue #116, opened 2017-03-23
Key serializer should not be set at producer construction


MrGlenThomas:
This is not a bug, but more a question on the design. I am building a system based on a large, existing set of data that has key properties of different data types. Some may be ints, some may be longs, some may be GUIDs, some may be strings. In partitioning examples I have seen, the key being used is a UserId, which could be an int. On another topic my key could be a RequestId, which is a GUID.

The producer is designed so that one instance can be used to produce messages for many topics, which is the optimum way of using the producer, but the key type for each topic may be different and the key serializer is set in the constructor. So, I am forced to have producer instances for each possible key type as the produce method does not accept a key serializer. This feels very awkward and I'm not sure how best to handle this situation. I don't want to have multiple producer instances (one per key serializer type), which I must select from based on the key type I am using. And I don't want to have to convert all my keys to strings and then always use a string serializer. I feel that the key serializer should be as dynamic as the topic name.

Please let me know what you think about this.

treziac:
See mutli producer example for an example of use (GetSerializingProducer)

```
using (var producer = new Producer(config))
            {
                // sProducer1 is a lightweight wrapper around a Producer instance that adds
                // (string, string) serialization. Note that sProducer1 does not need to be
                // (and cannot be) disposed.
                var sProducer1 = producer.GetSerializingProducer<string, string>(new StringSerializer(Encoding.UTF8), new StringSerializer(Encoding.UTF8));

                // sProducer2 is another lightweight wrapper around kafkaProducer that adds
                // (null, int) serialization. When you do not wish to write any data to a key
                // or value, the Null type should be used.
                var sProducer2 = producer.GetSerializingProducer<Null, int>(new NullSerializer(), new IntSerializer());

                // write (string, string) data to topic "first-topic", statically type checked.
                sProducer1.ProduceAsync("first-topic", "my-key-value", "my-value");

                // write (null, int) data to topic "second-data". statically type checked, using
                // the same underlying producer as the producer1.
                sProducer2.ProduceAsync("second-topic", null, 42);

                // producers are NOT tied to topics. Although it's unusual that you might want to
                // do so, you can use different serializing producers to write to the same topic.
                sProducer2.ProduceAsync("first-topic", null, 107);

                // ProducerAsync tasks are not waited on - there is a good chance they are still
                // in flight.
                producer.Flush();
            }
```

MrGlenThomas:
[...]

PR #183, opened 2017-05-16
removing UNSTABLE designation from non-serializing Producer

mhowlett:
there has been plenty of feedback that people want the non-serializing producer, e.g. #116, so removing the UNSTABLE designation. @ewencp

### DIFF
--- a/src/Confluent.Kafka/Producer.cs
+++ b/src/Confluent.Kafka/Producer.cs
@@ -31,12 +31,6 @@ namespace Confluent.Kafka
 {
     /// <summary>
     ///     Implements a high-level Apache Kafka producer (without serialization).
-    ///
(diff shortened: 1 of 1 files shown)


mhowlett:
If you LGTM this one @treziac i'll merge it.

treziac:
LGTM Some of the API are still unstable though (ProduceAsync, I'm looking at you)

mhowlett:
that's a good point (though it also applies to the serializing producer as well of course).

Another thing I started thinking is @chiru1205's perf. complaint over in #180 has a (very) small chance of impacting this design.

Ok, this can stay unstable in 0.9.6.. but I'll make good of your LGTM later as I see fit :-) thanks.

PR #184, opened 2017-05-16
Obsoleting or removing methods that can potentially block indefinitely

mhowlett:
@treziac @edenhill @ewencp 

People who use variants of these methods with an infinite timeout usually don't realize they can't cancel out of them or even that they generally might want to - better if they don't exist.

Actually we should add variants that take a `CancellationToken` as well, but that can happen in a latter PR.

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -151,13 +151,6 @@ namespace Confluent.Kafka
         public bool Consume(out Message<TKey, TValue> message, TimeSpan timeout)
             => Consume(out message, timeout.TotalMillisecondsAsInt());
 
-        /// <summary>
-        ///     Refer to <see cref="Consume(out Message{TKey, TValue}, int)" />.
-        ///     
-        ///     [UNSTABLE-API] - prefer to use <see cref="Poll()"/> / <see cref="OnMessage"/> instead of this method.
-        /// </summary>
-        public bool Consume(out Message<TKey, TValue> message)
-            => Consume(out message, -1);
 
         /// <summary>
         ///     Poll for new consumer events, including new messages
@@ -165,7 +158,7 @@ namespace Confluent.Kafka
         ///     event).
         /// </summary>
         /// <param name="millisecondsTimeout"> 
-        ///     The maximum time to block (in milliseconds).
+        ///     The maximum time to block (in milliseconds), or -1 to block indefinitely.
         /// </param>
         public void Poll(int millisecondsTimeout)
         {
@@ -195,12 +188,13 @@ namespace Confluent.Kafka
 
         /// <summary>
         ///     Poll for new consumer events, including new messages
-        ///     ready to be consumed(which will trigger the OnMessage
+        ///     ready to be consumed (which will trigger the OnMessage
         ///     event).
         /// </summary> 
         /// <remarks>
         ///     Blocks indefinitely until a new event is ready.
         /// </remarks>
+        [Obsolete("Indefinite polling cannot be cancelled so it's usually not what you want. Removing this method to discourage.", false)]
         public void Poll()
             => Poll(-1);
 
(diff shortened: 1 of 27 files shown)


src/Confluent.Kafka/Consumer.cs
```diff
@@ -151,21 +151,14 @@ public bool Consume(out Message<TKey, TValue> message, int millisecondsTimeout)
...
-            => Consume(out message, -1);
 
         /// <summary>
         ///     Poll for new consumer events, including new messages
         ///     ready to be consumed (which will trigger the OnMessage
         ///     event).
         /// </summary>
         /// <param name="millisecondsTimeout"> 
-        ///     The maximum time to block (in milliseconds).
+        ///     The maximum time to block (in milliseconds), or -1 to block indefinitely.
```
treziac: Add that -1 should not typically be used. Same remarks for the Poll(TimeSpan) : TimeSpan.Infinite should typically not be used

src/Confluent.Kafka/Consumer.cs
```diff
@@ -195,12 +188,13 @@ public void Poll(TimeSpan timeout)
...
         /// <summary>
         ///     Poll for new consumer events, including new messages
-        ///     ready to be consumed(which will trigger the OnMessage
+        ///     ready to be consumed (which will trigger the OnMessage
         ///     event).
         /// </summary> 
         /// <remarks>
         ///     Blocks indefinitely until a new event is ready.
         /// </remarks>
+        [Obsolete("Indefinite polling cannot be cancelled so it's usually not what you want. Removing this method to discourage.", false)]
```
treziac: Just "Use Poll overload with finite timeout" is enough I think (obsolete is already discouraging)

src/Confluent.Kafka/Consumer.cs
```diff
@@ -1015,7 +995,7 @@ public void Poll(TimeSpan timeout)
         ///     event).
         /// </summary>
         /// <param name="millisecondsTimeout"> 
-        ///     The maximum time to block (in milliseconds)
+        ///     The maximum time to block (in milliseconds), or -1 to block indefinitely.
```
treziac: Same remark

src/Confluent.Kafka/Consumer.cs
```diff
@@ -1034,6 +1014,7 @@ public void Poll(int millisecondsTimeout)
         /// <remarks>
         ///     Blocks indefinitely until a new event is ready.
         /// </remarks>
+        [Obsolete("Indefinite polling cannot be cancelled so it's usually not what you want. Removing this method to discourage.", false)]
         public void Poll()
```
treziac: Same remark

src/Confluent.Kafka/Producer.cs
```diff
@@ -360,6 +360,7 @@ public int Poll(TimeSpan timeout)
         /// <returns>
         ///     Returns the number of events served.
         /// </returns>
+        [Obsolete("Indefinite polling cannot be cancelled so it's usually not what you want. Removing this method to discourage.", false)]
```
treziac: Same remark

src/Confluent.Kafka/Producer.cs
```diff
@@ -624,6 +643,7 @@ public int Flush(int millisecondsTimeout)
         /// <returns>
         ///     Refer to <see cref="Flush(int)" />.
         /// </returns>
+        [Obsolete("This method cannot be cancelled so it's usually not what you want. Removing this method to discourage.", false)]
```
treziac: Same remark

src/Confluent.Kafka/Producer.cs
```diff
@@ -616,6 +617,24 @@ public int Flush(int millisecondsTimeout)
...
         /// <summary>
+        ///     Wait until all outstanding produce requests and delievery report
+        ///     callbacks are completed. Refer to <see cref="Flush(int)" /> for
+        ///     more information.
+        /// 
+        ///     [UNSTABLE-API] - the semantics and/or type of the return value is
+        ///     subject to change.
+        /// </summary>
+        /// <param name="timeout">
+        ///     The maximum length of time to block.
```
treziac: Same remark (avoid TimeStamp.Timeout which may block kafka thread or something like that)
The maximum time to block (no length of)

src/Confluent.Kafka/Producer.cs
```diff
@@ -1211,20 +1223,29 @@ public int Flush(int millisecondsTimeout)
...
+        ///     callbacks are completed. Refer to <see cref="Flush(int)" /> for
+        ///     more information.
+        /// 
         ///     [UNSTABLE-API] - the semantics and/or type of the return value is
         ///     subject to change.
         /// </summary>
-        public int Flush()
-            => producer.Flush();
+        /// <param name="timeout">
+        ///     The maximum length of time to block.
```
treziac: Same remark

src/Confluent.Kafka/Consumer.cs
mhowlett: i'd agree with that

PR #186, opened 2017-05-17
Fix" Too many open files" in travis osx

treziac:
Now that curl is fixed on travis-osx for .net core, there shouldn't be any more random error on build


mhowlett:
I thought I'd tried this - but maybe not. Tests pass anyway, so LGTM.

PR #187, opened 2017-05-17
adding metadata to project files

mhowlett:
@treziac @edenhill


src/Confluent.Kafka/Confluent.Kafka.csproj
```diff
@@ -1,6 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <Authors>Confluent Inc.;Andreas Heider</Authors>
+    <Description>Confluent's .NET Client for Apache Kafka</Description>
+    <Copyright>Copyright 2016-2017 Confluent Inc., Andreas Heider</Copyright>
+    <PackageTags>Kafka;Confluent;librdkafka</PackageTags>
```
treziac: Only this csproj metadata are important - the others are not packaged so I dont think the metadata will be shown anywhere. Some important metadata missing here:

- Add` <PackageProjectUrl>` (https://github.com/confluentinc/confluent-kafka-dotnet/) It's the first thing I visit before installing a package. And `<RepositoryUrl>`
- Add `<PackageIconUrl>` (confluent or kafka? Or is there an icon with both? 64*64 with transparent background)
- Add in the description that it use librdkafka.
- Add the `<PackageLicenseUrl>` (given this, should we add all the licences listed on librdkafka in this repo?)
- For later, we will want to add `<PackageReleaseNotes>`. Always good to know what we upgrades directly in nuget package manager
- Add Summary and modify Description : in nuget package manager and nuget.org, Summary is the line just under the package name (so one of the most important), description is shown when you focus on this package. Title should be short, Description can be long. Example:
```xml
<Summary >.NET Client for Apache Kafka</Summary >
<Description>
Confluent.Kafka is a high-performance, reliable client for Apache Kafka

It is a wrapper around librdkafka, a finely tuned C client common to all other confluent kafka client, kept up to date with the latest addition in kafka protocol.
Maintained by Confluent, the company founded by the creators of Apache Kafka.
</Description>
```
**Edit**: I said a mistake first, this is not `Title` but `Summary` we want. And this one is not supported by msbuild yet. So we either have to create a .nuspec, either create the package without it (it will then truncate the description, so we may just have the first line in description to be what we want and this will be ok)

examples/AdvancedConsumer/AdvancedConsumer.csproj
```diff
@@ -1,7 +1,10 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <Authors>Andreas Heider;Confluent Inc</Authors>
+    <Authors>Andreas Heider;Confluent Inc.</Authors>
+    <Description>Example demonstrating how to consume messages.</Description>
+    <Copyright>Copyright 2016-2017 Confluent Inc., Andreas Heider</Copyright>
+    <PackageTags>Kafka;Confluent;librdkafka</PackageTags>
```
treziac: Package tags are unnecessary as this will never be packed (the other metadata neither, but they are still informative). Same comment for every project except Confluent.Kafka

src/Confluent.Kafka/Confluent.Kafka.csproj
treziac: Note that for Confluent.Kafka.StrongName, we can modify the description when invoking dotnet build/pack (`/p:Description="Confluent.Kafka strongly signed. Use Confluent.Kafka for normal usage"`)

examples/AdvancedConsumer/AdvancedConsumer.csproj
mhowlett: Unnecessary, but still informative as you say so my thought was I might as well put this info everywhere whilst i was at it with the main project. I've changed my mind though - the info is available elsewhere, comments go out of date, they aren't critical here etc. so i'll take them out.

src/Confluent.Kafka/Confluent.Kafka.csproj
mhowlett: yep - i'll make some changes based on the above. thanks!

For release notes, I think we can add these with each PR and reset after a release. It's not a big deal if we forget / don't do this on each PR - the main thing is to make sure they are correct before tagging the repo with the release tag. The process is not ideal - there is potential to forget to do this, but I'm not sure what is better.

mhowlett: license: librdkafka is not included in the package so we just need to point to the license in our repo (which includes the rdkafka-dotnet one). I've pointed to master, which is technically incorrect, but in practice this won't change and I don't want to update this link for every release.

summary + description: pity this tag doesn't exist. I like the current description for this. Leaving description short and sweat for now, which I think is fine and means I don't have to think hard about what I think should and shouldn't be in here and how to say it.

release notes: I just put a link to the github releases page, which is where I'll update release notes.

examples/AdvancedConsumer/AdvancedConsumer.csproj
mhowlett: I also took out Authors (which is redundant as it's also included in the copyright notice in the source files).

PR #202, opened 2017-05-28
reworking serialization API for schema registry support

mhowlett:
this is a continuation of @treziac 's #129. 

I think I finally have a definite opinion on changes that should be made to (de)serializers for schema registry support (a new idea).

1. configuration works differently to java. the idea is a (de)serializer can take the configuration dict in their constructors, and the constructor of Producer/Consumer then ignores any properties used by the serializers. I also added capability for key and value serializers to be configured separately. we can update the string, int and long serializers to also optionally be configured from the main config dict. we can do the same architectural pattern for custom partitioners (allow them to be configured). 
2. no `IDisposable` - I don't see when this would be necessary. Only suggested it because it's there in java - what am i missing? we can have this if needed, but this raises questions of ownership - it's most convenient if the `Producer` and `Consumer` take ownership of the (de)serializers, and this is not without precedent in dotnet (see StreamReader), but it's a pattern that can cause confusion (in general, though probably not in our case) and best avoided. 
3. actual serializers and custom partitioner classes are not going in the configuration dict - they will be only arguments to the constructor. they can however be parameterized by values from the config. I think this is an optimal balance.
4. no explicit implementation of interfaces (for reasons noted in #129).

to use: 

```
using (var producer = new Producer<string, string>(config, new StringSerializer(config), new AvroSerializer(config)))
{
}
```

(the old way of configuring string serializer will remain as well).

note: WIP tests not updated, just putting the code changes out for comment.

### DIFF
--- a/examples/SimpleProducer/Program.cs
+++ b/examples/SimpleProducer/Program.cs
@@ -31,9 +31,13 @@ namespace Confluent.Kafka.Examples.SimpleProducer
             string brokerList = args[0];
             string topicName = args[1];
 
-            var config = new Dictionary<string, object> { { "bootstrap.servers", brokerList } };
+            var config = new Dictionary<string, object> 
+            { 
+                { "bootstrap.servers", brokerList },
+                { "string.serializer.encoding.key", "UTF8" }
+            };
 
-            using (var producer = new Producer<Null, string>(config, null, new StringSerializer(Encoding.UTF8)))
+            using (var producer = new Producer<Null, string>(config, null, new StringSerializer(config)))
             {
                 Console.WriteLine($"{producer.Name} producing on {topicName}. q to exit.");
 
--- a/src/Confluent.Kafka/Serialization/StringSerializer.cs
+++ b/src/Confluent.Kafka/Serialization/StringSerializer.cs
@@ -16,6 +16,8 @@
 
 using System;
 using System.Text;
+using System.Collections.Generic;
+
 
 namespace Confluent.Kafka.Serialization
 {
@@ -37,16 +39,7 @@ namespace Confluent.Kafka.Serialization
             this.encoding = encoding;
         }
 
-        /// <summary>
-        ///     Encodes a string value in a byte array.
-        /// </summary>
-        /// <param name="val">
-        ///     The string value to serialize.
-        /// </param>
-        /// <returns>
-        ///     <paramref name="val" /> encoded in a byte array (or null if <paramref name="val" /> is null).
-        /// </returns>
-        public byte[] Serialize(string val)
+        private byte[] Serialize(string val)
         {
             if (val == null)
             {
@@ -55,12 +48,31 @@ namespace Confluent.Kafka.Serialization
             return encoding.GetBytes(val);
         }
 
-        byte[] ISerializer<string>.Serialize(string topic, string data)
+        /// <summary>
+        ///     Encodes a string value in a byte array.
+        /// </summary>
+        /// <param name="data">
+        ///     The string value to serialize.
+        /// </param>
+        /// <param name="topic">
+        ///     The topic associated with the data (ignored by this serializer).
+        /// </param>
+        /// <param name="isKey">
+        ///     true: deserialization is for a key, 
+        ///     false: deserializing is for a value.
+        /// </param>
+        /// <returns>
+        ///     <paramref name="data" /> encoded in a byte array (or null if <paramref name="data" /> is null).
+        /// </returns>
+        public byte[] Serialize(string topic, string data, bool isKey)
         {
             return Serialize(data);
         }
 
-        void IDisposable.Dispose()
-        { }
+        /// <summary>
+        ///     Configuration properties used by the serializer.
+        /// </summary>
+        public IEnumerable<KeyValuePair<string, object>> Configuration 
+            => new List<KeyValuePair<string, object>>();
     }
 }
--- a/src/Confluent.Kafka/Serialization/StringSerializer.cs
+++ b/src/Confluent.Kafka/Serialization/StringSerializer.cs
@@ -17,6 +17,7 @@
 using System;
 using System.Text;
 using System.Collections.Generic;
+using System.Linq;
 
 
 namespace Confluent.Kafka.Serialization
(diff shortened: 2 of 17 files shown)


treziac:
1. Perhaps we should rater add the possibility to have config not used in librdkafka (and generate warnings like in java) rather than forbiding unknown conf (so we won't have to do the parsing to remove unknown config). No definitive opinion here...
2. Yep, I just implemented it since you asked ;) I also prefer the user to take care of it if needed rather than giving the control to the handle and having potential bug.
3. For the StringSerializer for example, Configuration would contain encoding, that's it? And all serializer would preferably have a constructor with a IEnumerable<KeyValuePair<String, object>>? In this case, we could also add a (facultative) base class wich deal with this automatically.
4. With this configuration object, we can't know the type of config object (for librdkafka as it's an external library it's not such an issue, here it's a bit more problematic) I would rather have properties, and have the base constructor I mentioned before do the mapping (so both can be used, constructor with initialization and/or constructor with config - I never liked initializing things at runtime (dictionary) rather than compile timpe (properties), I assume I'm not the only C# dev in this case

Perhaps add the StringSerializer with encoding as config to show exactly what you want to do

5. This may come later, but I see nothing which would prevent specifying the serializer/deserializer/partitioner via config also with this. For people who like everything in config (very useful when done via file without having to look at code with miscroservices and co, I must admit ;) )

edenhill:
> Perhaps we should rater add the possibility to have config not used in librdkafka (and generate warnings like in java) rather than forbiding unknown conf (so we won't have to do the parsing to remove unknown config). No definitive opinion here...

I believe the Java client's way of handling this is more error prone , it is too easy to typo a config property and miss the warning message.
What the Go and Python wrappers do is intercept the configuration before it passed to librdkafka, to find and remove the properties the wrapper handles.

One way of to achieve this is through the new interceptor interface:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L3360

treziac:
Ok - that's how it's done now, we can continue. How to make sure we won't use a config which would be later added to librdkafka and be in conflict? Are there some reserved word which will never be used by librdkafka (like beginning by _, or anything else?)

mhowlett:
Yes, I'm intercepting the configuration before passing to librdkafka. I don't think the new interceptor interface is going to be useful here as serializers are completely on the C# side and there isn't much to it. And yes, failing fast is generally better (as the current implementation does).

I'm not worried about config name clashes. 1. The config names are long and descriptive. 2. in the event there ever was overlap, nothing would break and the C# side greedily taking the config parameter (which only happens if the user configured it to do so) is probably what is wanted.

(De)serializers don't need to have a `IEnumerable<KeyValuePair<String, object>>` constructor, but doing so provides a mechanism to allow them to be easily configured from a config file. I don't think there is much benefit to a helper base class - not sure what it would do. Ok, I'll add implementations to this PR to demonstrate.

We could create another class for config info with static type checking that implements `IEnumerable<KeyValuePair<String, object>>` to provide it to the `Producer` and `Consumer`. For maximum safety, the interface to the `Producer` and `Consumer` should be constrained to this, however I believe what we have is an optimal compromise. 

We could ultimately allow serializers / partitioners to be specified in the config dict as well, however I don't think this is going to be necessary (I know earlier I was throwing this about as an option). I expect people will want the ability to be able to configure their serializers / partitioners via the one config dict (which could be read in from a file), but need to configure the actual serializer / partitioner implementation will be much less common. Also, in such scenarios this can be solved with some extra code. I could be wrong, but as you say, the design allows for it, so we can do it in the future if it comes up. I'm very much hoping to keep things simple - configuring of actual implementations is via constructor arguments. Configuration of strings/bools/values via config file.

src/Confluent.Kafka/Serialization/StringSerializer.cs
```diff
@@ -33,25 +38,88 @@ public class StringSerializer : ISerializer<string>
...
+
+        /// <summary>
+        ///     Initializes a new StringSerializer class instance.
+        /// </summary>
+        /// <param name="config">
+        ///     Configuration properties.
+        /// </param>
+        public StringSerializer(IEnumerable<KeyValuePair<string, object>> config)
+        {
+            if (config.Count(ci => ci.Key == "string.serializer.encoding.key") > 0)
```
mhowlett: Unfortunately the implementation is a bit ugly - the object is unaware of it's context when constructed, so it needs to be prepared for use with either keys or values, even though this is known at construction time and the serializer instance will only be ever used for one or the other.

examples/SimpleProducer/Program.cs
```diff
@@ -31,9 +31,13 @@ public static void Main(string[] args)
...
 
-            var config = new Dictionary<string, object> { { "bootstrap.servers", brokerList } };
+            var config = new Dictionary<string, object> 
+            { 
+                { "bootstrap.servers", brokerList },
+                { "string.serializer.encoding.key", "UTF8" }
+            };
 
-            using (var producer = new Producer<Null, string>(config, null, new StringSerializer(Encoding.UTF8)))
+            using (var producer = new Producer<Null, string>(config, null, new StringSerializer(config)))
```
mhowlett: An example of use. You'd only configure the StringSerializer via the config object like this if you wanted to allow it to be configurable via the same config file used for the producer. This provides an easy mechanism for that. Otherwise, you'd just pass in the encoding object directly.

src/Confluent.Kafka/Serialization/StringSerializer.cs
mhowlett: this is possibly unacceptable and making me think something more similar to java would be better - have a `Configure` method on `ISerializer`, but rather than return void, this could return a new config dictionary with the config property the serializer is interested in removed. there is a minor problem if two serializers want to pick up on the same property e.g. 'schema.registry.url'. If the calls to `Configure` are chained, the property will disappear, but the original dict could be used for both then the returned results merged.

From a user perspective, i think it's slightly worse (I prefer it being explicit that the config dict is being used to configure the serializers). It should probably be a runtime error if the user tries to configure both in the serializer constructor and via a config parameter.

Ok. I'll do another PR with the above variant and see how it looks.

mhowlett:
closing in favor or #205

PR #205, opened 2017-05-29
serialization API rework

mhowlett:
alternative to #202, following on from #129 

I like this better. Similar to Java, except (de)serializers specified via constructors can be configured using the config passed into the `Producer` or `Consumer`.

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -90,8 +90,8 @@ namespace Confluent.Kafka
 
             consumer = new Consumer(
                 config.Where(item => 
-                    adjustedConfig1.Count(ci => ci.Key == item.Key) > 0 &&
-                    adjustedConfig2.Count(ci => ci.Key == item.Key) > 0
+                    adjustedConfig1.Any(ci => ci.Key == item.Key) &&
+                    adjustedConfig2.Any(ci => ci.Key == item.Key)
                 )
             );
 
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -65,8 +65,6 @@ namespace Confluent.Kafka
             KeyDeserializer = keyDeserializer;
             ValueDeserializer = valueDeserializer;
 
-            // TODO: allow deserializers to be set in the producer config IEnumerable<KeyValuePair<string, object>>.
-
             if (KeyDeserializer == null)
             {
                 if (typeof(TKey) != typeof(Null))
@@ -87,7 +85,12 @@ namespace Confluent.Kafka
                 ValueDeserializer = (IDeserializer<TValue>)new NullDeserializer();
             }
 
-            consumer = new Consumer(config);
+            consumer = new Consumer(
+                config.Where(item => 
+                    !keyDeserializer.Configuration.Select(c => c.Key).Contains(item.Key) &&
+                    !valueDeserializer.Configuration.Select(c => c.Key).Contains(item.Key)
+                )
+            );
 
             consumer.OnConsumeError += (sender, msg) 
                 => OnConsumeError?.Invoke(this, msg);
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -85,15 +85,15 @@ namespace Confluent.Kafka
                 ValueDeserializer = (IDeserializer<TValue>)new NullDeserializer();
             }
 
-            var adjustedConfig1 = KeyDeserializer.Configure(config, true);
-            var adjustedConfig2 = ValueDeserializer.Configure(config, false);
+            var configWithoutKeyDeserializerProperties = KeyDeserializer.Configure(config, true);
+            var configWithoutValueDeserializerProperties = ValueDeserializer.Configure(config, false);
 
-            consumer = new Consumer(
-                config.Where(item => 
-                    adjustedConfig1.Any(ci => ci.Key == item.Key) &&
-                    adjustedConfig2.Any(ci => ci.Key == item.Key)
-                )
+            var configWithoutDeserializerProperties = config.Where(item => 
+                configWithoutKeyDeserializerProperties.Any(ci => ci.Key == item.Key) &&
+                configWithoutValueDeserializerProperties.Any(ci => ci.Key == item.Key)
             );
+
+            consumer = new Consumer(configWithoutDeserializerProperties);
 
             consumer.OnConsumeError += (sender, msg) 
                 => OnConsumeError?.Invoke(this, msg);
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -85,10 +85,13 @@ namespace Confluent.Kafka
                 ValueDeserializer = (IDeserializer<TValue>)new NullDeserializer();
             }
 
+            var adjustedConfig1 = KeyDeserializer.Configure(config, true);
+            var adjustedConfig2 = ValueDeserializer.Configure(config, false);
+
             consumer = new Consumer(
                 config.Where(item => 
-                    !keyDeserializer.Configuration.Select(c => c.Key).Contains(item.Key) &&
-                    !valueDeserializer.Configuration.Select(c => c.Key).Contains(item.Key)
+                    adjustedConfig1.Count(ci => ci.Key == item.Key) > 0 &&
+                    adjustedConfig2.Count(ci => ci.Key == item.Key) > 0
                 )
             );
 
--- a/src/Confluent.Kafka/Internal/Extensions/String.cs
+++ b/src/Confluent.Kafka/Internal/Extensions/String.cs
@@ -23,9 +23,9 @@ namespace Confluent.Kafka
     /// <summary>
     ///     Extension methods for the <see cref="String"/> class.
     /// </summary>
-    public static class StringExtensions
+    internal static class StringExtensions
     {
-        internal static Encoding AsEncoding(this string encodingName) 
+        internal static Encoding ToEncoding(this string encodingName) 
         {
             switch (encodingName)
             {
@@ -39,7 +39,7 @@ namespace Confluent.Kafka
                     return Encoding.ASCII;
                 case "Unicode":
                     return Encoding.Unicode;
-                case "BigEndignUnicode":
+                case "BigEndianUnicode":
                     return Encoding.BigEndianUnicode;
                 default:
                     throw new ArgumentException("unknown string encoding: " + encodingName);
--- a/src/Confluent.Kafka/Internal/Extensions/String.cs
+++ b/src/Confluent.Kafka/Internal/Extensions/String.cs
@@ -27,20 +27,17 @@ namespace Confluent.Kafka
     {
         internal static Encoding ToEncoding(this string encodingName) 
         {
-            switch (encodingName)
+            switch (encodingName.ToLower())
             {
-                case "UTF8":
+                // these names match up with static properties of the Encoding class
+                case "utf8":
                     return Encoding.UTF8;
-                case "UTF7":
+                case "utf7":
                     return Encoding.UTF7;
-                case "UTF32":
+                case "utf32":
                     return Encoding.UTF32;
-                case "ASCII":
+                case "ascii":
                     return Encoding.ASCII;
-                case "Unicode":
-                    return Encoding.Unicode;
-                case "BigEndianUnicode":
-                    return Encoding.BigEndianUnicode;
                 default:
                     return Encoding.GetEncoding(encodingName);
             }
--- a/src/Confluent.Kafka/Internal/Extensions/String.cs
+++ b/src/Confluent.Kafka/Internal/Extensions/String.cs
@@ -42,7 +42,7 @@ namespace Confluent.Kafka
                 case "BigEndianUnicode":
                     return Encoding.BigEndianUnicode;
                 default:
-                    throw new ArgumentException("unknown string encoding: " + encodingName);
+                    return Encoding.GetEncoding(encodingName);
             }
         }
     }
--- a/src/Confluent.Kafka/Internal/Extensions/String.cs
+++ b/src/Confluent.Kafka/Internal/Extensions/String.cs
@@ -1,0 +1,49 @@
+// Copyright 2016-2017 Confluent Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Refer to LICENSE for more information.
+
+using System;
+using System.Text;
+
+
+namespace Confluent.Kafka
+{
+    /// <summary>
+    ///     Extension methods for the <see cref="String"/> class.
+    /// </summary>
+    public static class StringExtensions
+    {
+        internal static Encoding AsEncoding(this string encodingName) 
+        {
+            switch (encodingName)
+            {
+                case "UTF8":
+                    return Encoding.UTF8;
+                case "UTF7":
+                    return Encoding.UTF7;
+                case "UTF32":
+                    return Encoding.UTF32;
+                case "ASCII":
+                    return Encoding.ASCII;
+                case "Unicode":
+                    return Encoding.Unicode;
+                case "BigEndignUnicode":
+                    return Encoding.BigEndianUnicode;
+                default:
+                    throw new ArgumentException("unknown string encoding: " + encodingName);
+            }
+        }
+    }
+}
--- a/src/Confluent.Kafka/Serialization/IDeserializer.cs
+++ b/src/Confluent.Kafka/Serialization/IDeserializer.cs
@@ -15,6 +15,8 @@
 // Refer to LICENSE for more information.
 
 using System;
+using System.Collections.Generic;
+
 
 namespace Confluent.Kafka.Serialization
 {
@@ -22,7 +24,7 @@ namespace Confluent.Kafka.Serialization
     ///     Implement this interface to define a deserializer 
     ///     for a particular type T.
     /// </summary>
-    public interface IDeserializer<T> : IDisposable
+    public interface IDeserializer<T>
     {
         /// <summary>
         ///     Deserialize a byte array to an instance of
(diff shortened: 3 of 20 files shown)


treziac:
Will review more throughfully, but I like this one more than the other :+1:

mhowlett:
no need to review in depth until I finish it off - only let me know if you disagree with the general idea. I think i've probably iterated through every possible way of doing this by now - finally happy with the this approach.

mhowlett:
this is not strictly in scope for 0.9.6, but it will unblock dev work on schema registry integration and a lot of thinking went into getting the api right, so it'd be good to get it in. 

@treziac if you feel inclined (and of course thanks for the original PR on which it is based).

src/Confluent.Kafka/Serialization/StringDeserializer.cs
```diff
@@ -37,21 +45,69 @@ public StringDeserializer(Encoding encoding)
         }
 
         /// <summary>
+        ///     Initializes a new StringDesrializer class instance.
+        ///     The encoding encoding to use must be provided via <see cref="Consumer" /> configuration properties.
+        /// </summary>
+        public StringDeserializer()
+        {
```
treziac: I thought norm was to write on one line?
```
public StringSerializer()
{ }
```

src/Confluent.Kafka/Serialization/StringDeserializer.cs
```diff
@@ -25,6 +29,10 @@ public class StringDeserializer : IDeserializer<string>
     {
         Encoding encoding;
 
+        private const string KeyEncodingConfigParam = "dotnet.string.deserializer.encoding.key";
```
treziac: Perhaps public, as it's usable from outside? Never sure on those ones

src/Confluent.Kafka/Serialization/StringSerializer.cs
```diff
@@ -25,6 +29,9 @@ public class StringSerializer : ISerializer<string>
     {
         private Encoding encoding;
 
+        private const string KeyEncodingConfigParam = "dotnet.string.serializer.encoding.key";
```
treziac: same remark: should it be public?

src/Confluent.Kafka/Serialization/StringSerializer.cs
```diff
@@ -37,21 +44,71 @@ public StringSerializer(Encoding encoding)
         }
 
         /// <summary>
+        ///     Initializes a new StringSerializer class instance.
+        ///     The encoding encoding to use must be provided via <see cref="Producer" /> configuration properties.
+        /// </summary>
+        public StringSerializer()
+        {
```
treziac: { }

test/Confluent.Kafka.UnitTests/Serialization/String.cs
```diff
@@ -25,10 +25,10 @@ public class StringTests
         [Fact]
         public void SerializeDeserialize()
         {
-            Assert.Equal("hello world", new StringDeserializer(Encoding.UTF8).Deserialize(new StringSerializer(Encoding.UTF8).Serialize("hello world")));
-            Assert.Equal("ឆ្មាត្រូវបានហែលទឹក", new StringDeserializer(Encoding.UTF8).Deserialize(new StringSerializer(Encoding.UTF8).Serialize("ឆ្មាត្រូវបានហែលទឹក")));
-            Assert.Equal("вы не банан", new StringDeserializer(Encoding.UTF8).Deserialize(new StringSerializer(Encoding.UTF8).Serialize("вы не банан")));
-            Assert.Equal(null, new StringDeserializer(Encoding.UTF8).Deserialize(new StringSerializer(Encoding.UTF8).Serialize(null)));
+            Assert.Equal("hello world", new StringDeserializer(Encoding.UTF8).Deserialize("topic", new StringSerializer(Encoding.UTF8).Serialize("topic", "hello world")));
```
treziac: Also add tests using the new config for serialization/Deserialization (this may be both integration tests as we need a working producer for this, and unit test on the configure function)

src/Confluent.Kafka/Serialization/StringSerializer.cs
```diff
@@ -37,21 +44,71 @@ public StringSerializer(Encoding encoding)
...
+                        .Value
+                        .ToString()
+                        .ToEncoding();
+                }
+                catch
+                {
+                    throw new ArgumentException($"{keyOrValue} StringSerializer encoding configuration parameter was specified twice.");
+                }
+
+                if (encoding != null)
```
treziac: This must be before the line 87. Here we will always throw as encoding will just have been set

src/Confluent.Kafka/Serialization/StringSerializer.cs
```diff
@@ -37,21 +44,71 @@ public StringSerializer(Encoding encoding)
...
+                {
+                    encoding = config
+                        .Single(ci => ci.Key == propertyName)
+                        .Value
+                        .ToString()
+                        .ToEncoding();
+                }
+                catch
+                {
+                    throw new ArgumentException($"{keyOrValue} StringSerializer encoding configuration parameter was specified twice.");
```
treziac: ToEncoding may also throw, separate both (first "Signle" with the try/catch with this message, second with ToEncoding witout try catch as it will simply throw if necessary
Also, when you throw an exception from a catch block, you should include the exception catch in InnerException so you keep the origin (here we would have lost the encoding issue for example)

src/Confluent.Kafka/Serialization/StringDeserializer.cs
```diff
@@ -37,21 +45,69 @@ public StringDeserializer(Encoding encoding)
...
+                        .Value
+                        .ToString()
+                        .ToEncoding();
+                }
+                catch
+                {
+                    throw new ArgumentException($"{keyOrValue} StringDeserializer encoding configuration parameter was specified twice.");
+                }
+
+                if (encoding != null)
```
treziac: See below (should be before encoding = ...)

src/Confluent.Kafka/Serialization/StringDeserializer.cs
```diff
@@ -37,21 +45,69 @@ public StringDeserializer(Encoding encoding)
...
+                {
+                    encoding = config
+                        .Single(ci => ci.Key == propertyName)
+                        .Value
+                        .ToString()
+                        .ToEncoding();
+                }
+                catch
+                {
+                    throw new ArgumentException($"{keyOrValue} StringDeserializer encoding configuration parameter was specified twice.");
```
treziac: See below (two different exceptions involved here)

src/Confluent.Kafka/Internal/Extensions/String.cs
```diff
@@ -0,0 +1,49 @@
...
+                case "UTF32":
+                    return Encoding.UTF32;
+                case "ASCII":
+                    return Encoding.ASCII;
+                case "Unicode":
+                    return Encoding.Unicode;
+                case "BigEndianUnicode":
+                    return Encoding.BigEndianUnicode;
+                default:
+                    throw new ArgumentException("unknown string encoding: " + encodingName);
```
mhowlett: Yep, this is better.

In general, I like the philosophy of there only being one way to do things, but I'm thinking I'll leave the switch statement here because the cases match up exactly with inline C# configuration (which people are likely to try) and I don't think having more than one variant matters. However, I'll point to the below URL for the available options in the API docs:

https://msdn.microsoft.com/en-us/library/system.text.encoding(v=vs.110).aspx

src/Confluent.Kafka/Consumer.cs
```diff
@@ -87,7 +85,15 @@ public class Consumer<TKey, TValue> : IDisposable
                 ValueDeserializer = (IDeserializer<TValue>)new NullDeserializer();
             }
 
-            consumer = new Consumer(config);
+            var adjustedConfig1 = KeyDeserializer.Configure(config, true);
+            var adjustedConfig2 = ValueDeserializer.Configure(config, false);
+
+            consumer = new Consumer(
+                config.Where(item => 
+                    adjustedConfig1.Count(ci => ci.Key == item.Key) > 0 &&
```
mhowlett: good idea (ditto in other places)

src/Confluent.Kafka/Serialization/IDeserializer.cs
```diff
@@ -26,13 +30,19 @@ public interface IDeserializer<T>
...
         ///     The serialized representation of an instance
         ///     of type T to deserialize.
         /// </param>
         /// <returns>
         ///     The deserialized value.
         /// </returns>
-        T Deserialize(byte[] data);
+        T Deserialize(string topic, byte[] data);
+
+        /// <include file='../include_docs.xml' path='API/Member[@name="IDeserializer_Configure"]/*' />
```
mhowlett: We have quite a bit of API doc duplication, so I want to start doing this in other places as well.

src/Confluent.Kafka/Serialization/StringDeserializer.cs
mhowlett: yep, I think it makes a bit more sense to have this public.

src/Confluent.Kafka/Serialization/StringDeserializer.cs
mhowlett: i'm not sure TBH but i like your suggestion better.

src/Confluent.Kafka/Serialization/StringSerializer.cs
mhowlett: good points.

src/Confluent.Kafka/Serialization/StringSerializer.cs
mhowlett: good catch.

test/Confluent.Kafka.UnitTests/Serialization/String.cs
mhowlett: given my slip-up with the logic above, good point.

src/Confluent.Kafka/Serialization/StringSerializer.cs
```diff
@@ -26,32 +30,94 @@ public class StringSerializer : ISerializer<string>
         private Encoding encoding;
 
         /// <summary>
+        ///     Name of the encoding configuration parameter for the case of key serialization.
```
edenhill: The use of "case" reads a bit strange to a Swede like me.
"Configuration property name for key serialization encoding type" or some such is clearer to me.

src/Confluent.Kafka/include_docs.xml
```diff
@@ -0,0 +1,35 @@
+<API>  
+
+<Member name="ISerializer_Configure">  
+    <summary>
+        Configure the serializer using relevant configuration parameter(s) in <paramref name="config" /> (if present)
```
edenhill: 'relevant' is a bit vague, is this serializer-specific configuration, or standard librdkafka config that is used by the serializer?

src/Confluent.Kafka/include_docs.xml
```diff
@@ -0,0 +1,35 @@
...
+    </summary>
+    <param name="config">
+        A collection containing configuration parameter(s) relevant to this serializer.
+    </param>
+    <param name="isKey">
+        true: if this serializer instance is used to serialize keys,
+        false: if this serializer instance is used to serialize values.
+    </param>
+    <returns>
+        A configuration collection with configuration parameter(s) relevant to this seriaizer removed.
```
edenhill: seriaizer typo

src/Confluent.Kafka/include_docs.xml
```diff
@@ -0,0 +1,35 @@
...
+    </summary>
+    <param name="config">
+        A collection containing configuration parameter(s) relevant to this deserializer.
+    </param>
+    <param name="isKey">
+        true: if this deserializer instance is used to serialize keys,
+        false: if this deserializer instance is used to serialize values.
+    </param>
+    <returns>
+        A configuration collection with configuration parameter(s) relevant to this deseriaizer removed.
```
edenhill: typo

test/Confluent.Kafka.UnitTests/Serialization/String.cs
```diff
@@ -25,13 +29,101 @@ public class StringTests
...
+                var deserializer = new StringDeserializer(Encoding.UTF32);
+                deserializer.Configure(config, false);
+            }
+            catch (ArgumentException)
+            {
+                return;
+            }
+
+            Assert.True(false, "Exception expected");
+        }
```
edenhill: add a negative test for a bogus encoding type

src/Confluent.Kafka/Consumer.cs
```diff
@@ -87,7 +85,15 @@ public class Consumer<TKey, TValue> : IDisposable
                 ValueDeserializer = (IDeserializer<TValue>)new NullDeserializer();
             }
 
-            consumer = new Consumer(config);
+            var adjustedConfig1 = KeyDeserializer.Configure(config, true);
+            var adjustedConfig2 = ValueDeserializer.Configure(config, false);
+
+            consumer = new Consumer(
+                config.Where(item => 
+                    adjustedConfig1.Any(ci => ci.Key == item.Key) &&
```
mhowlett: The (de)serializers could misbehave, but I don't see it as a real problem or any way of avoiding it except by maintaining a list of librdkafka properties, which I don't think is worth the overhead. The fundamental problem is we don't know which properties are librdkafka properties - I don't think thinking about this backwards helps.

will add a comment.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -87,7 +85,15 @@ public class Consumer<TKey, TValue> : IDisposable
                 ValueDeserializer = (IDeserializer<TValue>)new NullDeserializer();
             }
 
-            consumer = new Consumer(config);
+            var adjustedConfig1 = KeyDeserializer.Configure(config, true);
```
mhowlett: Yep, even better would be `configWithout[Key|Value]SerializerProperties`.

src/Confluent.Kafka/Internal/Extensions/String.cs
```diff
@@ -0,0 +1,49 @@
...
+namespace Confluent.Kafka
+{
+    /// <summary>
+    ///     Extension methods for the <see cref="String"/> class.
+    /// </summary>
+    internal static class StringExtensions
+    {
+        internal static Encoding ToEncoding(this string encodingName) 
+        {
+            switch (encodingName)
```
mhowlett: i'd agree with that.

src/Confluent.Kafka/Internal/Extensions/String.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                case "UTF8":
+                    return Encoding.UTF8;
+                case "UTF7":
+                    return Encoding.UTF7;
+                case "UTF32":
+                    return Encoding.UTF32;
+                case "ASCII":
+                    return Encoding.ASCII;
+                case "Unicode":
```
mhowlett: I took the position of that being stupid Microsoft's fault, but I'd agree with you it's better to take this out (as it's perpetuating something confusing). UTF16 will still be available via the default case in the switch statement. The idea was to provide config values matching up with all the property names of all the statically defined values in the `Encoding` class. `Unicode` is one of those.

src/Confluent.Kafka/Serialization/NullSerializer.cs
```diff
@@ -14,22 +14,33 @@
...
 // Refer to LICENSE for more information.
 
+using System;
+using System.Collections.Generic;
+
+
 namespace Confluent.Kafka.Serialization
 {
     /// <summary>
     ///     A dummy serializer for use with values that must be null (the <see cref="Null"/> class cannot be instantiated).
```
mhowlett: It's never used - it's only there to make the type system happy. So it is different to the other serializers, so I take the position that labelling it as dummy is somewhat meaningful..

src/Confluent.Kafka/Internal/Extensions/String.cs
edenhill: I suggest making "UTF16" an explicit encoding, removing the ambigious "Unicode", and also removing the default case and error out if the encoding type is not known.
It is better to fail early than to encode in an unexpected format, that data can never be fixed short of re-producing it.

src/Confluent.Kafka/Serialization/StringDeserializer.cs
```diff
@@ -26,32 +30,93 @@ public class StringDeserializer : IDeserializer<string>
...
+        /// </summary>
+        public const string ValueEncodingConfigParam = "dotnet.string.deserializer.encoding.value";
+
+
+        /// <summary>
         ///     Initializes a new StringDeserializer class instance.
         /// </summary>
         /// <param name="encoding">
-        ///     The encoding to use when deserializing.
+        ///     The encoding to use when deserializing. For available options, refer to:
```
mhowlett: well, the extra cases are just kind of for convenience because people who didn't read the documentation will probably try to use them (and I didn't see any reason for this not to work). they are all duplicates of options provided in this list. i didn't think this was worth mentioning - there is nothing people can do that isn't documented here.

src/Confluent.Kafka/Serialization/StringDeserializer.cs
```diff
@@ -26,32 +30,93 @@ public class StringDeserializer : IDeserializer<string>
...
+        ///     https://msdn.microsoft.com/en-us/library/system.text.encoding(v=vs.110).aspx
         /// </param>
         public StringDeserializer(Encoding encoding)
         {
             this.encoding = encoding;
         }
 
         /// <summary>
+        ///     Initializes a new StringDesrializer class instance.
+        ///     The encoding encoding to use must be provided via <see cref="Consumer" /> configuration properties.
```
mhowlett: good point.

src/Confluent.Kafka/Serialization/StringSerializer.cs
mhowlett: ok, i'll reword this (but will avoid use of the word 'type' too, as that's kind of overloaded in this context).

test/Confluent.Kafka.UnitTests/Serialization/String.cs
mhowlett: i was just checking if anyone would notice.

src/Confluent.Kafka/include_docs.xml
mhowlett: the serializer can use whatever it wants - generally not librdkafka config properties, but the possibility is there... i think this is ok.

src/Confluent.Kafka/Internal/Extensions/String.cs
mhowlett: I think you miss-understand the default case - it interprets the string as per the list half way down this page: https://msdn.microsoft.com/en-us/library/system.text.encoding(v=vs.110).aspx and throws an exception if not one of those. One of those cases is 'utf-16'. UTF16 would be a new symbol not present anywhere else. I'm indifferent as to whether it's added or not.

src/Confluent.Kafka/Serialization/StringDeserializer.cs
mhowlett: oh, I put the note about available encodings on the wrong constructor - meant this to be a note on the other one.

treziac:
There still are some issues on this - passing null in serializing consumer/producer does not work, and currently you can only use it when you have a single Producer or a single Consumer in your app (or config will contain some invalid values regarding librdkafka)
Either this is the target beahbiour and need to be documented, either we have to change that

PR #209, opened 2017-05-31
Add AddBrokers function

### LINKED ISSUE #204: Request: Expose AddBrokers API
I have a setup where my app connects to bunch of Kafka brokers running in Docker containers. Given nature of Docker, the IPs of those brokers can change when they restart or servers change. Now librdkafka is able to handle partial failures/changes in the topology of the Kafka cluster. However there is a case where all IPs of the brokers change completely. My app can discover those new IPs separately, however I have no way of telling librdkafka about those new brokers since it can't auto-discover them now. 

Based on this issue here (https://github.com/edenhill/librdkafka/issues/1073) it's suggested to use `rd_kafka_brokers_add()` This method is available in `SafeKafkaHandle` in `AddBrokers()` however it's not available to the producer or consumer. 

This is a feature request asking if it's possible to expose this internal API for public usage. 

Thanks

treziac:
Close #204 

@mhowlett 
Integation tests result here: https://travis-ci.org/treziac/confluent-kafka-dotnet/jobs/237930661

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -620,6 +620,21 @@ namespace Confluent.Kafka
         /// </summary>
         public Metadata GetMetadata(bool allTopics)
             => consumer.GetMetadata(allTopics);
+
+        /// <summary>
+        ///     Adds one or more brokers to the kafka bootstrap brokers.
+        ///     
+        ///     Use this when brokers IP address change at runtime.
+        /// </summary>
+        /// <param name="brokers">
+        ///     Coma-separated list of brokers in the same format as bootstrap.server
+        /// </param>
+        /// <example>
+        ///     AddBrokers("broker1:9092,broker2")
+        ///     AddBrokers("SSL://broker3:9000,ssl://broker2")
+        /// </example>
+        public int AddBrokers(string brokers)
+            => consumer.AddBrokers(brokers);
     }
 
     /// <summary>
@@ -1258,5 +1273,20 @@ namespace Confluent.Kafka
         /// </summary>
         public Metadata GetMetadata(bool allTopics)
             => kafkaHandle.GetMetadata(allTopics, null, -1);
+
+        /// <summary>
+        ///     Adds one or more brokers to the kafka bootstrap brokers.
+        ///     
+        ///     Use this when brokers IP address change at runtime.
+        /// </summary>
+        /// <param name="brokers">
+        ///     Coma-separated list of brokers in the same format as bootstrap.server
+        /// </param>
+        /// <example>
+        ///     AddBrokers("broker1:9092,broker2")
+        ///     AddBrokers("SSL://broker3:9000,ssl://broker2")
+        /// </example>
+        public int AddBrokers(string brokers)
+            => kafkaHandle.AddBrokers(brokers);
     }
 }
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -622,12 +622,18 @@ namespace Confluent.Kafka
             => consumer.GetMetadata(allTopics);
 
         /// <summary>
-        ///     Adds one or more brokers to the kafka bootstrap brokers.
-        ///     
-        ///     Use this when brokers IP address change at runtime.
+        ///     Adds one or more brokers to the Consumer's list of initial
+        ///     bootstrap brokers. 
+        ///
+        ///     Note: Additional brokers are discovered automatically as 
+        ///     soon as the Consumer connects to a broker by querying the 
+        ///     broker metadata. Calling this method is only required in 
+        ///     some scenarios where the address of all brokers in the 
+        ///     cluster changes.
         /// </summary>
         /// <param name="brokers">
-        ///     Coma-separated list of brokers in the same format as bootstrap.server
+        ///     Coma-separated list of brokers in the same format as 
+        ///     the bootstrap.server configuration parameter.
         /// </param>
         /// <example>
         ///     AddBrokers("broker1:9092,broker2")
@@ -1275,12 +1281,18 @@ namespace Confluent.Kafka
             => kafkaHandle.GetMetadata(allTopics, null, -1);
 
         /// <summary>
-        ///     Adds one or more brokers to the kafka bootstrap brokers.
-        ///     
-        ///     Use this when brokers IP address change at runtime.
+        ///     Adds one or more brokers to the Consumer's list of initial
+        ///     bootstrap brokers. 
+        ///
+        ///     Note: Additional brokers are discovered automatically as 
+        ///     soon as the Consumer connects to a broker by querying the 
+        ///     broker metadata. Calling this method is only required in 
+        ///     some scenarios where the address of all brokers in the 
+        ///     cluster changes.
         /// </summary>
         /// <param name="brokers">
-        ///     Coma-separated list of brokers in the same format as bootstrap.server
+        ///     Coma-separated list of brokers in the same format as 
+        ///     the bootstrap.server configuration parameter.
         /// </param>
         /// <example>
         ///     AddBrokers("broker1:9092,broker2")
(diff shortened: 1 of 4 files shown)


test/Confluent.Kafka.IntegrationTests/Tests/AddBroker.cs
```diff
@@ -0,0 +1,101 @@
...
+            }
+
+            using (var typedConsumer = new Consumer<Null, Null>(consumerConfig, new NullDeserializer(), new NullDeserializer()))
+            {
+                TestMetadata(
+                    () => typedConsumer.GetMetadata(false, TimeSpan.FromSeconds(3)),
+                    () => typedConsumer.AddBrokers(bootstrapServers));
+            }
+
+            void TestMetadata(Func<Metadata> getMetadata, Action addBrokers)
```
treziac: Local function, yeah =D

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -129,8 +129,8 @@ internal int Flush(int millisecondsTimeout)
             return OutQueueLength;
         }
 
-        internal long AddBrokers(string brokers)
```
treziac: Any reason for this to have been defined as long? (int on librdkafka)
Switched back to int (I don't think we will ever add so many brokers :) )

src/Confluent.Kafka/Consumer.cs
```diff
@@ -620,6 +620,21 @@ public Metadata GetMetadata(bool allTopics, TimeSpan timeout)
         /// </summary>
         public Metadata GetMetadata(bool allTopics)
             => consumer.GetMetadata(allTopics);
+
+        /// <summary>
+        ///     Adds one or more brokers to the kafka bootstrap brokers.
```
treziac: Partially copied from librdkafka, simplified

test/Confluent.Kafka.IntegrationTests/Tests/AddBroker.cs
```diff
@@ -0,0 +1,106 @@
...
+                { "group.id", Guid.NewGuid().ToString() },
+                { "bootstrap.servers", "unknown" },
+                { "session.timeout.ms", 6000 },
+                { "api.version.request", true }
+            };
+
+            using (var typedProducer = new Producer<Null, string>(producerConfig, null, new StringSerializer(Encoding.UTF8)))
+            {
+                TestMetadata(
+                    () => typedProducer.GetMetadata(false, null, TimeSpan.FromSeconds(3)),
```
treziac: Would take advise on how we should check if brokers are connected or not. This seemed a good way, perhaps there are better.

test/Confluent.Kafka.IntegrationTests/Tests/AddBroker.cs
```diff
@@ -0,0 +1,106 @@
...
+            }
+
+            using (var typedConsumer = new Consumer<Null, Null>(consumerConfig, new NullDeserializer(), new NullDeserializer()))
+            {
+                TestMetadata(
+                    () => typedConsumer.GetMetadata(false, TimeSpan.FromSeconds(3)),
+                    typedConsumer.AddBrokers);
+            }
+
+            void TestMetadata(Func<Metadata> getMetadata, Func<string, int> addBrokers)
```
treziac: Local function, yeah =D (first time I actually see an usage for this by using the parameter of the upper function)

src/Confluent.Kafka/Consumer.cs
mhowlett: More verbose summaries is what's blocking me LGTMing this PR - feel free to have another go at this, or I'll get to it myself later.

test/Confluent.Kafka.IntegrationTests/Tests/AddBroker.cs
```diff
@@ -0,0 +1,108 @@
...
+            {
+                try
+                {
+                    var metadata = getMetadata();
+                    Assert.True(false, "Broker should not be reached here");
+                }
+                catch (KafkaException e)
+                {
+                    Assert.Equal(ErrorCode.Local_Transport, e.Error.Code);
```
mhowlett: I don't actually understand why this should be expected.

mhowlett: (@edenhill - can you comment?)

test/Confluent.Kafka.IntegrationTests/Tests/AddBroker.cs
```diff
@@ -0,0 +1,108 @@
...
+                }
+
+                int brokersAdded = addBrokers(bootstrapServers);
+                Assert.True(brokersAdded > 0, "Should have added one broker or more");
```
mhowlett: We're adding the same broker again here so I would have expected this to return 0. I'm guessing the return value is equal to the number of valid brokers in the supplied list regardless of whether they are already known or not. correct @edenhill ?

mhowlett:
this is ready to merge, except I want to clarify my two questions in the integration test first.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -620,6 +620,27 @@ public Metadata GetMetadata(bool allTopics, TimeSpan timeout)
...
+        ///     some scenarios where the address of all brokers in the 
+        ///     cluster changes.
+        /// </summary>
+        /// <param name="brokers">
+        ///     Coma-separated list of brokers in the same format as 
+        ///     the bootstrap.server configuration parameter.
+        /// </param>
+        /// <example>
+        ///     AddBrokers("broker1:9092,broker2")
+        ///     AddBrokers("SSL://broker3:9000,ssl://broker2")
```
edenhill: the protocol notation is unnecessary since the protocol to use is set by `security.protocol`.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -620,6 +620,27 @@ public Metadata GetMetadata(bool allTopics, TimeSpan timeout)
         /// </summary>
         public Metadata GetMetadata(bool allTopics)
             => consumer.GetMetadata(allTopics);
+
+        /// <summary>
+        ///     Adds one or more brokers to the Consumer's list of initial
+        ///     bootstrap brokers. 
+        ///
+        ///     Note: Additional brokers are discovered automatically as 
+        ///     soon as the Consumer connects to a broker by querying the
```
edenhill: to any broker

src/Confluent.Kafka/Consumer.cs
```diff
@@ -620,6 +620,27 @@ public Metadata GetMetadata(bool allTopics, TimeSpan timeout)
...
+
+        /// <summary>
+        ///     Adds one or more brokers to the Consumer's list of initial
+        ///     bootstrap brokers. 
+        ///
+        ///     Note: Additional brokers are discovered automatically as 
+        ///     soon as the Consumer connects to a broker by querying the 
+        ///     broker metadata. Calling this method is only required in 
+        ///     some scenarios where the address of all brokers in the 
+        ///     cluster changes.
```
edenhill: Maybe add a note that there currently is no API to remove existing configured, added or learnt brokers.

test/Confluent.Kafka.IntegrationTests/Tests/AddBroker.cs
edenhill: Duplicates will be counted as added.

test/Confluent.Kafka.IntegrationTests/Tests/AddBroker.cs
edenhill: I'm guessing no brokers have been added yet, except for the "unknown" bootstrap.servers one, thus resulting in the Transport error since Metadata() can't find a functional broker connection.

mhowlett: oh whoops, I missed brokers.servers was being set to 'unknown'. thanks..

PR #210, opened 2017-05-31
[WIP] Proper error handling with delivery report

### LINKED ISSUE #201: TaskDeliveryHandlers leak memory when `delivery.report.only.error=true`
@chiru1205 points out in #180: I noticed memory leak. TypedDeliveryHandlerShim objects are not GC'ed if the setting is set to true. However I think for the error call backs they will be GC'ed. Removing the setting and using a single instance of IDeliveryHandler did work. As you mentioned TypedDeliveryHandlerShim objects are created for every send request.

This is a bug.

It's because they aren't freed until `SetResult` is called, but it never is most of the time if `delivery.report.only.error=true`.

### LINKED ISSUE #203: Error inconsistence
In Produce, we throw error when Produce return non 0 (librdkafka know it won't be produced),complete and fill the Error property of Message when it's a non immediate error
So in the case the topic is unknown, we will have an unthrown exception sometimes (a message with the error field not null, because it's not flagged as permanent unknown topic), then throw an exception - for the same call with the same argument
This seems shaky in the .NET client.
What about never throwing in the Produce and instead add a HandleException in IDeliveryHandler, which would be call instead of throwing? Produce would never throw in this case, the only errors from producer would come from OnError (and/or OnPoduceError if we add this- but that's also the problem between which one we chose)
I think there is many place now where we can get error, and we should document them more - in the message, OnConsumeError, OnError, throwing when producing...

I would go with this:

OnError: error in the internal library. Not bound to producing or consuming a specific message. Broker unavailable, network failure... The error ismore or less severe and could be only informative (link to a post in librdkafka)

OnConsumeError: failed to consume a message. This may or may not be specific to a message (in which case topic, partition and offset will be set accordingly in Message). See the Message.Error for more details

ProduceAsync: will not throw because of Confluent.Kafka (may because of async .NET framework) The delivery report may contain error, in which case it will be passed through IDeliveryHandler.HandleDeliveryReport(Message) with Error field set or IDeliveryHandler.HandleException(Error)

treziac:
Fix #203, fix #201

Note that for Task, I would rather setting task as not completed (SetException) when Message has an Error, this seems more correct

### DIFF
--- a/src/Confluent.Kafka/ErrorCode.cs
+++ b/src/Confluent.Kafka/ErrorCode.cs
@@ -29,6 +29,18 @@ namespace Confluent.Kafka
     public enum ErrorCode
     {
         /// <summary>
+        ///     Passed a delivery report with deliveryReport disabled.
+        ///     Message is still produced (fire and forget)
+        /// </summary>
+        Client_DisabledDeliveryReport = -999,
+
+        /// <summary>
+        ///     Passed a delivery report with delivery.report.only.error at true.
+        ///     Message is still produced (fire and forget)
+        /// </summary>
+        Client_TaskProduceWithOnlyErrorCallback = -998,
+
+        /// <summary>
         ///     Received message is incorrect
         /// </summary>
         Local_BadMsg = -199,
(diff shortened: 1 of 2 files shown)


mhowlett:
everything has changed since this PR... (1.0 is now out). not aware of any remaining issues here.

PR #215, opened 2017-06-01
[WIP] Cache log / error for event handlers registred at init

### LINKED ISSUE #137: Issue when Events are called before they are assigned
Events (OnLog but I assume also OnError and other) can be invoked before we can assign callbacks to them.

We should allow to set a callback before the handle is created to correct that. We could use the "log_cb" and "error_cb" config keys, they should not be used by the the users of confluent-kafka-dotnet.

Or perhaps the user should be able to define its own callbacks which would replace the one we created? I like it when library let me do low-level operation (though hidden enough so that nobody will try to override them) In this case, we need to define other config name

Simple reproductible code:

```
var config = new Dictionary<string, object> { { "bootstrap.servers", brokerList }, { "debug", "all" } };

using (var producer = new Producer<Null, string>(config, null, new StringSerializer(Encoding.UTF8)))
{
    producer.OnLog += (o, e) => Console.WriteLine("log");
}
```

will print (on windows) :
```
7|2017-04-03 17:59:13.415|rdkafka#producer-1|BRKMAIN| [thrd::0/internal]: :0/internal: Enter main broker thread
7|2017-04-03 17:59:13.415|rdkafka#producer-1|BROKER| [thrd:app]: kafka:9092/bootstrap: Added new broker with NodeId -1
7|2017-04-03 17:59:13.424|rdkafka#producer-1|BRKMAIN| [thrd:kafka:9092/bootstrap]: kafka:9092/bootstrap: Enter main broker thread
7|2017-04-03 17:59:13.424|rdkafka#producer-1|CONNECT| [thrd:kafka:9092/bootstrap]: kafka:9092/bootstrap: broker in state INIT connecting
7|2017-04-03 17:59:13.424|rdkafka#producer-1|STATE| [thrd::0/internal]: :0/internal: Broker changed state INIT -> UP
7|2017-04-03 17:59:13.425|rdkafka#producer-1|BROADCAST| [thrd::0/internal]: Broadcasting state change
7|2017-04-03 17:59:13.432|rdkafka#producer-1|CONNECT| [thrd:kafka:9092/bootstrap]: kafka:9092/bootstrap: Connecting to ipv4#10.150.12.67:9092 (plaintext) with socket 1028
7|2017-04-03 17:59:13.433|rdkafka#producer-1|STATE| [thrd:kafka:9092/bootstrap]: kafka:9092/bootstrap: Broker changed state INIT -> CONNECT
7|2017-04-03 17:59:13.434|rdkafka#producer-1|BROADCAST| [thrd:kafka:9092/bootstrap]: Broadcasting state change
log
log
log
log
log
log
log
...
```

treziac:
Fix #137 

Not really fan of this impl (volatile, other thread...) but it's the simplest I could find
- We could also save timeStart at beginning and compare with DateTime.Now in onLog/onError - no thread needed on begining, no volatile, not sure of cost
- I'm not sure of the ConcurrentQueue usage. May need a lock when iterating in OnLog{add}, to be sure no other message are added while registring the event.

This is more a WIP for feedback on it before consoliding (only on producer)

### DIFF
--- a/src/Confluent.Kafka/Producer.cs
+++ b/src/Confluent.Kafka/Producer.cs
@@ -25,6 +25,7 @@ using System.Runtime.InteropServices;
 using Confluent.Kafka.Impl;
 using Confluent.Kafka.Internal;
 using Confluent.Kafka.Serialization;
+using System.Collections.Concurrent;
 
 
 namespace Confluent.Kafka
@@ -54,7 +55,6 @@ namespace Confluent.Kafka
 
         private Task callbackTask;
         private CancellationTokenSource callbackCts;
-
         private const int POLL_TIMEOUT_MS = 100;
         private Task StartPollTask(CancellationToken ct)
             => Task.Factory.StartNew(() =>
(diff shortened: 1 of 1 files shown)


mhowlett:
I know I suggested this general approach previously, but I'm now thinking that just allowing a log handler to be specified in the constructor is probably better / more straightforward / more intuitive.

mhowlett:
thanks for your help!

in version 1.0, we moved specification of log / error event handlers to builder classes (a no compromise solution) so closing as this is no longer relevant.

Issue #220, opened 2017-06-08
System Null Reference Exception while using a simple consumer


chiru1205:
While testing out a consumer locally. I keep getting System.NullReferenceException

Exception: System.NullReferenceException Message: Object reference not set to an instance of an object. StackTrace: at Confluent.Kafka.Internal.Util.Marshal.PtrToStringUTF8(IntPtr strPtr) at Confluent.Kafka.Impl.SafeKafkaHandle.get_MemberId() at Confluent.Kafka.Consumer`2.get_MemberId() at Test.Monitor.<>c.<Start>b__28_4(Object obj, List`1 partitions) at Confluent.Kafka.Impl.LibRdKafka.NativeMethods.rd_kafka_destroy(IntPtr rk) at Confluent.Kafka.Impl.SafeKafkaHandle.ReleaseHandle() at System.Runtime.InteropServices.SafeHandle.InternalFinalize() at System.Runtime.InteropServices.SafeHandle.Dispose(Boolean disposing) at System.Runtime.InteropServices.SafeHandle.Finalize()

Error is occuring on OnParititionsRevoked even. At this point consumer.memberID is basically null.
![image](https://user-images.githubusercontent.com/16243876/26949519-e45b6d4a-4c5f-11e7-82d9-c252da6f38e0.png)

Not sure what  could be causing this. Here is my memory usage
![image](https://user-images.githubusercontent.com/16243876/26949560-06010392-4c60-11e7-9d83-f8d616f9e556.png)

Here is my consumer code at this moment

 _kafkaConsumer = new Consumer<Null, string>(consumerConfig, null,
                new StringDeserializer(System.Text.Encoding.UTF8));           

                _kafkaConsumer.OnMessage += (_, msg) =>
                {
                    lock (Lock)
                    {
                        LatestProcessedMessageOffset = msg.Offset;
                    }
                    ProcessMessage(msg);
                };

               _kafkaConsumer.OnError += (_, error) =>
                {
                                    Logger.Error($"Error [{error}], Consumer{_kafkaConsumer.MemberId} on Server {Environment.MachineName}");
                };

                _kafkaConsumer.OnPartitionsAssigned += (obj, partitions) =>
                {
                    var fromBeginning = partitions.Select(p => new TopicPartitionOffset(p.Topic, p.Partition,Offset.Beginning)).ToList();
                    _kafkaConsumer.Assign(fromBeginning);
[...]

PR #222, opened 2017-06-09
memberId: IntPtr == null -> IntPtr == IntPtr.Zero

mhowlett:
Fix for bug found in #220 @treziac @chiru1205


treziac:
Perhaps also add in documentation that it returns null when no groupid has been assigned? Given this, I'm not sure how it works - I get string.empty in my case rather than null with no connection to broker @edenhill

chiru1205:
Thank you so much for all your help.

mhowlett:
merging now so it gets in 0.9.6 with note to document in a separate issue. thanks for reviewing @treziac and thanks for the thorough testing @chiru1205 !

PR #230, opened 2017-06-16
Add AddBrokers function (#209)

### LINKED ISSUE #204: Request: Expose AddBrokers API

RammG:
* Add AddBrokers function

Close #204

* Remove empty line

* Test AddBroker return

* updated AddBrokers API documentation

* tweaking integration test

* review changes

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -620,6 +620,31 @@ namespace Confluent.Kafka
         /// </summary>
         public Metadata GetMetadata(bool allTopics)
             => consumer.GetMetadata(allTopics);
+
+        /// <summary>
+        ///     Adds one or more brokers to the Consumer's list of initial
+        ///     bootstrap brokers. 
+        ///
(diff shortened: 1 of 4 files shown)


mhowlett:
hmm... I merged 0.11.0.x onto master and github automatically connected that up to this PR for some reason.

PR #233, opened 2017-06-19
set title of strong name package

mhowlett:
nuget.org requires title to be unique.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,7 +11,7 @@ environment:
   PATH: $(PATH);$(PROGRAMFILES)\dotnet\
 
 build_script:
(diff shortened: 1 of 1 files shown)


mhowlett:
If the strong name package title is not different from the normal package, this can't be uploaded to nuget.org. @treziac or @edenhill 

I verified the title is correctly set in the generated artifact on appveyor.

treziac:
Strange, I though it was only the packageId which needed to be unique. Title should only be for UI
https://docs.microsoft.com/en-us/nuget/schema/nuspec
Whatever it's ok to change it (to be sure it does not take the title from non signed package)

mhowlett:
apparently it's an extra nuget.org constraint - wouldn't let me upload the package giving a messaging saying the duplicate name is likely to cause confusion.

PR #254, opened 2017-07-05
Handle null object case in equality operator overrides.

RyanMelenaNoesis:
Prevent NullReferenceException when doing == or != comparisons when Kafka object is null.


RyanMelenaNoesis:
[clabot:check]

mhowlett:
thanks @RyanMelenaNoesis - LGTM.

Issue #188, opened 2017-05-18
Consumer should implement an interface


michael-sherling:
I'm trying to make a unit test for my code that is using the Consumer class, but actually it is very difficult to do it because I can't mock it.
there are some workarounds, I thought creating some warping class and use it, but I think it is very important for the consumer to implement an interface (the producer is doing so)

mhowlett:
Thanks for the suggestion @michael-sherling - sounds like a great idea to me. 

Note that the interface `ISerializingProducer` implemented by `Producer` is actually for a subset of the functionality. We probably should add 4 new interfaces (for serializing and non-serializing variants of producer and consumer).

michael-sherling:
This will be very nice, thanks.
BTW, is this project opened for development outside of the confluent team only so I can create pull request also? if so I can implement this small change

treziac:
Yep it is. I'm a contributor outside confluent ^^

michael-sherling:
@treziac great! so basically I just clone the repo and committing to a new branch and creating new pull request?
is there some guide lines for this?

treziac:
That's it :)

No written guide lines for now. 
For interfaces, we will want it fully documented (the indentation in the project is not the default one so try to keep with what exists) Simply copy the one from the Producer class should be enough As often, just respect the global syntax in other files. Copying the one from producer should be enough.
Also feel free to add unit/integration tests (but I don't think there will be any necessary here)

Also (related to the issue and not the contribution) we will have to be sure this doesn't decrease perf (adding abstraction is not free) This should not be a problem usually as long as we don't use the interface internally (the CLR/JIT misses some optimizations here) I wonder if this could be the case in the Producer<K,V> by the way, will have to check this - this is negligible usually (perhaps <1ms), but as we plan to have a high-perf-library with hundred of thousands calls per second...

mhowlett:
yes - feel free to do a PR :-).

You might want to open a PR with just one interface first (maybe `IConsumer<TKey, TValue>`) which will be quicker to iterate on if we want any changes made.

regarding API docs, there is an `include` tag, and I was thinking now might be a good time to start using it: https://docs.microsoft.com/en-us/dotnet/articles/csharp/programming-guide/xmldoc/include 

I'm not worried about performance here - if there is a perf hit, it will be very small and it's a small price to pay for the benefit.

mhowlett:
[...]

PR #256, opened 2017-07-06
extracted IConsumer<,>

AndreasBieber:
Hi,

as in #188 discussed, extracted interface for Consumer<>.

Cheers
Andy

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -27,15 +27,22 @@ using Confluent.Kafka.Serialization;
 
 namespace Confluent.Kafka
 {
-    /// <inheritdoc cref="IConsumer{TKey,TValue}"/>
+    /// <summary>
+    ///     Implements a high-level Apache Kafka consumer (with 
+    ///     key and value deserialization).
+    /// </summary>
     public class Consumer<TKey, TValue> : IConsumer<TKey, TValue>
     {
         private readonly Consumer consumer;
 
-        /// <inheritdoc cref="IConsumer{TKey,TValue}.KeyDeserializer"/>
+        /// <summary>
+        ///     The IDeserializer implementation instance used to deserialize keys.
+        /// </summary>
         public IDeserializer<TKey> KeyDeserializer { get; }
 
-        /// <inheritdoc cref="IConsumer{TKey,TValue}.ValueDeserializer"/>
+        /// <summary>
+        ///     The IDeserializer implementation instance used to deserialize values.
+        /// </summary>
         public IDeserializer<TValue> ValueDeserializer { get; }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


AndreasBieber:
[clabot:check]

src/Confluent.Kafka/IConsumer.cs
```diff
@@ -0,0 +1,449 @@
...
+using System.Threading.Tasks;
+using Confluent.Kafka.Serialization;
+
+namespace Confluent.Kafka
+{
+    /// <summary>
+    ///     Implements a high-level Apache Kafka consumer (with 
+    ///     key and value deserialization).
+    /// </summary>
+    public interface IConsumer<TKey, TValue> : IDisposable
```
treziac: IConsumer has no IDisposable element in it, so IDisposable should stick on the implementation

src/Confluent.Kafka/Consumer.cs
```diff
@@ -27,22 +27,15 @@
 
 namespace Confluent.Kafka
 {
-    /// <summary>
-    ///     Implements a high-level Apache Kafka consumer (with 
-    ///     key and value deserialization).
-    /// </summary>
-    public class Consumer<TKey, TValue> : IDisposable
+    /// <inheritdoc cref="IConsumer{TKey,TValue}"/>
```
treziac: As far as I know, this project doesn't use sandcastle, and inheritdoc  is a sandcastle option.
Either wait for @mhowlett to see if sandcastle can be integrated, or use include and the indlude_docs.xml file. This would also allow to factorize some doc for Producer and non typed Consumer, so I would go for it

src/Confluent.Kafka/IConsumer.cs
AndreasBieber: `Consumer` implements `IDisposable` and `Consumer<TKey, TValue>` has a field `consumer`. Therefore I think `IConsumer<TKey,TValue>` should inherit from `IDisposable`. Please correct me if I`m wrong.

src/Confluent.Kafka/Consumer.cs
AndreasBieber: Sorry, my mistake. I'm using Rider for development instead of Visual Studio. And Rider (and also ReSharper) are able to interpret those `<inheritdoc>` tags. If it's ok for you, i'll remove comments in the implementation and keep them only in the interface.

src/Confluent.Kafka/IConsumer.cs
treziac: My point of view may be wrong, it's fine as you did

src/Confluent.Kafka/Consumer.cs
treziac: No, we need to keep the implementation documentation. All public code should be documented. Best thing to do for now is to use the indlude_docs.xml file

src/Confluent.Kafka/IConsumer.cs
mhowlett: yep, the way @AndreasBieber has it is what we want.

mhowlett:
There's a lot of API doc duplication here - should put API docs in this PR in`include_docs.xml`. See e.g. `IntSerializer`.

mhowlett:
yep, we also definitely want `IConsumer`, `IProducer` and `IProducer<K,V>` as @treziac says, though this can wait for another PR (I suspect `IConsumer<K,V>` will unblock most people.

I don't think we want an `IKafkaHandle` interface - `Producer` and `Consumer` are not handles, they use handles to achieve their job of being a producer or consumer. I very reluctantly kept the general handle methods on them, but they're marked as unstable and I would really like to factor this differently (though the departure in philosophy from librdkafka makes this a little tricky).

mhowlett:
in summary - thanks a lot for the contribution @AndreasBieber - I'm good to merge it if the docs are extracted out into `include_docs.xml`

mhowlett:
Update: I'm starting to think we should include all 4 interfaces simultaneously for consistency. Also the producer API is going to change (rename of some of the ProduceAsync methods + updates for custom timestamps) and therefore this change is blocked on that.

mhowlett:
I've implemented this (and `IProducer`) in a local branch an will be opening a PR for this in the near future. The PR will also include moving all the docs out to `include_docs_consumer.xml`.

PR #259, opened 2017-07-06
Incoherent EOL

treziac:
Mix of windows and linux end of lines, VS always complaining on it

We may want to add a .gitattribute to solve this


mhowlett:
I'll just merge this for now. Generally line endings shouldn't get muddled up - I think this probably happened when I edited a file checked out on an osx file system from a windows VM.

treziac:
Yep - we need a gittattribute to automate this

mhowlett:
feel free to do a PR :-) I didn't quickly see how to achieve what is needed. Note that I need to checkout with LF on linux/mac because I'm usually working in that environment. I think the problem occurred when I checked out with LF and then edited in Windows via a mounted disk. I think I just need to remember to not do that.

PR #261, opened 2017-07-07
Fix integration test on windows and/or distant cluster

### LINKED ISSUE #136: Integration test issues in VS2017

treziac:
Fix #136 
Fix integration tests on windows (path to file and bug on Console.Error)

Also fix integration tests when testing on distant cluster, where flush/poll need more time to complete properly

### DIFF
--- a/test/Confluent.Kafka.IntegrationTests/Tests/DeserializingConsumer_Poll_Error.cs
+++ b/test/Confluent.Kafka.IntegrationTests/Tests/DeserializingConsumer_Poll_Error.cs
@@ -43,8 +43,8 @@ namespace Confluent.Kafka.IntegrationTests
             using (var producer = new Producer(producerConfig))
             {
                 firstProduced = producer.ProduceAsync(topic, Encoding.UTF8.GetBytes("key"), null).Result.TopicPartitionOffset;
-                producer.ProduceAsync(topic, null, Encoding.UTF8.GetBytes("val"));
-                producer.Flush(30);
+                var secondProduced = producer.ProduceAsync(topic, null, Encoding.UTF8.GetBytes("val"));
+                producer.Flush(TimeSpan.FromSeconds(10));
             }
 
             var consumerConfig = new Dictionary<string, object>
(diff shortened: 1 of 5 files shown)


test/Confluent.Kafka.IntegrationTests/Tests/DeserializingConsumer_Poll_Error.cs
```diff
@@ -43,8 +43,8 @@ public static void DeserializingConsumer_Poll_Error(string bootstrapServers, str
             using (var producer = new Producer(producerConfig))
             {
                 firstProduced = producer.ProduceAsync(topic, Encoding.UTF8.GetBytes("key"), null).Result.TopicPartitionOffset;
-                producer.ProduceAsync(topic, null, Encoding.UTF8.GetBytes("val"));
```
mhowlett: `secondProduced` isn't used and furthermore it has a different type to `firstProduced`, so better not to make the change on this line IMO.

mhowlett:
LGTM after above change made - thanks!

test/Confluent.Kafka.IntegrationTests/Tests/DeserializingConsumer_Poll_Error.cs
treziac: Don't know why I added it, thanks

PR #263, opened 2017-07-12
WIP: Fix Accessviolation when handler creation fails

### LINKED ISSUE #125: AccessViolationException with SSL configuration (consumer)
I'm running Confluent.Kafka client (0.9.4) with .Net framework 4.5.1. While it works fine without SSL, consumer client throws **AccessViolationException** with SSL configuration: 

```
{
            {"group.id", kafkaGroup}, 
            {"client.id", clientId}, 
            {"bootstrap.servers", kafkaServer}, 
            {"enable.auto.commit", true},
            {"security.protocol", "ssl" },
            {"ssl.key.location", "sslkey/client.keystore.jks"},
            { "ssl.key.password", "testtest" },
            { "auto.commit.interval.ms", 5000},
            { "default.topic.config", new Dictionary<string, object>()
                {
                    { "auto.offset.reset", "smallest" }
                }
            }

        }
```

treziac:
Fix #125 
Not sure how to test given ssl configuration, but I can with providing a null socket_cb for example to test when rd_kafka_create fail.

### DIFF
--- a/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
+++ b/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
@@ -95,7 +95,6 @@ namespace Confluent.Kafka.Impl
                     (UIntPtr) errorStringBuilder.Capacity);
             if (skh.IsInvalid)
             {
-                LibRdKafka.conf_destroy(skh.DangerousGetHandle());
                 throw new InvalidOperationException(errorStringBuilder.ToString());
             }
             return skh;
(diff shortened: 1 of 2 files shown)


edenhill:
The caller should actually free the conf_t if rd_kafka_new() fails.
This has been a bit buggy in previous releases of librdkafka but 0.11.0 should get it right.

treziac:
Thanks! still a bug to fix in this case (replace skh/GetDangerousHandle by config, it's config which should be closed
I couldn't reprouce as rd_kafka_new always succeed, will complete it tomorrow

mhowlett:
Thanks @treziac - good catch. Will merge after you make that change.

edenhill:
@treziac You can probably specify an invalid `ssl.ca.location` or so.

treziac:
(tested on both 0.9.5 and 0.11.0 RC)

mhowlett:
☀️🍉LGTM

PR #292, opened 2017-08-12
Consumer Pause/Resume methods

mhowlett:
the go client just got these, and we can't let it win... straightforward addition so I just did it.

@treziac or @edenhill

### DIFF
--- a/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
+++ b/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
@@ -565,6 +565,50 @@ namespace Confluent.Kafka.Impl
         internal Task<CommittedOffsets> CommitAsync(IEnumerable<TopicPartitionOffset> offsets)
             => Task.Run(() => commitSync(offsets));
 
+        internal List<TopicPartitionError> Pause(IEnumerable<TopicPartition> partitions)
+        {
+            ThrowIfHandleClosed();
+            IntPtr list = LibRdKafka.topic_partition_list_new((IntPtr) partitions.Count());
+            if (list == IntPtr.Zero)
+            {
+                throw new Exception("Failed to create pause partition list");
+            }
+            foreach (var partition in partitions)
+            {
+                LibRdKafka.topic_partition_list_add(list, partition.Topic, partition.Partition);
+            }
+            ErrorCode err = LibRdKafka.pause_partitions(handle, list);
+            var result = GetTopicPartitionErrorList(list);
+            LibRdKafka.topic_partition_list_destroy(list);
+            if (err != ErrorCode.NoError)
+            {
+                throw new KafkaException(err);
+            }
+            return result;
+        }
+
+        internal List<TopicPartitionError> Resume(IEnumerable<TopicPartition> partitions)
+        {
+            ThrowIfHandleClosed();
+            IntPtr list = LibRdKafka.topic_partition_list_new((IntPtr) partitions.Count());
+            if (list == IntPtr.Zero)
+            {
+                throw new Exception("Failed to create resume partition list");
+            }
+            foreach (var partition in partitions)
+            {
+                LibRdKafka.topic_partition_list_add(list, partition.Topic, partition.Partition);
+            }
+            ErrorCode err = LibRdKafka.resume_partitions(handle, list);
+            var result = GetTopicPartitionErrorList(list);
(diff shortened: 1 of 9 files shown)


src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -565,6 +565,50 @@ internal Task<CommittedOffsets> CommitAsync()
         internal Task<CommittedOffsets> CommitAsync(IEnumerable<TopicPartitionOffset> offsets)
             => Task.Run(() => commitSync(offsets));
 
+        internal List<TopicPartitionError> Pause(IEnumerable<TopicPartition> partitions)
+        {
+            ThrowIfHandleClosed();
```
mhowlett: it's about time the boiler plate stuff here got factored out, but this applies to other methods too, so I think that can be an exercise for a later pr.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Pause_Resume.cs
```diff
@@ -0,0 +1,75 @@
...
+                    assignedPartitions = partitions;
+                };
+
+                consumer.Subscribe(topic);
+
+                while (assignedPartitions == null) 
+                {
+                    consumer.Poll(TimeSpan.FromSeconds(1));
+                }
+                Assert.False(consumer.Consume(out message, TimeSpan.FromSeconds(1)));
```
mhowlett: Calling consume (or poll) here is necessary in order to make the consume on line 65 work (why?). In related news, 9 of our integration tests are failing now (the ones I checked are due to a call to Consume). Tested on MacOS against Confluent OSS 3.3.0 and 3.2.2 brokers. anything spring to mind @edenhill ?

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -565,6 +565,50 @@ internal Task<CommittedOffsets> CommitAsync()
...
+            IntPtr list = LibRdKafka.topic_partition_list_new((IntPtr) partitions.Count());
```
edenhill: The IEnumerable<TopicPartition*> to partition_list_t mangling (and reverse) is pretty common, might want to refactor into a helper method to avoid code duplication.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Pause_Resume.cs
edenhill: Is line 65 returning without a message before the 30 second timeout?
Is the topic unique? Does it exist?
Run with debug=cgrp,fetch,topic  to see what is going on.

test/Confluent.Kafka.UnitTests/TopicPartitionError.cs
```diff
@@ -0,0 +1,100 @@
...
+    public class TopicPartitionErrorTests
+    {
+        [Fact]
+        public void Constuctor()
+        {
+            var tpe = new TopicPartitionError("mytopic", 42, ErrorCode.Local_BadMsg);
+
+            Assert.Equal(tpe.Topic, "mytopic");
+            Assert.Equal(tpe.Partition, 42);
+            Assert.Equal(tpe.Error, new Error(ErrorCode.Local_BadMsg));
```
edenhill: Might want to add a negative test here, tpe.Error is not equal to some other error code.

test/Confluent.Kafka.UnitTests/TopicPartitionError.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        [Fact]
+        public void Equality()
+        {
+            var a = new TopicPartitionError("a", 31, ErrorCode.NoError);
+            var a2 = new TopicPartitionError("a", 31, ErrorCode.NoError);
+            var nes = new List<TopicPartitionError> {
```
edenhill: Took me a while to decipher `nes`, maybe spell it out `notEquals`

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
mhowlett: yep, I note that above as well. I decided a separate PR for that is appropriate.

PR #295, opened 2017-08-17
Removed unstable designation from Consume method

mhowlett:
The `Consume` method has now proven it's self useful in a number of scenarios in which the `Poll` method is inconvenient. we should definitely keep both.

@edenhill


mhowlett:
uses: 1. testing - many of the integration tests are awkward without this and as Poll is a lightweight wrapper around Consume, coverage is pretty much good for Poll too 2. basis of iterator APIs.

PR #313, opened 2017-08-30
WIP : Delete topics

treziac:
Binding to https://github.com/edenhill/librdkafka/pull/1399

Need to include proper librdkafka dll for it to work

## Admin interface / client
This is more for discussion on how to implement AdminClient : either create a completly different handle (like here in the AdminClient, in which case we will have to deprecate all admin method from producer/consumer to put them in), either only on existing handle (like here on the Producer), either on both, either AdminClient.CreateFrom(producer | consumer). I like it to be able to use admin api from existing handle, so I'm just again the "new AdminClient only"

I made the producer explicitly implement IAdmin more for testing (prevent producer.DeleteTopics usage), not necesseraly a good idea. Idealy we should add the IProducer, ISerializingProducer api, so that we can distinguish producing method from admin method easily (having a IProducer producer = new Producer())

I let some existing common method in AdminClient, will have to decide which one make it to IAdmin api and which one don't.

## Delete topics
The way to treat timeout is more something to discuss on librdkafka PR first than here, but the way exception have to be treated in C# can be reviewed also

### DIFF
--- a/src/Confluent.Kafka/AdminClient.cs
+++ b/src/Confluent.Kafka/AdminClient.cs
@@ -1,0 +1,374 @@
+// Copyright 2016-2017 Confluent Inc., 2015-2016 Andreas Heider
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Refer to LICENSE for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Linq;
+using Confluent.Kafka.Impl;
+using Confluent.Kafka.Internal;
+
+
+namespace Confluent.Kafka
+{
+    /// <summary>
+    ///     Implements a high-level Apache Kafka producer (without serialization).
+    /// </summary>
+    public class AdminClient : IDisposable, IAdmin
+    {
+        private bool manualPoll;
+        private bool disableDeliveryReports;
+        
+        private SafeKafkaHandle kafkaHandle;
+
+        private Task callbackTask;
+        private CancellationTokenSource callbackCts;
+
+        private const int POLL_TIMEOUT_MS = 100;
+        private Task StartPollTask(CancellationToken ct)
+            => Task.Factory.StartNew(() =>
+                {
+                    while (!ct.IsCancellationRequested)
+                    {
+                        kafkaHandle.Poll((IntPtr)POLL_TIMEOUT_MS);
+                    }
+                }, ct, TaskCreationOptions.LongRunning, TaskScheduler.Default);
+
+        private LibRdKafka.ErrorDelegate errorDelegate;
+        private void ErrorCallback(IntPtr rk, ErrorCode err, string reason, IntPtr opaque)
+        {
+            OnError?.Invoke(this, new Error(err, reason));
+        }
+
+        private LibRdKafka.StatsDelegate statsDelegate;
+        private int StatsCallback(IntPtr rk, IntPtr json, UIntPtr json_len, IntPtr opaque)
+        {
+            OnStatistics?.Invoke(this, Util.Marshal.PtrToStringUTF8(json));
+            return 0; // instruct librdkafka to immediately free the json ptr.
+        }
+
+        private LibRdKafka.LogDelegate logDelegate;
+        private void LogCallback(IntPtr rk, int level, string fac, string buf)
+        {
+            var name = Util.Marshal.PtrToStringUTF8(LibRdKafka.name(rk));
+
+            if (OnLog == null)
+            {
+                // Log to stderr by default if no logger is specified.
+                Loggers.ConsoleLogger(this, new LogMessage(name, level, fac, buf));
+                return;
+            }
+
+            OnLog.Invoke(this, new LogMessage(name, level, fac, buf));
+        }
+
+        /// <summary>
+        ///     Initializes a new Producer instance.
+        /// </summary>
+        /// <param name="config">
+        ///     librdkafka configuration parameters (refer to https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md).
+        ///     Topic configuration parameters are specified via the "default.topic.config" sub-dictionary config parameter.
+        /// </param>
+        /// <param name="manualPoll">
+        ///     If true, does not start a dedicated polling thread to trigger events or receive delivery reports -
+        ///     you must call the Poll method periodically instead. Typically you should set this parameter to false.
+        /// </param>
+        public AdminClient(IEnumerable<KeyValuePair<string, object>> config, bool manualPoll)
+        {
+            this.manualPoll = manualPoll;
+
+            var configHandle = SafeConfigHandle.Create();
+            config
+                .Where(prop => prop.Key != "default.topic.config")
+                .ToList()
+                .ForEach((kvp) => { configHandle.Set(kvp.Key, kvp.Value.ToString()); });
+
+            IntPtr configPtr = configHandle.DangerousGetHandle();
+            
+            // Explicitly keep references to delegates so they are not reclaimed by the GC.
+            errorDelegate = ErrorCallback;
+            logDelegate = LogCallback;
+            statsDelegate = StatsCallback;
+
+            // TODO: provide some mechanism whereby calls to the error and log callbacks are cached until
+            //       such time as event handlers have had a chance to be registered.
+            LibRdKafka.conf_set_error_cb(configPtr, errorDelegate);
+            LibRdKafka.conf_set_log_cb(configPtr, logDelegate);
+            LibRdKafka.conf_set_stats_cb(configPtr, statsDelegate);
+
+            this.kafkaHandle = SafeKafkaHandle.Create(RdKafkaType.Producer, configPtr);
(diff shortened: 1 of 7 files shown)


src/Confluent.Kafka/IAdmin.cs
```diff
@@ -0,0 +1,68 @@
...
+using System;
+using System.Collections.Generic;
+
+
+namespace Confluent.Kafka
+{
+    /// <summary>
+    ///     Expose administrative commands to interact with broker state and configuration.
+    /// </summary>
+    public interface IAdmin
```
mhowlett: `IAdminClient` would be more consistent.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -805,5 +820,54 @@ private List<GroupInfo> ListGroupsImpl(string group, int millisecondsTimeout)
                 throw new KafkaException(err);
             }
         }
+
+        internal List<TopicError> AdminDeleteTopics(IEnumerable<string> topics, int millisecondsTimeout, bool waitForDeletion)
+        {
+            if(topics == null)
```
mhowlett: space after `if` (also below)

src/Confluent.Kafka/Producer.cs
```diff
@@ -32,7 +32,7 @@ namespace Confluent.Kafka
     /// <summary>
     ///     Implements a high-level Apache Kafka producer (without serialization).
     /// </summary>
-    public class Producer : IDisposable
+    public class Producer : IDisposable, IAdmin
```
mhowlett: I don't think we should expose admin operations on `Producer`, only on `AdminClient`.

mhowlett: Oh, explicit implementation. Probably better. Still, I was thinking of having a `Handle` property and allowing AdminClient to be constructed from this. Have not completely thought through all the details though.

src/Confluent.Kafka/AdminClient.cs
```diff
@@ -0,0 +1,374 @@
...
+        /// <summary>
+        ///     Raised on critical errors, e.g. connection failures or all 
+        ///     brokers down. Note that the client will try to automatically 
+        ///     recover from errors - these errors should be seen as 
+        ///     informational rather than catastrophic
+        /// </summary>
+        /// <remarks>
+        ///     Called on the Producer poll thread.
+        /// </remarks>
+        public event EventHandler<Error> OnError;
```
mhowlett: Ahh yes, I guess if we have another client type we're going to need all this stuff again. Perhaps we should have a base class `KafkaClient`. Maybe it should be`internal`.

treziac: Yep. I tried to add one once, but it's not straight forward. Will be a nice refacto though :)

src/Confluent.Kafka/Producer.cs
treziac: As I say, it's a discussion ;)
What I would like is to be able to use admin method from existing handler - an IAdminClient AdminHelper.GetAdmin(Producer producer) will work fine for this -and this is where explicit implementation is nice, you just have to return producer as IAdmin.
Here, we can't write producer.DeleteTopics - producer must be cast to IAdmin first

Other reason is that we will have to list what is admin api and what is not (is ListGroup/QueryWatermarkOffsets/etc an admin API? should we remove it / depreciate it from existing Consumer/Producer?)

src/Confluent.Kafka/AdminClient.cs
```diff
@@ -0,0 +1,374 @@
...
+
+        /// <summary>
+        ///     Get information pertaining to all groups in the Kafka cluster (blocking)
+        ///
+        ///     [UNSTABLE-API] - The API associated with this functionality is subject to change.
+        /// </summary>
+        /// <param name="timeout">
+        ///     The maximum period of time the call may block.
+        /// </param>
+        public List<GroupInfo> ListGroups(TimeSpan timeout)
```
mhowlett: The reason I marked this as unstable originally on Producer and Consumer is I really want the client types to be very focussed on their purpose. Does this method make sense on a thing called producer? No IMO. 

If we start exposing a new `Handle` type on clients, and allow other clients to be constructed from this (and note: clients use the handle, not are a handle), we can get both a clean API and avoid the need for unnecessary creation of connections to brokers. This is not possible with Producers and Consumers, because we need to specify the type at construction time, but it is possible to construct and AdminClient in this way.

But should this method be on AdminClient? Actually no as well - AdminClient should probably be focussed on what is provided by the Kafka Administrative API.

Should this method be on Consumer? Probably not IMO as well - the group protocol is intended as a general thing (not just for consumer groups), and our implementation does not yet provide anything specific to it's use for consumer groups. If we started interpreting the MemberMetadata and MemberAssignment byte arrays as consumer group data, we should definitely put that functionality on the Consumer though.

Perhaps ListGroups and GetMetadata should go on this new publicly exposed Handle class. Maybe all things that don't belong on Producer, Consumer or AdminClient make sense on Handle.

Actually, very much liking that idea ATM.

Handle would be very much like SafeKafkaHandle, except it would only expose these selected methods. It would probably wrap SafeKafkaHandle (which would be an internal property).

treziac: What seems important from me (from a user perspective) is to have interfaces IProducer, IConsumer, IHandle, IAdmin. IProducer and IConsumer are linked to librdkafka reserved method for given type, but everything else is accessible from the Handle, which would implement IHandle and IAdmin (Handle being internal)

One difficulty is how to decide if some api are from IHandle or from IAdmin (ie, I don't see why 
describeTopics woul be an admin api, and getMetadata an handle one...) We can just stick to java, it's an open question :)

I'm ok with the Handle class. Just want to be able to access IAdmin from Producer (either Producer implement IAdmin, but I understand you don't want that, either, have an IAdminClient AdminHelper.CreateFrom(Producer), or a AdminClient(Producer), but AdminClient should not be disposable in this case, so i don't like it - as the handle implements IAdmin, it will be no problem)

mhowlett: disposabiltiy is a good point that i didn't think of above.

treziac: Other solution for consumer / producer: only expose producer / consumer api (Producer only expose consume), and have an Handle / Admin property which allow access on IHandle / IAdmin api (producer.Handle.GetMetadata, producer.Admin.CreateTopics) I don't know if I like this (great separation of api) or dislike it (harder to call basic method on producer/consumer)

mhowlett: Yeah, that idea has come up before and been discarded.

Another point is the Dispose method doesn't necessarily need to dispose the SafeKafkaHandle. If an AdminClient was constructed using a Handle, the Dispose method could do nothing. Alternatively we could reference count and dispose when the last reference is disposed. 

(just some quick brainstorming - no clear opinion yet).

mhowlett: I'm going off the idea of having any public methods on Handle (in part because of the usage you point out above). Maybe we have just Producer, Consumer, and AdminClient (and corresponding interfaces as you suggest).

Maybe typical usage is like:

```
using (var adminClient = new AdminClient(config))
{
    ...
}
```

or 

```
using (var producer = new Producer<string, string>(config, new StringSerializer(Encoding.UTF8), new StringSerializer(Encoding.UTF8)))
using (var adminClient = new AdminClient(producer.Handle))
{
    ...
}
```

but disposal of adminClient in the latter case actually does nothing. And maybe it's ok that if the producer gets destroyed, adminClient can no longer be used.

mhowlett: the below is just brainstorming - thinking out load. thought it might help to write this down, but i don't think it's going anywhere.

It'd be kind of cool if this were possible:

```
using (var client = new KafkaClient(config))
{
    var producer = (IProducer<string, string>)client;
    var adminClient = (IAdminClient)client;
    var consumer = (IConsumer)client;
}
```

where explicit interface implementations are used (as you are suggesting). 

of course we can't convert between Producer and Consumer. Perhaps:

```
using (var client = new KafkaClient(config, ClientType.Producer))
{
    var producer = (IProducer<string, string>)client;
    var adminClient = (IAdminClient)client;
    var consumer = (IConsumer)client; // disallowed
}
```

and/or

```
using (var producer = KafkaClient.CreateProducer<string, string>(config, new StringSerializer(Encoding.UTF8), new StringSerializer(Encoding.UTF8)))
{
    var adminClient = (IAdminClient)producer; 
}
```

where CreateProducer returns IProducer<string,string> and is an instance of`KafkaClient` (type Producer).

This is quite a departure from where we are currently of course. I'm not sure it's better.

The Handle idea also isn't ideal, but i think it's fairly inoffensive.

With java, you can't reuse connections for the different client types. We could also do that (but it'd be a pity).

mhowlett: Maybe we:

1. Have a new client type `AdminClient`. We don't allow this to be constructed in a way which reuses a librdkafka handle. It's a new client. Same as in Java. Same as what you have done in this PR (good job!)

2. Additionally, allow an explicit cast of `Producer` or `Consumer` to `IAdminClient` as kind of a backdoor way of re-using these connections for admin client tasks. Same as what you have in this PR (good job!). People have dissuaded me from using explicit casts here in the past, but i'm thinking it's a good idea. It's kind of an advanced usage.

3. Make `GetMetadata`, `DeleteTopics`, `ListGroups` only be available via the IAdminClient interface to make the `Producer` and `Consumer` APIs nice and clean.

4. Figure out some nice factoring to reduce code duplication (though not necessarily in this PR).

src/Confluent.Kafka/Producer.cs
mhowlett: Yep - we make a call on what belongs where typical need and remove / depreciate from Consumer/Producer as appropriate. All methods that may be affected are currently marked UNSTABLE, so it's no problem.

mhowlett:
thanks for your efforts here @treziac. admin client (and delete topics) has now been finalized in another commit.

PR #319, opened 2017-09-12
[WIP] Schema Registry / Avro support

mhowlett:
This is pretty much just a reorganization of @treziac 's PoC.
I plan to iterate on this tomorrow and provide some comments. 

Note: Initially Schema Registry / Avro support will be .NET Framework only (no .NET Core). 
It will be distributed as a separate nuget package.

@treziac - I'm v. keen for your name to be against all these lines of code (recognition where it's due!). To reorganise though, I did a new branch off master and copied the relevant files in from your branch. I did amend the commit to have your name as the author but not sure if this is good enough (it shows us both as authors).


src/Confluent.Kafka.SchemaRegistry/Serialization/AvroSerializer.cs
```diff
@@ -0,0 +1,150 @@
...
+        public byte[] Serialize(string topic, object data)
+        {
+            string subject = SchemaRegistryClient.GetRegistrySubject(topic, IsKey);
+            // TODO check to use something else than 30 which is not optimal.
+            // For primitive type, we can "easily" generate an exact value
+            using (var stream = new MemoryStream(30))
+            {
+                int schemaId;
+                Avro.Schema schema = GetSchema(data);
+                schemaId = SchemaRegistryClient.RegisterAsync(subject, schema.ToString()).Result;
```
mhowlett: We probably should have made the `Serialize` method on `ISerializable` return a `Task` and made this method `async`.

Related: if this were the case, some of the incorrectly named `ProduceAsync` methods would then actually be correctly named (the ones on the serializing producer).

src/Confluent.Kafka.SchemaRegistry/Serialization/AvroSerializer.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// </summary>
+        /// <param name="data"></param>
+        /// <param name="topic"></param>
+        /// <param name="length">Length of the result to take into account </param>
+        /// <param name="isKey"></param>
+        /// <returns></returns>
+        public byte[] Serialize(string topic, object data)
+        {
+            string subject = SchemaRegistryClient.GetRegistrySubject(topic, IsKey);
+            // TODO check to use something else than 30 which is not optimal.
```
mhowlett: This could even be exposed as a property.

mhowlett:
Following a discussion with @mageshn, I think we should focus on implementing specific (typed) avro serializers / deserializers first - generic seems like it would be more uncommonly desired and much more uncommonly required. Feedback on that welcome.

src/Confluent.Kafka.SchemaRegistry/Serialization/AvroSerializer.cs
mhowlett: In practice this will be non-blocking the vast majority of the time, so I'm thinking we should just leave things as is. To justify changing the interface, I think we'd want to be aware of some scenarios where this would actually really matter. I can't think of any currently.

Opinions welcome.

examples/AvroGeneric/Program.cs
```diff
@@ -0,0 +1,117 @@
...
+            schemaRegistry.Setup(x => x.GetSchemaAsync(2)).ReturnsAsync(User._SCHEMA.ToString());
+
+            var producerConfig = new Dictionary<string, object> { { "bootstrap.servers", brokerList } };
+            var consumerConfig = new Dictionary<string, object> { { "bootstrap.servers", brokerList }, { "group.id", Guid.NewGuid() } };
+            var consumerDynamicConfig = new Dictionary<string, object> { { "bootstrap.servers", brokerList }, { "group.id", Guid.NewGuid() } };
+
+            var deserializer = new AvroDeserializer(schemaRegistry.Object);
+
+            using (var consumer = new Consumer<object, object>(consumerConfig, deserializer, deserializer))
+            using (var consumerDynamic = new Consumer<dynamic, dynamic>(consumerDynamicConfig, deserializer, deserializer))
```
mhowlett: I don't understand how dynamic helps here? The class delivered by the consumer is always the same right? (unless you were thinking of extending things to make some fancy anonymous class instances, but I don't think that's a great idea).

src/Confluent.Kafka.SchemaRegistry/Serialization/AvroSerializer.cs
treziac: If we make the serializer non async, the Produce order may not be fullfilled. I don't like mixing async and non async though, or blocking on Produce, so no good opinion on this, other than "you should call register before making call to produce". On this end we could also only expose a "RetrieveId" on schemaRegistry which would retrieve the id synchronously from cache and throw if Register was not call before, but this is not user friendly...

examples/AvroGeneric/Program.cs
treziac: The way on avro side for generics (well, it returns an object, so dynamic seems ok). I don't really like dynamic, but it seemed ok here.

comdw:
Any idea when this is likely to make it into a Confluent.Kafka official release?  I have pulled the code in this PR and it worked well for me with Kafka, Schema Registry and Connect.

edenhill:
I would guess early Q1

mhowlett:
closing in favor of #351

PR #324, opened 2017-09-19
[WIP] default.topic.config => top level

### LINKED ISSUE #322: How to set message.timeout.ms for Producer
I want to get error faster than after 5min if brokers are down. I found that options message.timeout.ms is for it. But when I set this options when I creating Producer instance and call ProduceAsync I get result with error after 5min.

I create producer like this
var config = new KeyValuePair<string, string>[]{ 
  new KeyValuePair<string, string>("message.timeout.ms", "1000"),
  new KeyValuePair<string, string>("bootstrap.servers", "server:9092"),
}
_producer = new Producer(config, false, false);

Thanks

Checklist
=========

 - [ ] Confluent.Kafka nuget version: 0.11.0

mhowlett:
quick change to resolve #322. WIP, but people can feel free to comment if they want. I need to think through some detail still and need to write some tests. Might be good to review what we want to do with the `disableDeliveryReports` producer constructor parameter in this PR as well.

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -756,12 +756,21 @@ namespace Confluent.Kafka
                 throw new ArgumentException("'group.id' configuration parameter is required and was not specified.");
             }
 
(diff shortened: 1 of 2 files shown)


mhowlett:
closing, this functionality was included in #410.

PR #338, opened 2017-10-13
Add args length check

mitchstewart:


mitchstewart:
[clabot:check]

mhowlett:
yep, would love to print out usage here (anything to streamline new user experience is high priority). what's stopping me from adding this in the past is i wasn't sure what the most proper thing would be for the usage string to say. to run this on my mac, I would generally do:

`dotnet run localhost:9092 topicname`

or I could do:

`dotnet SimpleProducer.dll localhost:9092 topicname`

(not `SimpleProducer localhost:9092 topicname`)

what do other projects do?

what you have in your PR is pretty self explanatory and definitely a step forward, and it might be the best thing we can do, but i'm not sure of that.

mhowlett:
ok, LGTM.

Issue #31, opened 2017-02-01
Auto Offset Commit | add offset_store() support


cgosiak:
I am not sure if this is a bug; but figured that it was not expected behavior. I am running an advanced consumer with the configuration:
```
new Dictionary<string, object>
            {
                {"group.id", _consumerGroup},
                {"enable.auto.commit", true},
                {"auto.commit.interval.ms", 5000},
                {"statistics.interval.ms", 60000},
                {"bootstrap.servers", brokerList},
                {
                    "default.topic.config", new Dictionary<string, object>()
                    {
                        {"auto.offset.reset", "largest"}
                    }
                }
            };
```
---
## Scenario 1)
- Advanced Producer Writing to Topic/Advanced Consumer Consuming
  - Offsets are committed every 5 seconds; as long as a new message comes in.
  - This is the expected behavior.

## Scenario 2)
- Advanced Producer is Turned Off/Topic Has Unconsumed Messages/Then Consumer is Turned On
  - Offsets are not committed until every message is received.
  - I do not believe this is the expected behavior.
---
I believe that offsets should be auto committed with the interval; regardless of the state of the producer.

edenhill:
Yes, your assumption is correct.

To clarify scenario 2:
your consumer starts reading from a previously committed offset, reads at least 1 message to catch up to the partition end, but even after 5000ms it hasn't committed that final offset?

cgosiak:
I have a topic and a consumer that has consumed all of the messages (the consumer is caught up)

I then close the consumer and produce some messages to the topic.  The consumer group is now behind.

I then open the consumer; it starts from where it last ended. However, none of the offsets are committed until each and every message from the topic is read. The commit happens after every message is received by the consumer.

edenhill:
And it takes longer than 5s for the consumer to catch up to the last message?

cgosiak:
Yes; to get through it takes upwards of a minute.

edenhill:
Is that due to message processing in your application or network/IO speed?

cgosiak:
The message processing in my application; I am just doing some testing when I noticed it. I could send a code snippet if that would help?

cgosiak:
I am running a modified version of the "Run_Poll" test case:
```
public static void Run_Poll(string brokerList, List<string> topics)
        {
            using (
                var consumer = new Consumer<Null, string>(constructConfig(brokerList, true), null,
                    new StringDeserializer(Encoding.UTF8)))
            {
[...]

PR #339, opened 2017-10-13
Add Consumer.Storeoffsets

vinodres:
Opening this PR for exposing Storeoffsets feature in .net api. Requested in #31.

As suggested I followed the implementation of this feature from Go language https://github.com/confluentinc/confluent-kafka-go/commit/e093d2be7dd0f3fc41f3a53c5127a139a9e4919d 

Regarding what to retrun from Consumer.cs StoreOffsets method, I was debating if I should use existing class called CommitedOffsets.cs to capture results, but the name of the class appears to be misleading. So I have created a new class called OffsetResults.cs, a more general name to capture such results. Please let me know your thoughts and suggestions. If this new name appears appropriate, then it can be used instead of CommitedOffsets.cs as well. I didn't go that far because it would be a breaking change.


vinodres:
[clabot:check]

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -510,6 +510,29 @@ internal void Assign(IEnumerable<TopicPartitionOffset> partitions)
         }
 
         /// <summary>
+        /// Store offsets to be comitted.
```
mhowlett: format comments here in the style of other methods.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -510,6 +510,29 @@ internal void Assign(IEnumerable<TopicPartitionOffset> partitions)
...
+            {
+                return new OffsetResults(new Error(err));
+            }
+
+            OffsetResults offsetResults = new OffsetResults(result);
+            return offsetResults;
+        }
+
+
+        /// <summary>
```
mhowlett: oh, we don't either. not a big deal.

mhowlett:
thanks! this looks good at first glance and is pretty small - should be able to get this in soon.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -414,6 +414,25 @@ public void Unassign()
...
 
         /// <summary>
+        ///     Offset is stored for commit based on the topic/partition/offset of a message.
+        ///     The next message to be read will be that following <paramref name="message" />.
+        /// </summary>
+        /// <param name="message">
+        ///     The message used to determine the stored offset.
+        /// </param>
+        /// <remarks>
+        ///     A consumer which has position N has consumed records with offsets 0
```
edenhill: This is incorrect, the consumer can start consuming at any offset.
I think this remark needs a some tweaking, it reads a bit hard.

Also make sure to pick up the information from here:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L2360

src/Confluent.Kafka/Impl/LibRdKafka.cs
```diff
@@ -160,6 +160,7 @@ static LibRdKafka()
             _consumer_close = NativeMethods.rd_kafka_consumer_close;
             _assign = NativeMethods.rd_kafka_assign;
             _assignment = NativeMethods.rd_kafka_assignment;
+            _offsets_store = NativeMethods.rd_kafka_offsets_store;
```
edenhill: This API was added in librdkafka 0.11.0, if we want to be compatible with older librdkafka versions we might want to conditionalize this, yesno?

src/Confluent.Kafka/Consumer.cs
vinodres: I will update the docs

src/Confluent.Kafka/Impl/LibRdKafka.cs
vinodres: Yes, I agree on your suggestion. Let me think through it.

vinodres: @edenhill :  I understood the part that this feature is only available from version 0.11.0. 

As I am not familiar with the direction on how LibRdKafka version dependency is managed. Usually any new features such as store offsets should be merged into a future version of Confluent.Kafka and lift the minVersion of LibRdKafka to 0.11.0. I mean this feature will be merged to master whenever it is safe to lift minVersion of LibRdKafka to 0.11.0. There must be something I am missing when you asked for conditionalize code.

When you said conditionalize this (_offsets_store = NativeMethods.rd_kafka_offsets_store;), I didn't fully understand your exact intention. Are you asking me to add a pre validation inside SafeKafkaHandle.cs StoreOffsets method to check if LibRdKafka version is less than 0.11.0 and throw something like NotSupportedException with a meaningful message?

edenhill: All valid questions @vinodres, I'll leave @mhowlett to answer them :)

treziac: For now, this could be via Librdkafka.cs (`const long minVersion = 0x00110000`), no code is doing checking (we surely should)

We could provide some more fined process for this. Either check at each call what is the version loaded (and throwing exception when version is not ok as @vinodres  stated), either do this via attibute `[MinVersion(0x00110000)]` (cleaner, but runtime cost would not be negligeable)

I think this will need some thinking in relation with #340, which will allow to more easily load any version of librdkafka

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -510,6 +510,39 @@ internal void Assign(IEnumerable<TopicPartitionOffset> partitions)
...
+        {
+            ThrowIfHandleClosed();
+            IntPtr cOffsets = GetCTopicPartitionList(offsets);
+            ErrorCode err = LibRdKafka.offsets_store(handle, cOffsets);
+            var result = GetTopicPartitionOffsetErrorList(cOffsets);
+            LibRdKafka.topic_partition_list_destroy(cOffsets);
+
+            if (err != ErrorCode.NoError)
+            {
+                return new OffsetResults(new Error(err));
```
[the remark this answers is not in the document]
treziac: @edenhill  after looking at librdkafka code, it seems there is an incoherence between doc and implem

rdkafka.h
```
 * @returns RD_KAFKA_RESP_ERR_NO_ERROR on success or an error code if
 *          none of the offsets could be stored.
```

rdkafka_offset.c
```
       return offsets->cnt > 0 && ok_cnt < offsets->cnt ?
              RD_KAFKA_RESP_ERR__UNKNOWN_PARTITION :
              RD_KAFKA_RESP_ERR_NO_ERROR;
```

We return unkown_partition if any offset could not be set because of unknown partition, even if other success.
I also wonder why there is no other possible error (transport or other)

On my previous comment in this case, this is more troublesome - we don't want to throw exception when it is not a complete failure, so either let it like this for now, either do some check on result (is there any correct offset stored?)

Also, this means the impl is not ok: in case a partition was not correctly set, we still want to create `offsetResults` from `result`, with corresponding error (so perhaps keeping a TopicPartitionOffsetError), to have both ok result and nok results.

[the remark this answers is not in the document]
edenhill: I'd say this is a bug in librdkafka. That should probably read `offsets->cnt > 0 && ok_cnt == 0`

src/Confluent.Kafka/Impl/LibRdKafka.cs
mhowlett: Yep, I think we should stick with just enforcing a minimum librdkafka version globally - more fine grained control doesn't seem to have enough benefit to be worth the hassle to me. We do have a situation at the moment where the optimal thing for some people to do is use librdkafka 0.9.5 with Confluent.Kafka 0.11.0 due to regressions in librdkafka, and if we'd enforced 0.11.0 in Confluent.Kafka they couldn't do this. However hopefully this should happen rarely, and it's also not such a big deal to downgrade to Confluent.Kafka 0.9.5.

edenhill: But is it possible to downgrade to v0.9.5 with this dependency on offsets_store that was added in 0.11.0?

mhowlett: `GetDelegateForFunctionPointer` (which we will use going forward due to #340) throws an exception in the case of a `IntPtr.Null` which is what it gets if a symbol lookup fails and so the way things are planned to be currently, all the symbols must exist in librdkafka on first `Producer` or `Consumer` construction and there is no way to recover from this. 

But you give me an idea - I could wrap `GetDelegateForFunctionPointer` in another function that catches exceptions and instead returns a new delegate of the appropriate type that throws an exception if it is used.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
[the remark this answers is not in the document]
vinodres: @treziac I agree that in case of global error, it makes sense to throw Exception. I am changing the global error checking to throw exception instead of returning OffsetResults with Error.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -414,6 +414,28 @@ public void Unassign()
             => consumer.Unassign();
 
         /// <summary>
+        ///     Store offsets for one or more partitions.
+        ///     
+        ///     The offset will be committed (written) to the offset store according
+        ///     to `auto.commit.interval.ms` or manual offset-less commit().
+        /// </summary>
+        /// <remarks>
+        ///     `enable.auto.offset.store` must be set to "false" when using this API.
```
mhowlett: What happens if `enable.auto.offset.store` is `true`? We probably want an exception, and to get that we may need to pull out this config property in the Consumer constructor and and check it down on line 1100.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -414,6 +414,28 @@ public void Unassign()
...
+        /// </remarks>
+        /// <param name="message">
+        ///     The message used to determine partition of the offset.
```
mhowlett: maybe 'A message used to determine the offset to store and topic/partition.'?

src/Confluent.Kafka/Consumer.cs
```diff
@@ -414,6 +414,28 @@ public void Unassign()
...
+        /// </param>
+        /// <returns>
+        ///     Offset results with per-partition errors.
+        /// </returns>
+        public OffsetResults StoreOffset(Message<TKey, TValue> message)
```
mhowlett: I don't think we should have an `OffsetResults class` - just return `List<TopicPartitionErrorOffset>` and throw an exception on global error.

src/Confluent.Kafka/OffsetResults.cs
```diff
@@ -0,0 +1,79 @@
+﻿// Copyright 2016-2017 Confluent Inc.
```
mhowlett: don't need this class

mhowlett:
thanks @vinodres - finally got to reviewing this! let's try and get this in the Nov 20 release we're going to do.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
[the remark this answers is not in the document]
vinodres: @edenhill , will you be able to fix this bug you mentioned above and push it into the Nov 20 release?

src/Confluent.Kafka/Consumer.cs
vinodres: @mhowlett , I am about to add the validation. There are two validations that must be performed.
1. if enbale.auto.commit not true, then throw an exception. If you look at my most recent commit, I have updated the remarks saying this setting must be set to true.
2. As you requested if enable.auto.offset.store is true, also throw an exception. 

Don't you think these validations must be performed inside Librdkafka library rather than here? If you still strongly feel, please confirm and I will add them here.

edenhill: I suggest not adding validations to language bindings, but rather to librdkafka, if needed.

It is perfectly fine to have auto commits disabled and use offsets_store(), the stored offsets will be used on your next argument-less commit() call.

For the validation of enable.auto.offset.store being disabled, I think that has questionable implementation complexity / payoff ratio. We already document the fact as a remark on the offset_store API itself.

vinodres: @edenhill , thank you for the info about offsets_store() can be invoked when auto commit is disabled. I will remove the line from remarks that says 'enable.auto.commit' must be true.

vinodres: Done, removed 'enable.auto.commit' must be true from remarks.

vinodres:
@mhowlett / @edenhill, for this PR is there anything else you are waiting from me?

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_StoreOffsets.cs
```diff
@@ -0,0 +1,78 @@
...
+                {
+                    consumer.Poll(TimeSpan.FromSeconds(1));
+                }
+
+                Assert.False(consumer.Consume(out message, TimeSpan.FromSeconds(1)));
+                Assert.False(producer.ProduceAsync(topic, null, "test store offset value").Result.Error);
+
+                Assert.True(consumer.Consume(out message, TimeSpan.FromSeconds(30)));
+                var results = consumer.StoreOffset(message);
+
```
edenhill: check that results matchines message length and matching (topic,partition,offset)

src/Confluent.Kafka/Consumer.cs
mhowlett: > For the validation of enable.auto.offset.store being disabled, I think that has questionable implementation complexity / payoff ratio

I disagree on this. We should fail fast when we know the user has configured something incorrectly. Kafka is complex, people don't read docs, and they configure it incorrectly all the time and get frustrated. We should take any opportunity we have to push people towards the right thing. Many small details like this add up to ease of use.

However, I agree that this check should be in librdkafka, not `Confluent.Kafka`, so this PR needs no additions.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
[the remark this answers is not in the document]
edenhill: Yes, I will make it return UNKNOWN_PARTITION only if none of the partitions were found.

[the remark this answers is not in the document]
edenhill: and __INVALID_ARG if enable.auto.offset.reset is set to true (default)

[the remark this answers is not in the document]
edenhill: Please review: https://github.com/edenhill/librdkafka/pull/1523

mhowlett:
@edenhill - I don't think we need a unit test for this. 
This is now LGTM and I'll merge it if there are no other comments.

One more thing I think would be good is an example demonstrating the difference between storing and committing, but that can come later.

edenhill:
librdkafka master is now fixed, please verify.

PR #340, opened 2017-10-13
load librdkafka library dynamically

mhowlett:
work in progress. not finished. this change is going to allow a sequence of versions of librdkafka to be loaded in turn until one that works is found. this change is part of the solution to the SSL dependency problem we have with newer versions of debian (and also more obscure distro versions). we will also publish a librdkafka.redist nuget package with a wider range of librdkafka versions. 

Also, I think we should make librdkafka.redist an explicit dependency as 1. it will be getting quite big, and people may want to use a leaner package instead (and we may even want to make some of these ourselves) 2. we're never going to cover all cases in our standard package. 

this change may have perf impact. I need to test that.

### DIFF
--- a/examples/AdvancedConsumer/AdvancedConsumer.csproj
+++ b/examples/AdvancedConsumer/AdvancedConsumer.csproj
@@ -10,10 +10,4 @@
     <ProjectReference Include="../../src/Confluent.Kafka/Confluent.Kafka.csproj" />
   </ItemGroup>
 
-  <ItemGroup>
-    <PackageReference Include="librdkafka.redist.osx-x64" Version="0.11.1-TEST4" />
-    <PackageReference Include="librdkafka.redist.debian-x64" Version="0.11.1-TEST4" />
-    <PackageReference Include="librdkafka.redist.rhel-x64" Version="0.11.1-TEST4" />
-  </ItemGroup>
-
 </Project>
--- a/examples/AdvancedConsumer/AdvancedConsumer.csproj
+++ b/examples/AdvancedConsumer/AdvancedConsumer.csproj
@@ -10,4 +10,10 @@
     <ProjectReference Include="../../src/Confluent.Kafka/Confluent.Kafka.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <PackageReference Include="librdkafka.redist.osx-x64" Version="0.11.1-TEST4" />
+    <PackageReference Include="librdkafka.redist.debian-x64" Version="0.11.1-TEST4" />
+    <PackageReference Include="librdkafka.redist.rhel-x64" Version="0.11.1-TEST4" />
+  </ItemGroup>
+
 </Project>
--- a/src/Confluent.Kafka/Confluent.Kafka.csproj
+++ b/src/Confluent.Kafka/Confluent.Kafka.csproj
@@ -13,7 +13,7 @@
     <Title>Confluent.Kafka</Title>
     <AssemblyName>Confluent.Kafka</AssemblyName>
     <VersionPrefix>0.11.2</VersionPrefix>
-    <TargetFrameworks>net45;net46;netstandard1.3</TargetFrameworks>
+    <TargetFrameworks>net45;net46;netstandard1.5</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
   </PropertyGroup>
--- a/src/Confluent.Kafka/Confluent.Kafka.csproj
+++ b/src/Confluent.Kafka/Confluent.Kafka.csproj
@@ -19,7 +19,7 @@
   </PropertyGroup>
 
   <ItemGroup>		
-    <PackageReference Include="librdkafka.redist" Version="0.11.1" />		
+    <PackageReference Include="librdkafka.redist.win-x64" Version="0.11.1-TEST4" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.3' ">
--- a/src/Confluent.Kafka/Confluent.Kafka.csproj
+++ b/src/Confluent.Kafka/Confluent.Kafka.csproj
@@ -19,16 +19,7 @@
   </PropertyGroup>
 
   <ItemGroup>		
-    <PackageReference Include="librdkafka.redist.win-x64" Version="0.11.1-TEST4">
-      <IncludeAssets>all</IncludeAssets>
-      <ExcludeAssets>none</ExcludeAssets>
-      <PrivateAssets>none</PrivateAssets>
-    </PackageReference>
-    <PackageReference Include="librdkafka.redist.win-x86" Version="0.11.1-TEST4">
-      <IncludeAssets>all</IncludeAssets>
-      <ExcludeAssets>none</ExcludeAssets>
-      <PrivateAssets>none</PrivateAssets>
-    </PackageReference>
+    <PackageReference Include="librdkafka.redist" Version="0.11.1-TEST6" />
   </ItemGroup>
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard1.5' ">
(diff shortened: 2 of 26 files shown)


src/Confluent.Kafka/Confluent.Kafka.csproj
```diff
@@ -13,7 +13,7 @@
     <Title>Confluent.Kafka</Title>
     <AssemblyName>Confluent.Kafka</AssemblyName>
     <VersionPrefix>0.11.1-RC1</VersionPrefix>
-    <TargetFrameworks>net45;net46;netstandard1.3</TargetFrameworks>
```
treziac: Lot of peoples are still under net45 I think (unfortunately), and having confluent.Kafka available in net45 allowed us to deploy some kafka logic without having to rewrite all our apps or make an upgrade (of all prods machine, which is not something companies will be eager to do... perhaps do some poll on #clients in slack and/or here to get some feedback? )

test/Confluent.Kafka.StrongName.UnitTests/Confluent.Kafka.StrongName.UnitTests.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <AssemblyName>Confluent.Kafka.StrongName.UnitTests</AssemblyName>
-    <TargetFrameworks>netcoreapp1.1;net452</TargetFrameworks>
+    <TargetFrameworks>netcoreapp1.1;</TargetFrameworks>
```
treziac: Still need to test against netframework - the lib is netstandard, but it should work on all implementation of netstandard, mainly netcoreapp (perhaps test netcoreapp2 also) and net461

src/Confluent.Kafka/Confluent.Kafka.csproj
mhowlett: yes, taking these out is a temporary convenience (for intellisense in vs code on mac/linux + build) - don't look too closely at this pr yet, just the general gist of it..

src/Confluent.Kafka/Impl/LibRdKafka.cs
```diff
@@ -80,106 +113,215 @@ static void LoadNet45Librdkafka()
...
-#endif
+#else
 
-        static LibRdKafka()
+        static IntPtr TryLoadLibrdkafka(string librdkafkaName, string librarySuffix, LibraryOpenDelegate openLib)
         {
-#if NET45 || NET46
-            try
+            string confluentKafkaAssemblyLocation = typeof(Error).GetTypeInfo().Assembly.Location;
+            if (!confluentKafkaAssemblyLocation.EndsWith("Confluent.Kafka.dll"))
```
treziac: Compare with OrdinalIgnoreCase. From experience, some scripts of deployment tends to use lower case everytime.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -213,13 +213,14 @@ internal ErrorCode Produce(string topic, byte[] val, int valOffset, int valLengt
                 // thread of the application calls poll() for a blocking produce() to ever unblock.
                 return LibRdKafka.producev(
                     handle,
```
mhowlett: it would be nicer structurally to wrap `producev` in `LibRdKafka` and expose a method in which the VarTags don't need to be specified, however it's only a minor improvement structurally, everything's internal so it hardly matters and we may get a slight performance benefit if the extra level of indirection doesn't get optimized away.

examples/AdvancedConsumer/AdvancedConsumer.csproj
```diff
@@ -10,4 +10,10 @@
     <ProjectReference Include="../../src/Confluent.Kafka/Confluent.Kafka.csproj" />
   </ItemGroup>
```
mhowlett: windows platform packages are transitively linked via Confluent.Kafka (x86 and x64). Any other platform requires an additional package reference.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_OffsetsForTimes.cs
```diff
@@ -30,12 +30,12 @@ public static partial class Tests
         ///     Basic OffsetsForTimes test on Consumer.
         /// </summary>
         [Theory, MemberData(nameof(KafkaParameters))]
-        public static async Task Consumer_OffsetsForTimes(string bootstrapServers, string topic, string partitionedTopic)
+        public static void Consumer_OffsetsForTimes(string bootstrapServers, string topic, string partitionedTopic)
```
mhowlett: this was failing. my theory was having it async meant other tests were interfering with it via interleaved execution. making it synchronous fixed the failure.

mhowlett:
Summary of problems:

1. librdkafka needs to dynamically link against libssl.so. We don't want to statically link because we want any security patches the system administrator might apply to apply to librdkafka.
2. different platforms may come with different (incompatible) versions of libssl. In particular, there is a version change between microsoft's dotnet docker images (1.1 -> 2.0). 
3. There is no RID ([runtime identifier](https://docs.microsoft.com/en-us/dotnet/core/rid-catalog)) for debian.9 upon which the 2.0 docker image is based. In fact, dotnet 2.0 seems to do away with these fine-grained RID's and the way forward seems to be to use linux-x64 as a general platform identifier. We should be able to make things work right now by including a libssl1.0.2 linked librdkafka in linux-x64 and a libssl1.0.0 linked librdkafka in debian.8-x64. This may solve our problem now, but it's not a robust solution going forward. I haven't actually tried to do this because it's not future proof, though I think it would work.
4. We could have a variety of different versions of librdkafka named different things - eg librdkafka-ssl1.0.2.so in the same RID directory (probably linux-x64). However, the `[DllImport]` attribute method of loading the library needs to have the library name statically specified. So we can't use this (how Confluent.Kafka has traditionally worked) if we do that.
5. Actually, maybe we could duplicate all the native method interop definitions (~50 of them) for every variant of the native library we supply, and catch exceptions if the library can't be loaded. This may be possible (though I need to think more about this, exceptions will be thrown on static method invokation, not class instantiation). Anyway, at best it's very messy and error prone, so not ideal.
6.  For the above reason, this PR includes changes to use platform specific dynamic library loading + symbol lookup calls (i.e. p/invoke to `dlopen` and `dlsym` in `libdl.so` on linux) - if we use  these, we can specify any name we like.
7. However these calls are outside the dotnet world. They do not know about RIDs or nuget directories. If we use these we effectively must specify an absolute path because the native libraries are not in the system library loading path.
8. But unfortunately, the way libraries are deployed in .NET Core is very complex. If you execute your software by doing dotnet run, the librdkafka runtime libraries aren't copied to the output directory, they remain in the nuget cache, which is not in the same place on all platforms, and is also user specifiable. If you do dotnet pack before running, these are placed side-by-side your application dlls. The dotnet runtime knows about all this, but for us to get absolute paths we need to reverse engineer it. This is getting very magical and error prone, and it's a lot of work to check it works in all deployment scenarios. And I'm scared I'll miss scenarios. e.g. what if Confluent.Kafka it's self is dynamically loaded at runtime? It's too fragile for my liking.
9. So that made me start thinking we should deploy the librdkafka variations as content. This way, they are in the application output directory always and we can reliably get an absolute path to them.
10. I've made this work, but I found achieving it in practice was a nightmare. The nuget package format is a mess of arbitrariness with extra arbitrariness added all over the place as support for different variations on the .net platform have been tacked on. I've been finding out how things work from github issues, not official documentation.
11. Having achieved this, it seems preview releases of VS2017 are now showing a visual representation of content that gets copied to the output directory in the solution explorer. This is not what people are used to seeing and it's freaked at least one person out #346. It feels messy. I want to avoid it if possible.
12. I have one more idea. If we include a 'dummy' managed library in the nuget package, we can reference it, then get the assembly location. I think the dummy managed library will always be side-by-side the other content in the package, so I can use this location to get an absolute path for the native libraries we want to try. 
13. Finally, the packaging code for librdkafka.redist was not built with extending to multiple  platform specific nuget packages in mind, and it's pretty complex because the librdkafka builds it pulls in were not designed for use by this script. so I've found this quite difficult to work with.

plan: I'm going to explore 12. If this doesn't work, we're probably stuck with the content method (currently working, but packaging scripts need tidying up).

mhowlett:
closing in favor of #355

PR #342, opened 2017-10-18
removed librdkafka.redist dependency. added note to readme

mhowlett:
@edenhill - as discussed. unfortunately this will fail with a non-descriptive exception if people don't reference librdkafka.redist. I don't think it's easy to give a better warning until the SSL dependency related changes go in. I'll test again on all platforms before tagging and deploying to nuget.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ client. Thanks Andreas!
 Reference the [Confluent.Kafka](https://www.nuget.org/packages/Confluent.Kafka/) and 
 [librdkafka.redist](https://www.nuget.org/packages/librdkafka.redist/) NuGet packages (version 0.11.1).
 
-**Important Note:** The librdkafka.redist package dependency is no longer transitive. You must now reference the 
+**Important Note:** The librdkafka.redist package dependency is no longer implicit. You must now reference the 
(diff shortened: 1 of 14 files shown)


mhowlett:
oh, I need to add the librdkafka.redist dependency to all the test and example projects.

README.md
```diff
@@ -29,11 +29,20 @@ client. Thanks Andreas!
 
 ## Usage
 
-Reference the [Confluent.Kafka NuGet package](https://www.nuget.org/packages/Confluent.Kafka/) (version 0.11.1).
+Reference the [Confluent.Kafka](https://www.nuget.org/packages/Confluent.Kafka/) and 
+[librdkafka.redist](https://www.nuget.org/packages/librdkafka.redist/) NuGet packages (version 0.11.1).
 
-To install confluent-kafka-dotnet from within Visual Studio, search for Confluent.Kafka in the NuGet Package Manager UI, or run the following command in the Package Manager Console:
+**Important Note:** The librdkafka.redist package dependency is no longer transitive. You must now reference the
```
edenhill: transitive? Probably correct but I would use something **I** understand, like implicit ;)

mhowlett: yep. it felt funny writing it.

PR #348, opened 2017-10-24
added back implicit reference to librdkafka.redist

mhowlett:
@edenhill - opening this PR so appveyor will produce a CI nuget package I can test. 
should be good to go, but will test it first.


mhowlett:
wrong branch

PR #354, opened 2017-10-29
Bugfix for .NET 4.7 Azure Functions project with librdkafka.redist 11.1.

Quass1m:
Something I came across while trying to create an Azure Function with Kafka client. The .dll files from explicitly specified librdkafka.redist version 11.1 were not being placed in the x64 or x86 subdirectory of bin folder.


mhowlett:
thanks @Quass1m. I'm in the process of completely changing how library loading works over in #340. I hadn't yet tried this out with net47, but based on your PR it looks like I probably need to.

mhowlett:
I added a commit to #355 that mimics your changes (feel free to review that commit).

PR #355, opened 2017-10-30
changes to librdkafka loading

mhowlett:
another approach to loading librdkafka. opening a PR now to run tests across platforms. WIP until they all pass.


src/Confluent.Kafka/Consumer.cs
```diff
@@ -751,6 +751,16 @@ private void LogCallback(IntPtr rk, int level, string fac, string buf)
         /// </remarks>
         public Consumer(IEnumerable<KeyValuePair<string, object>> config)
         {
+            var librdkafkaPath = config.FirstOrDefault(prop => prop.Key == "librdkafka.path").Value;
+            config = config.Where(prop => prop.Key != "librdkafka.path");
+
+            if (LibRdKafka.IsInitialized && librdkafkaPath != null)
```
mhowlett: there is actually a possible benign race condition here that will never matter in practice.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Pause_Resume.cs
```diff
@@ -30,7 +30,7 @@ public static partial class Tests
         ///     Simple Consumer Pause / Resume test.
         /// </summary>
         [Theory, MemberData(nameof(KafkaParameters))]
-        public static async Task Consumer_Pause_Resume(string bootstrapServers, string topic, string partitionedTopic)
+        public static void Consumer_Pause_Resume(string bootstrapServers, string topic, string partitionedTopic)
```
mhowlett: this was a benign bug, causes a warning.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_OffsetsForTimes.cs
```diff
@@ -30,12 +30,12 @@ public static partial class Tests
         ///     Basic OffsetsForTimes test on Consumer.
         /// </summary>
         [Theory, MemberData(nameof(KafkaParameters))]
-        public static async Task Consumer_OffsetsForTimes(string bootstrapServers, string topic, string partitionedTopic)
+        public static void Consumer_OffsetsForTimes(string bootstrapServers, string topic, string partitionedTopic)
```
mhowlett: this was failing. I believe it's because having it async meant other tests were interfering with it due to interleaved execution. making it synchronous fixed the failure.

Quass1m:
Thank you for adding path check from #354. After fixing the NativeMethods issue the code compiled and the proper location of librdkafka.dll was used.

mhowlett:
note to anyone following along: this PR is waiting for us to make a new librdkafka.redist compatible with the changes before being finalized / merging.

mhowlett:
@edenhill - this is ready for review (except I'll reference librdkafka.redist 0.11.3 when ready)

mhowlett:
Just checked, and neither of the librdkafka builds are compatible with centos unfortunately, but this is not a regression and at least now there is a workaround via `Library.Load`.

src/Confluent.Kafka/Impl/LibRdKafka.cs
```diff
@@ -37,155 +40,283 @@ internal static class LibRdKafka
...
+
+                    nativeMethodTypes.Add(typeof(NativeMethods.NativeMethods));
+                }
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+                {
+                    if (userSpecifiedPath != null)
+                    {
+                        if (LinuxNative.dlopen(userSpecifiedPath, RTLD_NOW) == IntPtr.Zero)
+                        {
+                            throw new DllNotFoundException($"Failed to load librdkafka: '{userSpecifiedPath}'");
```
edenhill: Add the error message (dlerror()) too.

src/Confluent.Kafka/Impl/LibRdKafka.cs
```diff
@@ -37,155 +40,283 @@ internal static class LibRdKafka
...
+                var nativeMethodTypes = new List<Type>();
+
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ||
+                    RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+                {
+                    if (userSpecifiedPath != null)
+                    {
+                        if (OsxNative.dlopen(userSpecifiedPath, RTLD_NOW) == IntPtr.Zero)
+                        {
+                            throw new DllNotFoundException($"Failed to load librdkafka: '{userSpecifiedPath}'");
```
edenhill: Add the error message (dlerror()) too.

src/Confluent.Kafka/Impl/LibRdKafka.cs
```diff
@@ -37,155 +40,283 @@ internal static class LibRdKafka
...
+                isInitialized = SetDelegates(typeof(NativeMethods.NativeMethods));
+
+#else
+
+                const int RTLD_NOW = 2;
+
+                var nativeMethodTypes = new List<Type>();
+
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX) ||
+                    RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
```
edenhill: Should Windows use OsxNative ?

src/Confluent.Kafka/Impl/LibRdKafka.cs
```diff
@@ -37,155 +40,283 @@ internal static class LibRdKafka
...
+                    {
+                        if (OsxNative.dlopen(userSpecifiedPath, RTLD_NOW) == IntPtr.Zero)
+                        {
+                            throw new DllNotFoundException($"Failed to load librdkafka: '{userSpecifiedPath}'");
+                        }
+                    }
+
+                    nativeMethodTypes.Add(typeof(NativeMethods.NativeMethods));
+                }
+                if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
```
edenhill: else if

src/Confluent.Kafka/Impl/LibRdKafka.cs
```diff
@@ -37,155 +40,283 @@ internal static class LibRdKafka
...
+                        }
+                    
+                        nativeMethodTypes.Add(typeof(NativeMethods.NativeMethods));
+                    }
+                    else 
+                    {
+                        nativeMethodTypes.Add(typeof(NativeMethods.NativeMethods));
+                        nativeMethodTypes.Add(typeof(NativeMethods.NativeMethods_Debian9));
+                    }
+                }
```
edenhill: else  raise unsupported platform

src/Confluent.Kafka/Impl/LibRdKafka.cs
```diff
@@ -37,155 +40,283 @@ internal static class LibRdKafka
...
+                {
+                    // catch the last win32 error by default and keep the associated default message
+                    var win32Exception = new Win32Exception();
+                    var additionalMessage =
+                        $"Error while loading librdkafka.dll or its dependencies from {path}. " +
+                        $"Check the directory exists, if not check your deployment process. " +
+                        $"You can also load the library and its dependencies by yourself " +
+                        $"before any call to Confluent.Kafka";
+
+                    throw new InvalidOperationException(additionalMessage, win32Exception);
```
edenhill: Add system error message (GetLastError())

src/Confluent.Kafka/Impl/NativeMethods/NativeMethods_Debian9.cs
```diff
@@ -0,0 +1,360 @@
...
+    /// <summary>
+    ///     This class should be an exact replica of other NativeMethods classes, except
+    ///     for the DllName const.
+    /// </summary>
+    /// <remarks>
+    ///     This copy/pasting is required because DllName must be const. 
+    ///     TODO: generate the NativeMethods classes at runtime (compile C# code) rather
+    ///     than copy/paste.
+    /// 
+    ///     Alternatively, we could have used dlopen to load the native library, but to
```
edenhill: Is this statement still correct? I thought we were infact using dlopen()

src/Confluent.Kafka/Library.cs
```diff
@@ -57,5 +57,39 @@ public static string VersionString
...
+        /// </summary>
+        public static bool IsLoaded
+            => LibRdKafka.IsInitialized;
+
+        /// <summary>
+        ///     Loads the native librdkafka library. Does nothing if the library is
+        ///     already loaded.
+        /// </summary>
+        /// <returns>
+        ///     true if librdkafka was loaded as a result of this call, false otherwise.
```
edenhill: Will loading exceptions be raised here?

src/Confluent.Kafka/Producer.cs
```diff
@@ -272,6 +272,11 @@ public void HandleDeliveryReport(Message message)
         /// </param>
         public Producer(IEnumerable<KeyValuePair<string, object>> config, bool manualPoll, bool disableDeliveryReports)
         {
+            if (!LibRdKafka.IsInitialized)
```
edenhill: dito, already checked by Initialize()

test/Confluent.Kafka.StrongName.UnitTests/Confluent.Kafka.StrongName.UnitTests.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <AssemblyName>Confluent.Kafka.StrongName.UnitTests</AssemblyName>
-    <TargetFrameworks>netcoreapp1.1;net452</TargetFrameworks>
+    <TargetFrameworks>netcoreapp1.1;net452;</TargetFrameworks>
```
edenhill: This is just noise, right? If so I suggest removing this diff for future git blamer's sake

test/Confluent.Kafka.UnitTests/Confluent.Kafka.UnitTests.csproj
```diff
@@ -4,7 +4,7 @@
     <ProjectTypeGuids>{3AC096D0-A1C2-E12C-1390-A8335801FDAB};{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
     <TestProjectType>UnitTest</TestProjectType>
     <AssemblyName>Confluent.Kafka.UnitTests</AssemblyName>
-    <TargetFrameworks>netcoreapp1.1;net452</TargetFrameworks>
+    <TargetFrameworks>netcoreapp1.1;net452;</TargetFrameworks>
```
edenhill: dito

test/Confluent.Kafka.StrongName.UnitTests/Confluent.Kafka.StrongName.UnitTests.csproj
```diff
@@ -24,7 +24,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="librdkafka.redist" Version="0.11.1" />
+    <PackageReference Include="librdkafka.redist" Version="0.11.1-TEST101" />
```
edenhill: dito, when change

test/Confluent.Kafka.UnitTests/TopicPartitionOffsetError.cs
```diff
@@ -104,6 +106,8 @@ public void Properties()
         [Fact]
         public void ExplicitCast()
         {
+            Library.Load();
```
edenhill: Suggest adding a negative test that attempts to specify a non-existent librdkafka to make sure error reporting works as expected.
Also a positive test specifying an existing librdkafka.
(but neither of these obviously not here, somewhere else)

edenhill:
How are we verifying this new functionality across affected platforms? (centos, debian < 9, debian 9, windows, osx)

Do we have something like a script that fires up docker on Linux and verifies the Linux scenarios?
I understand that this is hard to automate in its entirety, but per platform it shouldn't be too hard.

src/Confluent.Kafka/Confluent.Kafka.csproj
```diff
@@ -19,7 +19,7 @@
   </PropertyGroup>
 
   <ItemGroup>		
-    <PackageReference Include="librdkafka.redist" Version="0.11.1" />		
+    <PackageReference Include="librdkafka.redist" Version="0.11.1-TEST101" />
```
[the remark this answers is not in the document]
mhowlett: As soon as 0.11.3 is on nuget. I don't mind if we merge this PR before or after that happens.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -769,6 +769,11 @@ private void LogCallback(IntPtr rk, int level, string fac, string buf)
         /// </remarks>
         public Consumer(IEnumerable<KeyValuePair<string, object>> config)
         {
+            if (!LibRdKafka.IsInitialized)
```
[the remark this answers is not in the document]
mhowlett: good point, will fix. exactly what was happening here was in a state of flux for a while. missed this.

src/Confluent.Kafka/Impl/LibRdKafka.cs
mhowlett: good point, though using `GetLastError` is not ideal way to get this evidently. `Marshal.GetLastWin32Error` instead. https://stackoverflow.com/questions/17918266/winapi-getlasterror-vs-marshal-getlastwin32error

mhowlett: actually, this is automatic, the existing code is fine.

src/Confluent.Kafka/Impl/LibRdKafka.cs
```diff
@@ -37,155 +40,283 @@ internal static class LibRdKafka
...
+            public static extern IntPtr dlerror();
 
-        [DllImport("kernel32", SetLastError = true)]
-        private static extern IntPtr GetModuleHandle(string lpFileName);
+            [DllImport("libdl.so")]
+            public static extern IntPtr dlsym(IntPtr handle, String symbol);
+        }
 
-        static void LoadNet45Librdkafka()
+        private static class OsxNative
```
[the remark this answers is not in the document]
mhowlett: i did this on purpose for readability, but you're right, PosixNative is probably better, changing to that.

src/Confluent.Kafka/Impl/LibRdKafka.cs
```diff
@@ -37,155 +40,283 @@ internal static class LibRdKafka
...
-                    // If he did it before first call to Confluent.Kafka, GetModuleHandle will be nonZero
-                    // and we don't have to force the load (which would fail and throw otherwise)
-                    LoadNet45Librdkafka();
+                    return isInitialized;
                 }
             }
-            catch
+        }
+
+        public static bool Initialize(string userSpecifiedPath)
```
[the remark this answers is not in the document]
mhowlett: not a bad idea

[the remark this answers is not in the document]
mhowlett: this is actually a bit tricky to set up as the initialize code is called only once for the entire set of tests. there is little risk here, skipping.

src/Confluent.Kafka/Impl/LibRdKafka.cs
mhowlett: yep, I screwed up the logic here, fixed.

src/Confluent.Kafka/Impl/LibRdKafka.cs
mhowlett: good idea. not a problem now, but may be in the future if more platforms are added.

src/Confluent.Kafka/Impl/NativeMethods/NativeMethods_Debian9.cs
mhowlett: this comment is still close enough to correct I think. there is now the option of the user specifying a path, in which case `dlopen` is used. If not, the attribute method (that this comment is referring to) is used.

src/Confluent.Kafka/Library.cs
mhowlett: i will clarify. it should read 'false if the library has already been loaded prior to this call'

test/Confluent.Kafka.StrongName.UnitTests/Confluent.Kafka.StrongName.UnitTests.csproj
mhowlett: good point but I think it's only a very small amount of noise

test/Confluent.Kafka.StrongName.UnitTests/Confluent.Kafka.StrongName.UnitTests.csproj
mhowlett: we can't change this until librdkafka.redist 0.11.3 is live on nuget. any time after that.

test/Confluent.Kafka.UnitTests/TopicPartitionOffsetError.cs
mhowlett: I'll add a test for the negative case.

mhowlett:
Just noticed `libdl.so` is not in the microsoft/dotnet docker images by default. This means specifying a library manually using `Library.Load(string path)` won't work out of the box. Minor point but slightly annoying.

src/Confluent.Kafka/Impl/LibRdKafka.cs
edenhill: We want a meaningful error message (e.g., "Platform not supported") if people try to run this on an unsupported platform, regardless of which platforms we currently support.

edenhill:
@mhowlett 
How are we verifying this new functionality across affected platforms? (centos, debian < 9, debian 9, windows, osx)

Do we have something like a script that fires up docker on Linux and verifies the Linux scenarios?
I understand that this is hard to automate in its entirety, but per platform it shouldn't be too hard.

src/Confluent.Kafka/Impl/LibRdKafka.cs
mhowlett: I mean we cover all platforms microsoft support, so the else will only get hit if microsoft start supporting additional platforms.

mhowlett:
@edenhill - for the moment I'm verifying all cases manually.

mhowlett:
I don't think automated testing of this is something to block the release on, it's not something we do already.

edenhill:
I think we should write a script that automates the actual tests, which is not  really the same as automated tests (automated test environment, bla bla, ci, etc..)
It should be wortwhile to jot down the steps in a bash script that fires up some docker thingies, that cost will be payed off by the next release, as opposed to "manual testing instructions".
My two cents

mhowlett:
i'm still just ad hoc verifying everything works (it doesn't). formalizing it a bit would be good yes.

mhowlett:
This is good to merge now from my perspective. I've tested librdkafka auto-loading / loading from user specified path on microsoft/dotnet docker images v1.1 and v2 + on Windows x64 with .NET Core & .NET Framework + OSX. centos is not currently supported. We can update the reference to librdkafka.redist to 0.11.3 when this is available on nuget at the same time we update the library version numbers. 

I don't think there is payback in automating these tests right now since it will take me a while to do well enough to check in here, it's v. easy to do manually, on windows I explicitly want to do the steps inside visual studio + we should be doing smoke tests implicitly sooner rather than later by automating integration tests.

PR #357, opened 2017-10-31
avro and schema registry support

mhowlett:
importing @treziac's schema registry PoC into my own branch and re-opening PR.


comdw:
In our use case we want to be able to prevent automatic registration of schemata/versions that haven't previously been registered.  I think we'd also need the proposed `Task<bool> CheckSchemaAsync(string subject, string schema)` implemented in order to do that - calling `GetLatestSchemaAsync` followed by `GetSchemaAsync` for each earlier version seems a little inefficient.

Is CheckSchemaAsync likely to make it in this version?

LorenzoRompianesi:
Any news for the AVRO Schema-registry integration?

LorenzoRompianesi:
Dear mhowlett,
  as I can see the error from travis-ci can be addressed by using a newer version of log4net 1.2.10 is not compatible with netcore2. Can you do this adjustment?

mhowlett:
@LorenzoRompianesi - we prioritised some other things for the up-coming release. I expect to get this merged to master January sometime.

mateen-dar:
Hello @mhowlett ,

I am looking for a simple example for a .NET consume that can read AVRO Messages. Can you please point to an example or a timeline when this is expected to be available?

mhowlett:
End of January.

Note that the initial merge will support .NET Framework only. Do you require support for .NET Core? 

Note there is a build of this PR on nuget already, but it's not ready for production use.

https://www.nuget.org/packages/Confluent.Kafka.SchemaRegistry
https://www.nuget.org/packages/Confluent.Kafka.Avro

joelduvall:
+1 For .Net Core 2.0 support

mhowlett:
I've just uploaded another CI build of this (beta quality) to nuget.org.

https://www.nuget.org/packages/confluent.kafka.avro (version 1.0.0-ci-199)

You're now encouraged to use this and report any problems, which are likely to get fixed quickly. I'm not expecting major issues - this PR has been worked through quite thoroughly.

`Confluent.Kafka.Avro` has implicit dependencies on `Confluent.Kafka` 0.11.3-ci-199, `Confluent.Kafka.Schema` 1.0.0-ci-199 and `Confluent.Apache.Avro` which is a fork of the official `Apache.Avro` project. Also `Newtonsoft.Json` 10.0.3 which some may find problematic, this will be relaxed to a lower version in a future build.

Requires .NET Framework >= 4.5.2. Only includes Specific serializers - you need to create classes from your avro schemas using the avrogen.exe tool. You can get this here: https://github.com/apache/avro (you need to build it yourself).

Example usage is here: https://github.com/mhowlett/confluent-kafka-dotnet/blob/0a8af98f0ff618ca7243322c1b56881144c7941f/examples/AvroSpecific/Program.cs

mhowlett:
Note: For compatibility with .NET Standard (i.e. .NET Core), we will need to do some work on the apache avro library. This is relatively high priority but I don't expect we'll be doing it immediately.  If anyone would like to contribute this, feel free to open a PR on: https://github.com/confluentinc/avro 

Since .NET Standard 2.0 is getting pretty complete c.f. .NET Framework now, I don't expect this is a huge amount of work, but it might take a bit of effort.

mhowlett:
@svantreeck - we'll provide `Avro.Generic.GenericRecord` serializer/deserializers very similar to java  (e.g. see https://github.com/confluentinc/schema-registry/blob/master/avro-serializer/src/test/java/io/confluent/kafka/serializers/KafkaAvroSerializerTest.java), however not in this PR - we decided to implement specific serdes first.

examples/AvroSpecific/Program.cs
```diff
@@ -0,0 +1,124 @@
...
+            // };
+            // 
+            // using (var schemaRegistryClient = new CachedSchemaRegistryClient(schemaRegistryConfig))
+            // using (var consumer = new Consumer<User, User>(consumerConfig, new AvroDeserializer<User>(schemaRegistryClient), new AvroDeserializer<User>(schemaRegistryClient)))
+            // using (var producer = new Producer<User, User>(producerConfig, new AvroSerializer<User>(schemaRegistryClient), new AvroSerializer<User>(schemaRegistryClient)))
+            // {
+            // ...
+
+            // Note: The User class in this project was generated using the Confluent fork of the avrogen.exe tool 
+            // (avaliable from: https://github.com/confluentinc/avro/tree/confluent-fork) which includes modifications
```
mhowlett: I'll make our version of avrogen.exe available via some more convenient means at release time.

src/Confluent.Kafka.Avro/AvroDeserializer.cs
```diff
@@ -0,0 +1,223 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Refer to LICENSE for more information.
+
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Net;
+using Avro.Generic;
```
mhowlett: The .NET version of the apache avro library will be **much** less battle tested than the java version. I hope it's not going to cause us too many problems.

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -0,0 +1,191 @@
...
+                this.schemaByVersionBySubject.Sum(x => x.Value.Count) +
+                this.idBySchemaBySubject.Sum(x => x.Value.Count)
+                    >= identityMapCapacity)
+            {
+                throw new OutOfMemoryException("Local schema cache maximum capacity exceeded");
+            }
+        }
+
+        /// <include file='include_docs.xml' path='API/Member[@name="ISchemaRegistryClient_GetIdAsync"]/*' />
+        public async Task<int> GetIdAsync(string subject, string schema)
```
mhowlett: Using `string` here instead of `Avro.Schema` differs from the Java implementation and is a little bit controversial. It would be good to have another discussion about this with @mageshn or @ewencp.

examples/AvroSpecific/Program.cs
```diff
@@ -0,0 +1,124 @@
...
+                // schema.registry.url property for redundancy (comma separated list). 
+                // The property name is not plural to follow the convention set by
+                // the Java implementation.
+                { "schema.registry.url", schemaRegistryUrl },
+                // optional avro / schema registry client properties:
+                { "avro.buffer.bytes", 50 },
+                { "avro.auto.register.schema", true },
+                { "schema.registry.timeout.ms", 5000 },
+                { "schema.registry.cache.capacity", 10 }
+            };
```
mhowlett: currently no authorisation / authentication capability. I suspect this will be requested soon enough..

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
ewencp: I prefer `Schema` for a few reasons:

1. If you already have a `Schema` object, you are forced to do the conversion, which seems a bit awkward.
2. Catch issues earlier. If the schema isn't even valid, we should know this before calling this method. I suspect this is expressed as a string here since you could potentially skip any validation and just pass along the string in the request to the server and assume it will validate. This could be fine since we do specifically have a special error code 42201 to indicate this issue, but I prefer catching the issue before ever making the request.
3. Related to 2, separation of concerns. I like keeping the methods here simpler if possible (i.e. no parsing in the method, we already know we have a valid schema so the types of possible errors are reduced).

That said, it kinda depends on what APIs the .net Avro library has. In the Java ones, people are frequently using SpecificRecord and the `Schema` is available directly as a `static` field, making it trivial to get at and natural to use. If the .net library was different and, e.g., just exposed it as a `string` instead, this might be a fairly natural fit.

mhowlett: 1. Yes. Actually we do have this in the `ISpecificRecord`'s in the dotnet case as well, and the serializer caches a `ToString`'d version of this so the conversion doesn't need to be done on every serialize. It's awkward but only very slightly and hidden.

2, 3. On the flip side, `Confluent.SchemaRegistry` currently has no dependency on `Apache.Avro` and accepting `Avro.Schema` would require this. You might argue most people using `Confluent.SchemaRegistry` will want this dependency anyway, but I also quite like that the current implementation is a more basic building block. I guess this is the main thing driving my opinion here (since the benefit to introducing the dependency although there is not huge). Still, the arguments the other way are perhaps stronger.

Two more things to note: 

1. This implementation is not susceptible to the reference comparison problems we saw in the Java version due to different instances of the Schema object representing the same schema being used. Not that that is inherently due to use of `Schema`, but in this implementation, there is no problem by default.

2. We can add an overload of these methods later if need be (numerous method overloads is common in .NET). I don't think i'll regret having the `string` overload, i.e. won't want to depreciate it.

mhowlett:
@edenhill - ball is now in your court - ready for review!

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
ewencp: The other argument for using string is if we ever extend this to non-Avro as well. Although maybe string is worse in that case as you need some separate overrides for other languages.

re: the reference comparison thing -- that specifically has to do with caching. To properly resolve this, wouldn't an implementation here need to parse and then re-stringify the avro to ensure it has the canonicalized version? Otherwise schema strings with slightly different spacing would have effectively the same issue as, no?

src/Confluent.SchemaRegistry/Confluent.SchemaRegistry.csproj
```diff
@@ -0,0 +1,12 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net452;netstandard1.4;</TargetFrameworks>
```
edenhill: the previous csproj files only had net452

src/Confluent.SchemaRegistry/Rest/DataContracts/Schema.cs
```diff
@@ -0,0 +1,183 @@
...
+        public string Subject { get; set; }
+
+        /// <summary>
+        ///     The schema version.
+        /// </summary>
+        [DataMember(Name = "version")]
+        public int Version { get; set; }
+
+        /// <summary>
+        ///     Globally unique identifier of the schema.
```
edenhill: Not so sure about Globally

src/Confluent.SchemaRegistry/Rest/DataContracts/Schema.cs
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        /// <summary>
+        ///     Returns a string representation of the Schema object.
+        /// </summary>
+        /// <returns>
+        ///     A string that represents the schema object.
+        /// </returns>
+        public override string ToString()
+            => $"{{subject={Subject}, version={Version}, id={Id}, schema={SchemaString}}}";
```
edenhill: do you really want to print the full schema def here? Seems like subject, version, id are sufficient.

src/Confluent.SchemaRegistry/Rest/DataContracts/Schema.cs
```diff
@@ -0,0 +1,183 @@
...
+        /// <returns>
+        ///     A 32-bit signed integer that indicates whether this instance precedes, follows, or
+        ///     appears in the same position in the sort order as the other parameter. Less than 
+        ///     zero: this instance precedes other. Zero: this instance has the same position in
+        ///     the sort order as other. Greater than zero: This instance follows other OR other 
+        ///     is null.
+        /// </returns>
+        public int CompareTo(Schema other)
+        {
+            int result = string.Compare(Subject, other.Subject, StringComparison.Ordinal);
```
edenhill: there is no check for other being null

src/Confluent.SchemaRegistry/Rest/DataContracts/Schema.cs
```diff
@@ -0,0 +1,183 @@
...
+            {
+                return false;
+            }
+
+            Schema that = (Schema)obj;
+            return Equals(that);
+        }
+
+        /// <summary>
+        ///     Determines whether this instance and another specified Schema object have the same value.
```
edenhill: not clear what "value" is here

src/Confluent.SchemaRegistry/Rest/RestService.cs
```diff
@@ -0,0 +1,302 @@
...
+        /// <summary>
+        ///     HttpClient instances corresponding to each provided schema registry Uri.
+        /// </summary>
+        private readonly List<HttpClient> clients;
+
+        
+        /// <summary>
+        ///     Initializes a new instance of the RestService class.
+        /// </summary>
+        public RestService(string schemaRegistryUris, int timeoutMs)
```
edenhill: nit: Uris is probably technically correct, but since url is used in other places it'd be better if they were the same (for grepping, etc)

src/Confluent.SchemaRegistry/Rest/RestService.cs
```diff
@@ -0,0 +1,302 @@
...
+            bool gotOkAnswer = false;
+            bool firstError = true;
+            
+            int startClientIndex;
+            lock (lastClientUsedLock)
+            {
+                startClientIndex = this.lastClientUsed;
+            }
+
+            for (int i = startClientIndex; i < clients.Count + startClientIndex && !gotOkAnswer; ++i)
```
edenhill: This will just try client startClientIndex..end, it should roll-over and start from 0 until it reaches startClientIndex again.

src/Confluent.SchemaRegistry/Rest/RestService.cs
```diff
@@ -0,0 +1,302 @@
...
+                var errorObject = JObject.Parse(await response.Content.ReadAsStringAsync().ConfigureAwait(false));
+                var message = errorObject.Value<string>("message");
+                var errorCode = errorObject.Value<int>("error_code");
+                throw new SchemaRegistryException(message, response.StatusCode, errorCode);
+            }
+
+            // if we had success, set last client used so we start with this next time.
+            lock (lastClientUsedLock)
+            {
+                this.lastClientUsed = (this.lastClientUsed + 1) % clients.Count;
```
edenhill: shouldn't this be `lastClientUsed = i` ?

src/Confluent.SchemaRegistry/SchemaRegistryException.cs
```diff
@@ -0,0 +1,59 @@
...
+// Refer to LICENSE for more information.
+
+using System;
+using System.Net;
+
+
+namespace Confluent.SchemaRegistry
+{
+    /// <summary>
+    ///     Represents an error returned by Confuent Schema Registry.
```
edenhill: Confluent

test/Confluent.Kafka.Avro.IntegrationTests/Tests/ConsumeIncompatibleTypes.cs
```diff
@@ -0,0 +1,101 @@
...
+
+                Assert.True(hadError);
+            }
+
+            using (var consumer = new Consumer<string, string>(consumerConfig, new AvroDeserializer<string>(), new AvroDeserializer<string>()))
+            {
+                bool hadError = false;
+                consumer.OnConsumeError += (_, m) =>
+                {
+                    if (m.Error.Code == ErrorCode.Local_ValueDeserialization)
```
edenhill: Since the topic is reused for other stuff it would be good if this verified the offset or some msgid of the message being checked.

test/Confluent.Kafka.Avro.IntegrationTests/Tests/ConsumeIncompatibleTypes.cs
```diff
@@ -0,0 +1,101 @@
...
+                    {
+                        hadError = true;
+                    }
+                };
+
+                consumer.Assign(new List<TopicPartitionOffset> { new TopicPartitionOffset(topic, 0, 0) });
```
edenhill: start consuming at the produced offset

test/Confluent.Kafka.Avro.IntegrationTests/test.parameters.json
```diff
@@ -0,0 +1,4 @@
+﻿{
+  "bootstrapServer": "localhost:9092",
```
edenhill: not clear why these are not standard client config properties? (bootstrap.servers,..)

src/Confluent.Kafka.Avro/AvroSerializer.cs
```diff
@@ -0,0 +1,294 @@
...
+
+        private HashSet<string> topicsRegistered = new HashSet<string>();
+
+        private bool disposeClientOnDispose;
+
+        private string writerSchemaString { get; set; }
+
+        private const string InitialBufferSizePropertyName = "avro.buffer.bytes";
+
+        private const string AutoRegisterSchemaPropertyName = "avro.auto.register.schema";
```
edenhill: Where are these properties documented for the user?
And shouldnt it be schema.registry.auto.register.schemas?
Also, the Java client has a plural s at the end.

manuel-zulian:
Hi everyone. To begin with, I'd like to thank you for your work, I'm testing the prerelease version on a kafka+debezium+schema registry configuration and it works almost out of the box. 
There is one annoying thing missing, that is support for logicalTypes. I've looked into it, it seems it's almost entirely dependent on the avro library, so I tried to fork it (https://github.com/manuel-zulian/avro/tree/feature/logical-types). 
Now, the code is far from being decent, but I was in a rush to see if it could work, and it does indeed at least for reading decimal values. I changed the codegen so that the generated schema takes decimals in account as well.
There is a little bug due to the updated Json library (version 9 from version 3.5, that's not surprising) so you have to fix the generated schema manually.
This probably deserves a pull request of its own on the avro fork project. The question is: do you plan to implement it on your own (meanwhile I can rely on my version) or do you think we can build on my code? I must warn you though, I just stepped through with a debugger and changed where it made most sense, copying some stuff from the java version of Avro. It's very likely that I missed something even only in the reader part, because the project is a bit complex, but I think I could make something decent out of it with a bit of guidance.

Confluent.Kafka.sln
```diff
@@ -31,56 +30,246 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SimpleConsumer", "examples\
...
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "AvroSpecific", "examples\AvroSpecific\AvroSpecific.csproj", "{45B586D3-7CCA-46B0-A118-A070D84E8364}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Confluent.SchemaRegistry.UnitTests", "test\Confluent.SchemaRegistry.UnitTests\Confluent.SchemaRegistry.UnitTests.csproj", "{29DA188E-CB20-4A9C-AC4E-50B442B9AC7D}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Confluent.SchemaRegistry.IntegrationTests", "test\Confluent.SchemaRegistry.IntegrationTests\Confluent.SchemaRegistry.IntegrationTests.csproj", "{2F5F0775-3708-40B7-A71F-442E83F6BB3D}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		Debug|x64 = Debug|x64
```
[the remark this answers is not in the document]
mhowlett: I don't know, Visual Studio did this automatically. It's not causing problems anywhere.

src/Confluent.Kafka.Avro/AvroDeserializer.cs
```diff
@@ -0,0 +1,228 @@
...
+    /// </remarks>
+    public class AvroDeserializer<T> : IDeserializer<T>
+    {
+        private bool disposeClientOnDispose;
+
+        /// <remarks>
+        ///     A datum reader cache (one corresponding to each write schema that's been seen) 
+        ///     is maintained so that they only need to be constructed once.
+        /// </remarks>
+        private readonly Dictionary<int, DatumReader<T>> datumReaderBySchemaIdBigEndian = new Dictionary<int, DatumReader<T>>();
```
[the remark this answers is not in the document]
mhowlett: I thought you were into extreme optimizations? 
This is from @treziac's implementation and I saw no reason to change it.

src/Confluent.Kafka.Avro/AvroDeserializer.cs
```diff
@@ -0,0 +1,228 @@
...
+using System.Net;
+using Avro.Generic;
+using Avro.IO;
+using Avro.Specific;
+using Confluent.SchemaRegistry;
+using System.Reflection;
+using System;
+
+
+namespace Confluent.Kafka.Serialization
```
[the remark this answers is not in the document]
mhowlett: This is not a hard requirement in .NET (as it is in Java), and there is plenty of precedent deviating from this (e.g. the EntityFramework package has classes under `System.Data` mostly, and some under `System.ComponentModel`). I think these classes logically belong under `Confluent.Kafka.Serialization` and decided that argument was most important. I do admit to swapping to `Confluent.Kafka.Avro.Serialization` at one point, but swapped back - i.e. I'm not 100% confident I'm seeing this the right way. An advantage of putting classes under `Confluent.Kafka.Avro` like you suggest is it makes them easier to find for people who just installed the package with this name. @treziac's initial PR is how I have it now, so maybe he agrees.

src/Confluent.Kafka.Avro/AvroDeserializer.cs
```diff
@@ -0,0 +1,228 @@
...
+            {
+                ReaderSchema = Avro.Schema.Parse("long");
+            }
+            else if (typeof(T).Equals(typeof(byte[])))
+            {
+                ReaderSchema = Avro.Schema.Parse("bytes");
+            }
+            else
+            {
+                throw new InvalidOperationException(
```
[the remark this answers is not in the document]
mhowlett: No, but I'll change it to `ArgumentException` which is better.

src/Confluent.Kafka.Avro/AvroDeserializer.cs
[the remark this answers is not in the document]
mhowlett: ok, I changed it.

src/Confluent.Kafka.Avro/AvroDeserializer.cs
```diff
@@ -0,0 +1,228 @@
...
+        /// <param name="array">
+        ///     A byte array containing the object serialized in the format produced
+        ///     by <see cref="AvroSerializer" />.
+        /// </param>
+        /// <returns>
+        ///     The deserialized <see cref="T"/> value.
+        /// </returns>
+        public T Deserialize(string topic, byte[] array)
+        {
+            // Note: topic is not necessary for deserialization (or knowing if it's a key
```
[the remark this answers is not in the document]
mhowlett: yes, but subject is derived from topic, and i'm more talking about topic here.

src/Confluent.Kafka.Avro/AvroDeserializer.cs
```diff
@@ -0,0 +1,228 @@
...
+            // or value) only the schema id is needed.
+
+            using (var stream = new MemoryStream(array))
+            using (var reader = new BinaryReader(stream))
+            {
+                int magicByte = reader.ReadByte();
+                if (magicByte != Constants.MagicByte)
+                {
+                    // may change in the future.
+                    throw new InvalidDataException("magic byte should be 0");
```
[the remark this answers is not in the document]
mhowlett: spoken like someone with prior debugging experience.

src/Confluent.Kafka.Avro/AvroDeserializer.cs
```diff
@@ -0,0 +1,228 @@
...
+                    datumReaderBySchemaIdBigEndian.TryGetValue(writerIdBigEndian, out datumReader);
+                }
+                if (datumReader == null)
+                {
+                    int writerId = IPAddress.NetworkToHostOrder(writerIdBigEndian);
+                    string writerSchemaJson = SchemaRegistryClient.GetSchemaAsync(writerId).Result;
+                    var writerSchema = Avro.Schema.Parse(writerSchemaJson);
+
+                    datumReader = new SpecificReader<T>(writerSchema, ReaderSchema);
+                    lock (datumReaderLock)
```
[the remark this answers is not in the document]
mhowlett: yes, was aware of that - it's benign though, since the datumReader will just get overwritten and there is no resource leak since SpecificReader is not disposable. Still, I'll put an if statement here to prevent the race condition since that is even simpler than writing a note explaining that it doesn't matter.

src/Confluent.Kafka.Avro/AvroSerializer.cs
mhowlett: they are documented right here :-). but good point, i didn't document because it wasn't clear exactly where to do so. we need better docs for this client, and they need to include this. Note that both properties are noted in the example which will go a long way, since most people will look at that.

good catch with the plural.

I think the properties shouldn't be schema.registry.[prop], since they relate to the schema registry client. 

I think better is `avro.serializer.auto.register.schemas`, since also they only relate to the serializer, no deserializer, and that makes that explicit, which is a good thing.

src/Confluent.Kafka.Avro/AvroSerializer.cs
```diff
@@ -0,0 +1,294 @@
...
+            else if (writerType.Equals(typeof(byte[])))
+            {
+                // Note: It would arguably be better to make this a union with null, to
+                // exactly match the .NET byte[] type, however we don't for consistency
+                // with the Java avro serializer.
+                WriterSchema = Avro.Schema.Parse("bytes");
+            }
+            else
+            {
+                throw new InvalidOperationException(
```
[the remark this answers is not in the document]
mhowlett: dito `ArgumentException`..

src/Confluent.Kafka.Avro/AvroSerializer.cs
```diff
@@ -0,0 +1,294 @@
...
+        private bool disposeClientOnDispose;
+
+        private string writerSchemaString { get; set; }
+
+        private const string InitialBufferSizePropertyName = "avro.buffer.bytes";
+
+        private const string AutoRegisterSchemaPropertyName = "avro.auto.register.schema";
+
+        /// <summary>
+        ///     The schema id corresponding to the schema of T in big-endian
```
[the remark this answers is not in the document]
mhowlett: also, something I missed - the method here should be `HostToNetworkOrder` not `NetworkToHostOrder`.

src/Confluent.Kafka.Avro/AvroSerializer.cs
```diff
@@ -0,0 +1,294 @@
...
+                {
+                    SchemaId = SchemaRegistryClient.RegisterSchemaAsync(subject, writerSchemaString).Result;
+                }
+                else
+                {
+                    SchemaId = SchemaRegistryClient.GetSchemaIdAsync(subject, writerSchemaString).Result;
+                }
+
+                schemaIdBigEndian = IPAddress.NetworkToHostOrder(SchemaId.Value);
+                topicsRegistered.Add(topic);
```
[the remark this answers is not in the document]
mhowlett: no - the serializer is implicitly a key or value one (as tracked by the IsKey property), using subject here would be extra work.

src/Confluent.Kafka.Avro/AvroSerializer.cs
```diff
@@ -0,0 +1,294 @@
...
+            if (srConfig.Count() != 0)
+            {
+                if (SchemaRegistryClient != null)
+                {
+                    throw new ArgumentException($"{keyOrValue} AvroSerializer schema registry client was configured via both the constructor and configuration parameters.");
+                }
+ 
+                disposeClientOnDispose = true;
+                SchemaRegistryClient = new CachedSchemaRegistryClient(config);
+            }
```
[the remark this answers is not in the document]
mhowlett: good point, forgot about that, though it belongs in `CachedSchemaRegistryClient`, not here.

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -0,0 +1,202 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using System.Linq;
+using System;
+
+
+namespace Confluent.SchemaRegistry
+{
+    /// <summary>
+    ///     A Schema Registry client that caches request responses.
```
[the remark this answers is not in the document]
mhowlett: it caches various responses from requests to the schema registry client. I changed this to "A caching Schema Registry client".

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -0,0 +1,202 @@
...
+        private IRestService restService;
+        private readonly int identityMapCapacity;
+        private readonly Dictionary<int, string> schemaById = new Dictionary<int, string>();
+        private readonly Dictionary<string /*subject*/, Dictionary<string, int>> idBySchemaBySubject = new Dictionary<string, Dictionary<string, int>>();
+        private readonly Dictionary<string /*subject*/, Dictionary<int, string>> schemaByVersionBySubject = new Dictionary<string, Dictionary<int, string>>();
+        
+        /// <summary>
+        ///     The default timeout value for Schema Registry REST API calls.
+        /// </summary>
+        public const int DefaultTimeout = 10000;
```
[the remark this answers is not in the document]
mhowlett: sounds good.

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -0,0 +1,202 @@
...
+
+        /// <summary>
+        ///     The default maximum capacity of the local schema cache.
+        /// </summary>
+        public const int DefaultMaxCapacity = 1024;
```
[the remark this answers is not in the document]
mhowlett: not going to disagree

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -0,0 +1,202 @@
...
+            }
+
+            var schemaRegistryUrisMaybe = config.Where(prop => prop.Key.ToLower() == "schema.registry.url").FirstOrDefault();
+            if (schemaRegistryUrisMaybe.Value == null)
+            {
+                throw new ArgumentException("schema.registry.url configuration property must be specified.");
+            }
+            var schemaRegistryUris = (string)schemaRegistryUrisMaybe.Value;
+
+            var timeoutMsMaybe = config.Where(prop => prop.Key.ToLower() == "schema.registry.timeout.ms").FirstOrDefault();
```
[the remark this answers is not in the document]
mhowlett: good idea

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -0,0 +1,202 @@
...
+            var timeoutMs = timeoutMsMaybe.Value == null ? DefaultTimeout : (int)timeoutMsMaybe.Value;
+
+            var identityMapCapacityMaybe = config.Where(prop => prop.Key.ToLower() == "schema.registry.cache.capacity").FirstOrDefault();
```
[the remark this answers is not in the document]
mhowlett: good idea

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -0,0 +1,202 @@
...
+            this.restService = new RestService(schemaRegistryUris, timeoutMs);
+        }
+
+        /// <remarks>
+        ///     Ensure memory doesn't explode in the case of incorrect usage.
+        /// </remarks>
+        private void CheckIfCacheFull()
+        {
+            if (
+                this.schemaById.Count +
```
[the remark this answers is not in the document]
mhowlett: i'd agree with that.

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -0,0 +1,202 @@
...
+        /// </remarks>
+        private void CheckIfCacheFull()
+        {
+            if (
+                this.schemaById.Count + 
+                this.schemaByVersionBySubject.Sum(x => x.Value.Count) +
+                this.idBySchemaBySubject.Sum(x => x.Value.Count)
+                    >= identityMapCapacity)
+            {
+                throw new OutOfMemoryException("Local schema cache maximum capacity exceeded");
```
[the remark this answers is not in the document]
mhowlett: I put in a todo. I think if the cache fills up this will most likely be unexpected, in which case failing fast is probably actually preferable to evicting. anyway, it's more work, and the java client just throws an exception, so this is good enough for an mvp.

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -0,0 +1,202 @@
...
+                schemaById[schemaId] = schema;
+            }
+
+            return schemaId;
+        }
+
+        /// <include file='include_docs.xml' path='API/Member[@name="ISchemaRegistryClient_RegisterSchemaAsync"]/*' />
+        public async Task<int> RegisterSchemaAsync(string subject, string schema)
+        {
+            CheckIfCacheFull();
```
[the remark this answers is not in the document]
mhowlett: right you are

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -0,0 +1,202 @@
...
+                schemaById[schemaId] = schema;
+            }
+
+            return schemaId;
+        }
+
+        /// <include file='include_docs.xml' path='API/Member[@name="ISchemaRegistryClient_GetSchemaAsync"]/*' />
+        public async Task<string> GetSchemaAsync(int id)
+        {
+            CheckIfCacheFull();
```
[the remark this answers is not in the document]
mhowlett: right again! you're on a roll.

src/Confluent.SchemaRegistry/Confluent.SchemaRegistry.csproj
mhowlett: `Conflient.SchemaRegistry` doesn't depend on `Apache.Avro`, so it's compatible with .net standard. we might as well target it so .NET Core people can use at least this.

src/Confluent.SchemaRegistry/Rest/DataContracts/Schema.cs
mhowlett: removed the word 'Globally'.

src/Confluent.SchemaRegistry/Rest/DataContracts/Schema.cs
mhowlett: yep, this is silly.

src/Confluent.SchemaRegistry/Rest/DataContracts/Schema.cs
mhowlett: Good point. The spec calls for an `ArgumentException` in this case, so i added that.

src/Confluent.SchemaRegistry/Rest/DataContracts/Schema.cs
mhowlett: changed to avoid using the terminology 'value'.

src/Confluent.SchemaRegistry/Rest/RestService.cs
mhowlett: ok, but this whole situation makes me sad.

mhowlett: btw. we hardly ever need grep with our fancy C# editors.

src/Confluent.SchemaRegistry/Rest/RestService.cs
mhowlett: it does? it iterates through clients.Count times and modulo arithmetic is used in the loop.

src/Confluent.SchemaRegistry/Rest/RestService.cs
mhowlett: whoa, that's due to a bad refactor. refactored again, please check this logic again.

test/Confluent.Kafka.Avro.IntegrationTests/Tests/ConsumeIncompatibleTypes.cs
mhowlett: this is a new topic, so it's always going to be 0, right?

test/Confluent.Kafka.Avro.IntegrationTests/test.parameters.json
mhowlett: the test framework requires them to be the same as the method parameter names in the test methods (so they can't include .'s).

test/Confluent.Kafka.Avro.IntegrationTests/Tests/ConsumeIncompatibleTypes.cs
mhowlett: the topic is different each time the test is run.

mhowlett:
@manuel-zulian - thanks for the input. Hopefully the avro library is high enough quality that it's not going to cause us too many problems .... 

Feel free to open a PR against the `confluent-fork` branch of https://github.com/confluentinc/avro 

Note that I'm moving onto some other priorities after this PR is merged (also on the .net client), so I likely won't get to merging difficult-to-review PRs immediately, but if you keep the changes small and incremental i'll get to them quickly.

manuel-zulian:
@mhowlett I've started a clean branch for the logical types, I will try to keep things clear. At some point I will start a pull request.
However, you might want to have a look only at this commit: https://github.com/manuel-zulian/avro/commit/2537aecd2c58573cef385c109ebc605473a8f5bf
This fixes the small problem I was talking about with the Json library version bump. Essentially, from version 3.5 to v9 the method ToString was changed and is no longer storing the escaped string in the properties, but only the pure string. If I'm correct (double check on your side as well) this change should also have fixed two tests in the AliasTest file (Avro project). You might want to include just this change in the first release of the avro support for Kafka.

mhowlett:
awesome, thanks @manuel-zulian

src/Confluent.Kafka.Avro/Confluent.Kafka.Avro.csproj
```diff
@@ -0,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net452;</TargetFrameworks>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Confluent.Apache.Avro" Version="1.0" />
```
edenhill: Dont forget to update this version to reflect the avro version

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -0,0 +1,215 @@
...
+            // contains very few elements and the schema string passed in is always
+            // the same instance.
+
+            if (!idBySchema.TryGetValue(schema, out int schemaId))
+            {
+                CheckIfCacheFull();
+
+                schemaId = (await restService.CheckSchemaAsync(subject, schema, true)).Id;
+                idBySchema[schema] = schemaId;
+                schemaById[schemaId] = schema;
```
edenhill: this stuff needs to be protected by a lock, right?

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -0,0 +1,215 @@
...
+            }
+
+            if (!schemaByVersion.TryGetValue(version, out string schemaString))
+            {
+                CheckIfCacheFull();
+
+                var schema = await restService.GetSchemaAsync(subject, version).ConfigureAwait(false);
+                schemaString = schema.SchemaString;
+                schemaByVersion[version] = schemaString;
+                schemaById[schema.Id] = schemaString;
```
edenhill: dito el locko

src/Confluent.SchemaRegistry/Rest/RestService.cs
```diff
@@ -0,0 +1,306 @@
...
+
+                try
+                {
+                    response = await clients[clientIndex]
+                            .SendAsync(request).ConfigureAwait(false);
+
+                    // In the case of an internal server error, try another server 
+                    //   (reason could be e.g. "error while forwarding the request to the master")
+                    // 4xx errors should not be retried (these are conclusive)
+                    if (response.StatusCode == HttpStatusCode.InternalServerError)
```
edenhill: check for 500 range rather than 500 specifiicificialciacally

src/Confluent.SchemaRegistry/Rest/RestService.cs
```diff
@@ -0,0 +1,306 @@
...
+                        catch
+                        {
+                            aggregatedErrorMessage += $"[{clients[clientIndex].BaseAddress}] {response.StatusCode}";
+                        }
+
+                        aggregatedErrorMessage += $"[{clients[clientIndex].BaseAddress}] {response.StatusCode} {errorCode} {message}";
+                    }
+                    else 
+                    {
+                        gotOkAnswer = true;
```
edenhill: is !500 always ok? What about 4xx?

test/Confluent.Kafka.Avro.UnitTests/Configuration.cs
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        [Fact]
+        public void SerializerConfigure()
+        {
+            var avroSerializer = new AvroSerializer<int>(schemaRegistryClient);
+
+            var config = new Dictionary<string, object>
+            {
+                { "avro.serializer.buffer.bytes", 42 },
```
edenhill: is there a test that checks that growing beyond this number actually works?
e.g. set it low and serialize something bigish

src/Confluent.Kafka.Avro/AvroDeserializer.cs
```diff
@@ -0,0 +1,245 @@
...
+                {
+                    // may change in the future.
+                    throw new InvalidDataException($"magic byte should be 0, not {magicByte}");
+                }
+                var writerIdBigEndian = reader.ReadInt32();
+                var writerId = IPAddress.NetworkToHostOrder(writerIdBigEndian);
+
+                DatumReader<T> datumReader = null;
+                // TODO: A r/w lock would be better, but the improvement probably negligible.
+                lock (datumReaderLock)
```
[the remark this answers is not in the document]
mhowlett: yep, silly me.

src/Confluent.Kafka.Avro/AvroSerializer.cs
```diff
@@ -0,0 +1,309 @@
...
+        ///     configuration property collection:
+        ///     
+        ///     schema.registry.url (required) - A comma-separated list of URLs for schema registry 
+        ///         instances that can be used to register or lookup schemas.
+        ///                           
+        ///     schema.registry.connection.timeout.ms (default: 30000) - Timeout for requests to 
+        ///         Confluent Schema Registry.
+        ///     
+        ///     schema.registry.max.cached.schemas (default: 1000) - The maximum number of schemas 
+        ///         to cache locally.
```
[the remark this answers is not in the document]
mhowlett: LRU is a little tricky here as we have 3 separate dictionaries (two with sub-dictionaries). I'll put a todo to make this better, and for now just evict everything if the cache fills up. I'm actually still not convinced fail fast is worse - i'd like to understand more where this trips people up in the Java client in practice before fully admitting defeat here.

src/Confluent.Kafka.Avro/Confluent.Kafka.Avro.csproj
mhowlett: done

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
mhowlett: yep, you're right.

test/Confluent.Kafka.Avro.UnitTests/Configuration.cs
mhowlett: There isn't but there is no risk, it just leverages `MemoryStream` in the implementation which works.

src/Confluent.SchemaRegistry/Rest/RestService.cs
mhowlett: yes, I think that makes sense.

src/Confluent.SchemaRegistry/Rest/RestService.cs
mhowlett: The java client considers them all the same and I'll follow that here too as you suggest. I think my reasoning was that a 4xx error implied a valid response from the server that would be mimicked by all servers. I actually still think that is probably right, though I'm not sure.

mhowlett: I retained the separate 4xx functionality in the end when SR responds with valid content, since I think it's potentially useful for the caller to be able to differentiate between the two types of errors. I reworked the logic however, and this should be re-reviewed.

src/Confluent.SchemaRegistry/Rest/RestService.cs
```diff
@@ -93,78 +92,61 @@ public RestService(string schemaRegistryUrl, int timeoutMs)
                     response = await clients[clientIndex]
                             .SendAsync(request).ConfigureAwait(false);
 
-                    // In the case of an internal server error, try another server 
-                    //   (reason could be e.g. "error while forwarding the request to the master")
-                    // 4xx errors should not be retried (these are conclusive)
-                    if (response.StatusCode == HttpStatusCode.InternalServerError)
+                    if (response.StatusCode == HttpStatusCode.OK ||
+                        response.StatusCode == HttpStatusCode.NoContent)
```
edenhill: que NoContent?

mhowlett: matching the java client

David-V-Spencer:
@mhowlett Just wanted to clarify what is supported in the final release as we are depending on .NET Standard/Core 2.0?

Reviewing the history here it seemed that initially this PR would only support .NET Framework but midway through support for .NET Standard/Core 2.0 also included. I was reading rather quickly so I could have easily missed some detail that could have clarified it.

mhowlett:
@David-V-Spencer - for .NET Core to be supported, the Apache Avro library just needs to be updated to target .NET Standard. I can't guarantee we'll put effort into doing that in the near term, though if I take a look and it's easy, I'll do it.

jeddawson:
@mhowlett I just landed in this thread because I'm hoping to take advantage of the great work that has been done with the Avro Serializer here, but using .NET Core 2.0. I started checking out the Apache Avro library with the intention of starting work on updating the target, but discovered [there is already a PR](https://github.com/apache/avro/pull/261) that updates the target to .NET Standard.

For now I'm cloning from the [source branch](https://github.com/moljac/avro/tree/master-based-dotnet-standard) and building locally to get things running. If I run into issues or have success, I'll report my experience back here just in case it's helpful.

mhowlett:
Thanks @jeddawson - I don't see any changes to the code in that PR (just to project files etc), so i'm guessing (though I could be wrong), it's not going to get us all the way there since I did previously look into making the library target .net standard and based on that experimentation believe some updates to the code are going to be required.

mhowlett:
@LorenzoRompianesi @joelduvall @thedavejay @noderat @neekgreen @mjr5749 @unfw @David-V-Spencer @jeddawson - just a heads up that Confluent.Kafka.Avro 0.11.3-ci-329 targets netstandard2.0 (in addition to net452), and seems to be working well in my own testing. Also, i've tested the avrogen tool from: https://github.com/confluentinc/avro/pull/5 on netcoreapp2.0 by compiling a very complex schema and it's working well, including on mac os / linux (though i'm only able to build it on windows presently). feel free to try out this functionality and provide feedback on any problems. No guarantees about stability or release date at this point.

lixiandai:
@mhowlett we had a question related schema and schema-less data
so if we have a production env which was keeping producing the data without registry feature supported, then we want to migrate those data to our latest production env which has the schema registered, what is the best practices to those data migration?

what i tested out so far is our latest consumer is not able to consume any data without schema at all although the data's members did not change at all.

Thanks.

mhowlett:
@lixiandai - you probably want a new topic for the new data and migrate things over one by one.

PR #359, opened 2017-11-05
added IgnoreDeserializer

mhowlett:
@edenhill - this PR includes a number of quick wins for things that commonly trip people up. it needs to target a new branch though, because we don't want the examples to be incompatible with the current release. what version are we moving to next, 0.12.0.x?

- Added an 'Ignore' deserializer - people don't expect the null deserializer to throw an exception if it receives a non-null value and it confuses them. I've retained the null deserializer (it's important), but changed the examples to use the new `IgnoreDeseiralizer` instead.
- Added `OnConsumeError` handlers to the examples. People will cut and paste from the examples, and it's important they know this is something they should be handling.
- Consumers and Producers now throw an exception if the same instance of a serializer / deserializer is passed into the constructor for both key and value. this is an error (they are initialized separately by the producer/consumer), and it's not clear from the API that it is.

### DIFF
--- a/examples/SimpleConsumer/Program.cs
+++ b/examples/SimpleConsumer/Program.cs
@@ -42,9 +42,11 @@ namespace Confluent.Kafka.Examples.SimpleConsumer
             {
                 consumer.Assign(new List<TopicPartitionOffset> { new TopicPartitionOffset(topics.First(), 0, 0) });
                 
+                // Raised on critical errors, e.g. connection failures or all brokers down.
                 consumer.OnError += (_, error)
                     => Console.WriteLine($"Error: {error}");
 
+                // Raised on deserialization errors or when a consumed message has an error != NoError.
                 consumer.OnConsumeError += (_, error)
                     => Console.WriteLine($"Consume error: {error}");
                 
--- a/examples/SimpleConsumer/Program.cs
+++ b/examples/SimpleConsumer/Program.cs
@@ -38,13 +38,19 @@ namespace Confluent.Kafka.Examples.SimpleConsumer
                 { "bootstrap.servers", brokerList }
             };
 
-            using (var consumer = new Consumer<Null, string>(config, null, new StringDeserializer(Encoding.UTF8)))
+            using (var consumer = new Consumer<Ignore, string>(config, null, new StringDeserializer(Encoding.UTF8)))
             {
                 consumer.Assign(new List<TopicPartitionOffset> { new TopicPartitionOffset(topics.First(), 0, 0) });
+                
+                consumer.OnError += (_, error)
+                    => Console.WriteLine($"Error: {error}");
 
+                consumer.OnConsumeError += (_, error)
+                    => Console.WriteLine($"Consume error: {error}");
+                
                 while (true)
                 {
-                    Message<Null, string> msg;
+                    Message<Ignore, string> msg;
                     if (consumer.Consume(out msg, TimeSpan.FromSeconds(1)))
                     {
                         Console.WriteLine($"Topic: {msg.Topic} Partition: {msg.Partition} Offset: {msg.Offset} {msg.Value}");
--- a/src/Confluent.Kafka/Serialization/IgnoreDeserializer.cs
+++ b/src/Confluent.Kafka/Serialization/IgnoreDeserializer.cs
@@ -1,0 +1,49 @@
+// Copyright 2016-2017 Confluent Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Refer to LICENSE for more information.
+
+using System;
+using System.Collections.Generic;
+
+
+namespace Confluent.Kafka.Serialization
+{
+    /// <summary>
+    ///     A 'deserializer' that returns null regardless of the input data.
+    /// </summary>
+    public class IgnoreDeserializer : IDeserializer<Ignore>
+    {
+        /// <summary>
+        ///     'Deserializes' any data to a null value.
+        /// </summary>
+        /// <param name="data">
+        ///     The data to deserialize.
+        /// </param>        
+        /// <param name="topic">
+        ///     The topic associated with the data (ignored by this deserializer).
+        /// </param>
+        /// <returns>
+        ///     null
+        /// </returns>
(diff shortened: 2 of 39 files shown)


examples/SimpleConsumer/Program.cs
```diff
@@ -38,13 +38,19 @@ public static void Main(string[] args)
                 { "bootstrap.servers", brokerList }
             };
 
-            using (var consumer = new Consumer<Null, string>(config, null, new StringDeserializer(Encoding.UTF8)))
+            using (var consumer = new Consumer<Ignore, string>(config, null, new StringDeserializer(Encoding.UTF8)))
             {
                 consumer.Assign(new List<TopicPartitionOffset> { new TopicPartitionOffset(topics.First(), 0, 0) });
+
```
edenhill: It'd be nice with a comment that tells the user what the difference is between these two. Global errors (disconnects, etc) vs consume stream errors (failure to commit, etc).

src/Confluent.Kafka/Serialization/IgnoreDeserializer.cs
```diff
@@ -0,0 +1,49 @@
...
+// Refer to LICENSE for more information.
+
+using System;
+using System.Collections.Generic;
+
+
+namespace Confluent.Kafka.Serialization
+{
+    /// <summary>
+    ///     A 'deserializer' that returns null regardless of the input data.
```
edenhill: Why the fnutts? (quotes)

src/Confluent.Kafka/Serialization/IgnoreDeserializer.cs
```diff
@@ -0,0 +1,49 @@
...
+    /// </summary>
+    public class IgnoreDeserializer : IDeserializer<Ignore>
+    {
+        /// <summary>
+        ///     'Deserializes' any data to a null value.
```
edenhill: dito

src/Confluent.Kafka/Serialization/IgnoreDeserializer.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// </param>
+        /// <returns>
+        ///     null
+        /// </returns>
+        public Ignore Deserialize(string topic, byte[] data)
+        {
+            return null;
+        }
+
+        /// <include file='../include_docs.xml' path='API/Member[@name="IDeserializer_Configure"]/*' />
```
edenhill: Does this serdes take any config at all?

test/Confluent.Kafka.IntegrationTests/Tests/Ignore.cs
```diff
@@ -0,0 +1,82 @@
...
+using System.Collections.Generic;
+using Xunit;
+
+
+namespace Confluent.Kafka.IntegrationTests
+{
+    public static partial class Tests
+    {
+        /// <summary>
+        ///     Test that the ignore deserialier behaves as expected.
```
edenhill: Document that this requires a single-partition topic.

test/Confluent.Kafka.UnitTests/Consumer.cs
```diff
@@ -40,6 +40,18 @@ public void Constuctor()
...
+                var validConfig = new Dictionary<string, object>
+                {
+                    { "bootstrap.servers", "localhost:9092" },
+                    { "group.id", "my-group" }
+                };
+                var deserializer = new StringDeserializer(Encoding.UTF8);
+                var c = new Consumer<string, string>(validConfig, deserializer, deserializer); 
+            });
+            Assert.True(e.Message.Contains("must not be the same object"));
+
```
edenhill: Add unit test for IgnoreSerializer.

test/Confluent.Kafka.IntegrationTests/Tests/Ignore.cs
```diff
@@ -0,0 +1,82 @@
...
+
+                Assert.True(consumer.Consume(out msg, TimeSpan.FromMinutes(1)));
+                Assert.NotNull(msg);
+                Assert.Null(msg.Key);
+                Assert.Null(msg.Value);
+            }
+        }
+
+    }
+}
```
edenhill: Also test the combination of Ignore,SomeNonIgnoreSerDes

src/Confluent.Kafka/Serialization/IgnoreDeserializer.cs
mhowlett: because it doesn't deserialize anything.

src/Confluent.Kafka/Serialization/IgnoreDeserializer.cs
mhowlett: all serdes are passed the config, and can take properties out of them if they want. this one does not.

test/Confluent.Kafka.UnitTests/Consumer.cs
mhowlett: didn't make an `IgnoreSerializer` - didn't think it would ever be useful (so more confusing to have it).

test/Confluent.Kafka.IntegrationTests/Tests/Ignore.cs
mhowlett: ok

examples/SimpleConsumer/Program.cs
mhowlett: good idea.

src/Confluent.Kafka/Serialization/IgnoreDeserializer.cs
mhowlett: I like having the quotes because they are informational - there is a good reason they are there and it makes people wonder why.

src/Confluent.Kafka/Serialization/IgnoreDeserializer.cs
mhowlett: ditto

test/Confluent.Kafka.IntegrationTests/Tests/Ignore.cs
mhowlett: (also elsewhere, via a argument rename)

mhowlett: good idea

test/Confluent.Kafka.UnitTests/Consumer.cs
```diff
@@ -40,6 +40,18 @@ public void Constuctor()
...
 
+            e = Assert.Throws<ArgumentException>(() => 
+            {
+                var validConfig = new Dictionary<string, object>
+                {
+                    { "bootstrap.servers", "localhost:9092" },
+                    { "group.id", "my-group" }
+                };
+                var deserializer = new StringDeserializer(Encoding.UTF8);
+                var c = new Consumer<string, string>(validConfig, deserializer, deserializer);
```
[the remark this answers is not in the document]
mhowlett: I put this check in because I saw you doing this in the avro PR... It's an error because the Consumer calls Configure on both serializers, and furthermore this call takes a bool parameter indicating whether it's a value of key serializer. Most serializers do nothing in the Configure method, so it will work fine, but in the avro case it won't.

PR #377, opened 2017-11-23
store_offsets fixes

mhowlett:
@edenhill

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -432,7 +432,8 @@ namespace Confluent.Kafka
             => consumer.Unassign();
 
         /// <summary>
-        ///     Store offsets for one or more partitions.
+        ///     Store offsets for a single partition based on the topic/partition/offset
+        ///     of a message.
         ///     
         ///     The offset will be committed (written) to the offset store according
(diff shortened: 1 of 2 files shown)


src/Confluent.Kafka/Consumer.cs
```diff
@@ -446,8 +447,8 @@ public void Unassign()
         /// <returns>
         ///     Current stored offset or a partition specific error.
         /// </returns>
-        public List<TopicPartitionOffsetError> StoreOffset(Message<TKey, TValue> message)
-            => consumer.StoreOffsets(new[] { new TopicPartitionOffset(message.TopicPartition, message.Offset + 1) });
+        public TopicPartitionOffsetError StoreOffset(Message<TKey, TValue> message)
+            => consumer.StoreOffsets(new[] { new TopicPartitionOffset(message.TopicPartition, message.Offset + 1) })[0];
```
edenhill: Is it safe to assume that it always returns at least one so the [0] deref can be used?

edenhill:
LGTM

src/Confluent.Kafka/Consumer.cs
mhowlett: yep - if it's not then it'd be an internal error and something to fix somewhere else.

PR #401, opened 2018-01-04
Optimize topic handle lookup

szKarlen:
At the moment `getKafkaTopicHandle()` method relies on a custom dictionary-like class `SafeDictionary`, which always grows while adding an item and lacks `StringComparer`.

It's much better to use `System.Collections.Concurrent.ConcurrentDictionary` instead.


szKarlen:
[clabot:check]

mhowlett:
This looks great, as does your other one #403.
In this one, you could get rid of the SafeDictionary file entirely, that's the only thing I can see after a quick look through.
I'll review them carefully before the next release.
thanks @szKarlen !

szKarlen:
Thanks @mhowlett !

I've pushed an additional commit with the `SafeDictionary` file removal.

src/Confluent.Kafka/Producer.cs
```diff
@@ -88,38 +88,15 @@ private void LogCallback(IntPtr rk, int level, string fac, string buf)
...
-            // note: there is a possible (benign) race condition here - topicHandle could have already
-            // been created for the topic (and possibly added to topicHandles). If the topicHandle has
-            // already been created, rdkafka will return it and not create another. the call to rdkafka
-            // is threadsafe.
-            var topicHandle = kafkaHandle.Topic(topic, configPtr);
-            topicHandles.Add(topic, topicHandle);
-
-            return topicHandle;
-        }
+        private SafeTopicHandle getKafkaTopicHandle(string topic) => topicHandles.GetOrAdd(topic, topicHandlerFactory);
```
mhowlett: I don't think there is any benefit to using this variant of `GetOrAdd`? If that's right, I think using the other variant is better as it reduces the levels of indirection making the code more straightforward.

other than this, this PR looks good to me, and I'll merge after I've verified the integration tests pass (which they currently don't due to a problem unrelated to you).

szKarlen: Currently `Producer.topicConfig` gets captured in a lambda, which forces the compiler to emit a new delegate allocation at every `getKafkaTopicHandle` call.
That's why we have to cache the callback manually.

mhowlett:
@szKarlen - I made some changes to `getKafkaTopicHandle` and `Dispose` on master since this PR and rebased. I'll merge this to master after you've verified my changes.

src/Confluent.Kafka/Producer.cs
```diff
@@ -315,6 +322,22 @@ public Producer(IEnumerable<KeyValuePair<string, object>> config, bool manualPol
                 callbackCts = new CancellationTokenSource();
                 callbackTask = StartPollTask(callbackCts.Token);
             }
+
+            // note: ConcurrentDictionary.GetorAdd() method is not atomic
+            this.topicHandlerFactory = (string topicName) =>
+            {
+                if (topicHandles.ContainsKey(topicName))
```
szKarlen: This additional check is unnecessary. factory handlers will be for a topic handle creation always.
The rest merge looks good.

szKarlen:
@mhowlett I pushed a small fix (0c5af58cd3414617f57acfa156689e69968cb0fb). 
The rest looks good. Thanks!

mhowlett:
oh yeah, whoops! thanks.

PR #410, opened 2018-01-18
set produce.offset.report to true by default

mhowlett:
@edenhill

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -824,12 +824,11 @@ namespace Confluent.Kafka
             if (defaultTopicConfig != null)
             {
                 var topicConfigHandle = SafeTopicConfigHandle.Create();
-                if (config != null)
-                {
-                    defaultTopicConfig.ToList().ForEach((kvp) => { topicConfigHandle.Set(kvp.Key, kvp.Value.ToString()); });
-                }
+                defaultTopicConfig.ToList().ForEach(
+                    (kvp) => { topicConfigHandle.Set(kvp.Key, kvp.Value.ToString()); }
+                );
                 LibRdKafka.conf_set_default_topic_conf(configPtr, topicConfigHandle.DangerousGetHandle());
-                topicConfigHandle.SetHandleAsInvalid(); // topic config object is no longer useable.
+                topicConfigHandle.SetHandleAsInvalid();
             }
 
             LibRdKafka.conf_set_error_cb(configPtr, errorDelegate);
--- a/src/Confluent.Kafka/Producer.cs
+++ b/src/Confluent.Kafka/Producer.cs
@@ -40,7 +40,7 @@ namespace Confluent.Kafka
         internal const int RD_KAFKA_PARTITION_UA = -1;
         internal const long RD_KAFKA_NO_TIMESTAMP = 0;
 
-        private IEnumerable<KeyValuePair<string, object>> topicConfig;
+        private IEnumerable<KeyValuePair<string, object>> defaultTopicConfig;
 
         private SafeDictionary<string, SafeTopicHandle> topicHandles
             = new SafeDictionary<string, SafeTopicHandle>();
@@ -102,9 +102,9 @@ namespace Confluent.Kafka
             }
 
             var topicConfigHandle = SafeTopicConfigHandle.Create();
-            if (topicConfig != null)
+            if (defaultTopicConfig != null)
             {
-                topicConfig
+                defaultTopicConfig
                     .ToList()
                     .ForEach((kvp) => { topicConfigHandle.Set(kvp.Key, kvp.Value.ToString()); });
             }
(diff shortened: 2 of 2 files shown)


src/Confluent.Kafka/Consumer.cs
```diff
@@ -824,12 +824,11 @@ public Consumer(IEnumerable<KeyValuePair<string, object>> config)
             if (defaultTopicConfig != null)
             {
                 var topicConfigHandle = SafeTopicConfigHandle.Create();
-                if (config != null)
```
mhowlett: this check is pointless.

src/Confluent.Kafka/Producer.cs
```diff
@@ -302,6 +302,31 @@ public Producer(IEnumerable<KeyValuePair<string, object>> config, bool manualPol
...
+            {
+                var topicConfigHandle = SafeTopicConfigHandle.Create();
+                this.defaultTopicConfig.ToList().ForEach(
+                    (kvp) => { topicConfigHandle.Set(kvp.Key, kvp.Value.ToString()); }
+                );
+                if (this.defaultTopicConfig.FirstOrDefault(prop => prop.Key == "produce.offset.report").Value == null)
+                {
+                    topicConfigHandle.Set("produce.offset.report", "true");
+                }
+                LibRdKafka.conf_set_default_topic_conf(configPtr, topicConfigHandle.DangerousGetHandle());
```
mhowlett: currently, set_default_topic_conf is not called at all in the producer. Maybe I took this out since these properties can all be set at the top level now... I think I vaguely remember that. however the current code is not good since it just ignores "default.topic.config" without warning. I think it's best to leave the default.topic.config functionality in for now and warn it's depreciated in docs (which are non-existent). Then remove it at some point in the future, throwing an exception if it's used.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -824,12 +824,11 @@ public Consumer(IEnumerable<KeyValuePair<string, object>> config)
...
-                {
-                    defaultTopicConfig.ToList().ForEach((kvp) => { topicConfigHandle.Set(kvp.Key, kvp.Value.ToString()); });
-                }
+                defaultTopicConfig.ToList().ForEach(
+                    (kvp) => { topicConfigHandle.Set(kvp.Key, kvp.Value.ToString()); }
+                );
                 LibRdKafka.conf_set_default_topic_conf(configPtr, topicConfigHandle.DangerousGetHandle());
-                topicConfigHandle.SetHandleAsInvalid(); // topic config object is no longer useable.
```
edenhill: That comment wasn't all that bad though.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -824,12 +824,11 @@ public Consumer(IEnumerable<KeyValuePair<string, object>> config)
...
             {
                 var topicConfigHandle = SafeTopicConfigHandle.Create();
-                if (config != null)
-                {
-                    defaultTopicConfig.ToList().ForEach((kvp) => { topicConfigHandle.Set(kvp.Key, kvp.Value.ToString()); });
-                }
+                defaultTopicConfig.ToList().ForEach(
+                    (kvp) => { topicConfigHandle.Set(kvp.Key, kvp.Value.ToString()); }
+                );
                 LibRdKafka.conf_set_default_topic_conf(configPtr, topicConfigHandle.DangerousGetHandle());
```
edenhill: This will overwrite any topic configurations set by fallthru on the main config, i.e.,  this will overwrite the implicitly created default topic config (if any).
I think it is better to set the key-vals from defaultTopicConfig directly on the configPtr instead.

src/Confluent.Kafka/Producer.cs
edenhill: See earlier comment on not using set_default_topic_conf().

src/Confluent.Kafka/Producer.cs
```diff
@@ -284,6 +271,26 @@ public Producer(IEnumerable<KeyValuePair<string, object>> config, bool manualPol
...
+            }
+
+            // Note: changing the default value of produce.offset.report at the binding level is less than
+            // ideal since it means the librdkafka configuration docs will no longer completely match the 
+            // .NET client. The default should probably be changed in librdkafka as well.
+            if (config.FirstOrDefault(prop => prop.Key == "produce.offset.report").Value == null &&
+                (defaultTopicConfig == null ||
+                 defaultTopicConfig.FirstOrDefault(prop => prop.Key == "produce.offset.report").Value == null))
+            {
+                configHandle.Set("produce.offset.report", "true");
```
edenhill: Isn't it easier to just set this (and other defaults) at the creation of the configHandle?

Inverse priority order:
 * librdkafka default settings (produce.offset.report=false)
 * dotnet default settings (produce.offset.report=true)  <== set here
 * global explicit config
 * default.topic.config explicit config

mhowlett: Not 100% sure I understand exactly what you're saying, but I simplified as a result of your comment. A modifiedConfig is now prepared, then all properties set in one loop (this is logically simpler).

Issue #412, opened 2018-01-18
Intermittent AggregateException exception on Producer Dispose


mhowlett:
Description
===========

I have seen the below exception thrown intermittently whilst running integration tests.

Failed   Confluent.Kafka.IntegrationTests.Tests.TypedProducer_ClosedHandle(bootstrapServers: "localhost:9092", topic: "test-topic-1", partitionedTopic: "test-topic-2")
Error Message:
 System.AggregateException : One or more errors occurred. (A task was canceled.)
---- System.Threading.Tasks.TaskCanceledException : A task was canceled.
Stack Trace:
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Task.Wait()
   at Confluent.Kafka.Producer.Dispose() in /git/confluent-kafka-dotnet/src/Confluent.Kafka/Producer.cs:line 688
   at Confluent.Kafka.Producer`2.Dispose() in /git/confluent-kafka-dotnet/src/Confluent.Kafka/Producer.cs:line 1325
   at Confluent.Kafka.IntegrationTests.Tests.TypedProducer_ClosedHandle(String bootstrapServers, String topic, String partitionedTopic) in /git/confluent-kafka-dotnet/test/Confluent.Kafka.IntegrationTests/Tests/ClosedHandle.cs:line 74
----- Inner Stack Trace -----

mhowlett:
resolved by #413

PR #413, opened 2018-01-18
use ThrowIfCancellationRequested to exit producer poll loop

mhowlett:
@edenhill - fix for #412. 

This new functionality follows the example in: https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/task-cancellation 

I don't actually understand why the existing way of doing this is causing intermittent problems, however it doesn't matter as the new way is valid, and avoids the problem.

### DIFF
--- a/src/Confluent.Kafka/Producer.cs
+++ b/src/Confluent.Kafka/Producer.cs
@@ -52,9 +52,10 @@ namespace Confluent.Kafka
         private Task StartPollTask(CancellationToken ct)
             => Task.Factory.StartNew(() =>
                 {
-                    while (!ct.IsCancellationRequested)
+                    while (true)
                     {
                         this.kafkaHandle.Poll((IntPtr)POLL_TIMEOUT_MS);
+                        ct.ThrowIfCancellationRequested();
                     }
                 }, ct, TaskCreationOptions.LongRunning, TaskScheduler.Default);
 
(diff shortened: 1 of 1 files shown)


src/Confluent.Kafka/Producer.cs
```diff
@@ -675,14 +676,29 @@ public int Flush()
         /// </remarks>
         public void Dispose()
         {
+            // TODO: If this method is called in a finalizer, can callbackTask potentially be null?
```
edenhill: should you add a test for that case?

src/Confluent.Kafka/Producer.cs
```diff
@@ -675,14 +676,29 @@ public int Flush()
...
-                // TODO: If called in a finalizer, can callbackTask be potentially null?
                 callbackCts.Cancel();
-                callbackTask.Wait();
+                try
+                {
+                    // Note: It's necessary to wait on callbackTask before disposing kafkaHandle
+                    // since the poll loop makes use of this.
+                    callbackTask.Wait();
+                }
+                catch (AggregateException e)
```
edenhill: Why is an exception better than waiting for the task to exit?

mhowlett: The old code was intermittently throwing the AggregationException (InnerException == TaskCanceledException). I don't understand where the race condition is (i did some testing to try and find it), and as far as I can tell the old code is valid. 

This definitely fixes the problem, and is valid, and I propose not spending any more time on it :-).

src/Confluent.Kafka/Producer.cs
mhowlett: yes. this is an old todo. lower priority.

src/Confluent.Kafka/Producer.cs
mhowlett: it's not

src/Confluent.Kafka/Producer.cs
```diff
@@ -52,9 +52,10 @@ public class Producer : IDisposable
         private Task StartPollTask(CancellationToken ct)
             => Task.Factory.StartNew(() =>
                 {
-                    while (!ct.IsCancellationRequested)
+                    while (true)
                     {
                         this.kafkaHandle.Poll((IntPtr)POLL_TIMEOUT_MS);
+                        ct.ThrowIfCancellationRequested();
```
edenhill: move this line above the poll since the task may have already been cancelled by the time it comes to the first iteration (from MS example)

edenhill:
LGTM after moving the throwIf.. line

src/Confluent.Kafka/Producer.cs
mhowlett: good point

Issue #400, opened 2018-01-03
Runtime error:  DLLNotFoundException: kernel32


edhalsim:
Description
===========

I'm getting the following error when running a sample program on CentOS 7:
Unhandled Exception:  System.DllNotFoundException: kernel32
  at (wrapper managed-to-native) Confluent.Kafka.Impl.LibRdKafka+WindowsNative:LoadLibraryEx (string,intptr,Confluent.Kafka.Impl.LibRdKafka/WindowsNative/LoadLibraryFlags)
  at Confluent.Kafka.Impl.LibRdKafka.Initialize (System.String userSpecifiedPath) [0x000b6] in <f8eec87f7e1a4d588a368976ab219188>:0
  at Confluent.Kafka.Consumer..ctor (System.Collections.Generic.IEnumerable`1[T] config) [0x0000c] in <f8eec87f7e1a4d588a368976ab219188>:0
  at Confluent.Kafka.Consumer`2[TKey,TValue]..ctor (System.Collections.Generic.IEnumerable`1[T] config, Confluent.Kafka.Serialization.IDeserializer`1[T] keyDeserializer, Confluent.Kafka.Serialization.IDeserializer`1[T] valueDeserializer) [0x0013f] in <f8eec87f7e1a4d588a368976ab219188>:0

This is using the librdkafka.so library I pulled and recompiled on CentOS.

Note that the SimpleProducer and SimpleConsumer apps. work fine when running with "dotnet run".  My app. I can't run using "dotnet run" because my app. needs to be installed as a Spark process using Mobius, which requires running via:  sparkclr-submit.sh.

Any ideas please let me know.  Thanks for your time.

mhowlett:
You'll need to use the `Library.Load` method to explicitly set the path to your librdkafka.so binary. We will hopefully have out of the box support for CentOS before too long, this is becoming a common issue.

edhalsim:
Thanks for your reply.  I tried that and I'm having the same issue.  I added:
    Library.Load("/home/<path>/Mobius-master/examples/Streaming/Kafka/bin/Release/librdkafka.so");
Then did a rebuild of the Mobius examples on Linux, but when I run it I get exactly the same error. except now it's coming from Library.Load:
  at Confluent.Kafka.Impl.LibRdKafka.Initialize (System.String userSpecifiedPath) [0x000b6] in <f8eec87f7e1a4d588a368976ab219188>:0
[...]

PR #418, opened 2018-01-22
Support mono (issue #400) 

ngeor:
I haven't tested it yet but this should prevent kernel32 from being invoked under Mac and Linux. From there, it should be possible to point to the correct file location via a configuration file using a dllmap

http://www.mono-project.com/docs/advanced/pinvoke/dllmap/

#400


ngeor:
[clabot:check]

jebidiah252:
Is there a reason why this wasn't merged in? I cloned the repo and then implemented the changes on CentOS 7 and saw the issue resolved from the changes made by @ngeor

PR #422, opened 2018-01-24
Confluent.SchemaRegistry + Confluent.Kafka.Avro project metadata

mhowlett:
@edenhill - forgot to put metadata in the sr and avro project files (which ends up in the nuget packages & on nuget.org)


edenhill:
LGTM

PR #424, opened 2018-01-25
Map r v0.11.3

rskibitskiy:
This is not final version. Added .nuspec for build nuget package. Changes with variables are needed because of C# 5 not supports this syntax like(out TYPE variable).

### DIFF
--- a/src/Confluent.Kafka/.nuspec
+++ b/src/Confluent.Kafka/.nuspec
@@ -1,0 +1,18 @@
+﻿<?xml version="1.0"?>
+<package>
+    <metadata>
+    <id>MapR.Streams.Dotnet</id>
+    <version>0.11.3</version>
+    <title>MapR Streams client for .Net</title>
+    <authors>MapR Inc.</authors>
+    <owners>MapR Inc.</owners>
+    <!--<licenseUrl></licenseUrl>
+    <projectUrl></projectUrl>
+    <iconUrl></iconUrl>-->
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>MapR Streams client for .Net</description>
+    <releaseNotes>Firs release of MapR dotnet client</releaseNotes>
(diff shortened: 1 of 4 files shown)


src/Confluent.Kafka/.nuspec
```diff
@@ -0,0 +1,18 @@
+﻿<?xml version="1.0"?>
+<package>
+    <metadata>
+    <id>MapR.Streams.Dotnet</id>
```
mhowlett: you're suggesting Confluent develop a library and then brand it with a competitors name?

src/Confluent.Kafka/Confluent.Kafka.csproj
```diff
@@ -1,33 +1,125 @@
...
+    <Reference Include="System.Core" />
+    <Reference Include="System.Xml.Linq" />
+    <Reference Include="System.Data.DataSetExtensions" />
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="System.Data" />
+    <Reference Include="System.Net.Http" />
+    <Reference Include="System.Xml" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="BrokerMetadata.cs" />
```
mhowlett: why all of this?

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -186,6 +187,19 @@ internal ErrorCode Produce(string topic, byte[] val, int valOffset, int valLengt
             var gchValue = default(GCHandle);
             var gchKey = default(GCHandle);
 
+            var matcher = new Regex("/([^:]+):(.*)");
+            var full_topic = topic;
+            if (matcher.Match(topic) == null)
```
mhowlett: you're not concerned of the perf impact of doing a regex match on every produce?

PR #445, opened 2018-03-02
Performance improvements on deserialization and publish report.

fryderykhuang:
We're going to use this library in production in the near future, so we did some load tests and found quite significant performance issue, finally decided to write our own consumer class.
Hope this modification can shed some light on future development.

### DIFF
--- a/src/Confluent.Kafka.Avro/AvroNativeDeserializer.cs
+++ b/src/Confluent.Kafka.Avro/AvroNativeDeserializer.cs
@@ -1,0 +1,256 @@
+﻿// Copyright 2016-2018 Confluent Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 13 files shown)


mhowlett:
thanks @fryderykhuang - looks great. in hindsight, it's indeed not ideal that the serialization interface uses byte[], and I have wondered about the perf increase by doing something along the lines of what you have done.

can you share any more details of your performance analysis? what is the increase in throughput? memory usage (quantitative, or qualitative)? i think it would be relatively straightforward to add a second serialization interface along side the one we have, and if the performance benefits are substantial, there is good chance we will do this.

mhowlett:
Rather than add unsafe overloads to the serializers, it looks like we're going to be able to use the (super awesome) new `Span<T>` without compromising on compatibility

https://msdn.microsoft.com/en-us/magazine/mt814808.aspx
https://channel9.msdn.com/events/Connect/2017/T125?term=c%23

From the first article:
"The core types like Span<T>, as well as the new types like Utf8Parser, are also on track to being made available in a System.Memory.dll package that’s compatible with .NET Standard 1.1. That will make the functionality available for existing releases of .NET Framework and .NET Core, albeit without some of the optimizations implemented when built into to the platform"

mhowlett:
This is going to land in .NET Core 2.1 (first preview release is out now).

PR #446, opened 2018-03-02
handle OperationCanceledException

mhowlett:
a problem introduced in #413 - the exception used to exit the while loop should be caught, it is not.

### DIFF
--- a/src/Confluent.Kafka/Producer.cs
+++ b/src/Confluent.Kafka/Producer.cs
@@ -52,11 +52,15 @@ namespace Confluent.Kafka
         private Task StartPollTask(CancellationToken ct)
             => Task.Factory.StartNew(() =>
                 {
-                    while (true)
+                    try
                     {
-                        ct.ThrowIfCancellationRequested();
-                        this.kafkaHandle.Poll((IntPtr)POLL_TIMEOUT_MS);
+                        while (true)
+                        {
+                            ct.ThrowIfCancellationRequested();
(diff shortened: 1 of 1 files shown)


src/Confluent.Kafka/Producer.cs
```diff
@@ -52,11 +52,15 @@ public class Producer : IDisposable
         private Task StartPollTask(CancellationToken ct)
             => Task.Factory.StartNew(() =>
                 {
-                    while (true)
+                    try
                     {
-                        ct.ThrowIfCancellationRequested();
-                        this.kafkaHandle.Poll((IntPtr)POLL_TIMEOUT_MS);
+                        while (true)
```
edenhill: Why not just this?:
```
  while (!ct.IsCancellationRequested) {
     this.kafkaHandle.Poll((IntPtr)POLL_TIMEOUT_MS);
  }
```

mhowlett: see comments in #413

edenhill: And the reason for the AggregationException is still unknown?

edenhill:
LGTM

src/Confluent.Kafka/Producer.cs
mhowlett: yes, I haven't looked into it again. I don't believe it is a symptom of something bigger.

PR #447, opened 2018-03-02
Serialize / Deserialize Avro.Generic.GenericRecord 

mhowlett:
@edenhill 

the diff is a bit awkward, probably best to look at the code first.

The `GenericRecord` specialization of `AvroSerializer` / `AvroDeserializer` calls for a different implementation to the existing supported types, so I added an abstraction to support this. This refactoring will also come in handy in the future (there are still parts of the avro spec not supported).


src/Confluent.Kafka.Avro/AvroSerializer.cs
```diff
@@ -40,118 +36,25 @@ namespace Confluent.Kafka.Serialization
...
 
         private const string InitialBufferSizePropertyName = "avro.serializer.buffer.bytes";
 
         private const string AutoRegisterSchemaPropertyName = "avro.serializer.auto.register.schemas";
 
-        /// <summary>
-        ///     The SchemaId corresponding to type <see cref="T"/>, or null if 
-        ///     the serializer has yet to be invoked.
-        /// </summary>
-        public int? SchemaId { get; private set; } = null;
```
mhowlett: removed this from the public api, because in the general case, the serializer manages more than one schema id. note that this has not yet been included in a release version.

mhowlett: (in particular this does not fit with using GenericRecord)

src/Confluent.Kafka.Avro/AvroSerializer.cs
```diff
@@ -40,118 +36,25 @@ namespace Confluent.Kafka.Serialization
...
         ///     The default initial size (in bytes) of buffers used for message 
         ///     serialization.
         /// </summary>
         public const int DefaultInitialBufferSize = 128;
 
-        /// <summary>
-        ///     True if the serializer will attempt to auto-register un-recognized schemas
-        ///     with Confluent Schema Registry, false if not.
-        /// </summary>
-        public bool AutoRegisterSchema { get; private set; } = true;
```
mhowlett: i also removed these. they're actually still applicable in all cases, but i feel there's no real need to expose these and safer not to.

src/Confluent.Kafka.Avro/AvroSerializer.cs
```diff
@@ -40,118 +36,25 @@ namespace Confluent.Kafka.Serialization
...
-
-        /// <summary>
-        ///     The avro schema used to write values of type <see cref="T"/>
-        /// </summary>
-        public Avro.Schema WriterSchema { get; private set; }
-
-        private void Initialize()
-        {
-            Type writerType = typeof(T);
-            if (typeof(ISpecificRecord).IsAssignableFrom(writerType) || writerType.IsSubclassOf(typeof(SpecificFixed)))
```
mhowlett: the logic is wrong here for `SpecificFixed` so I removed it. we can add this later.

mhowlett:
@DavidBetteridge identifies a deadlock situation related to this code in #449. I'd prefer to make changes to address this in a separate PR.

examples/AvroGeneric/Program.cs
```diff
@@ -0,0 +1,121 @@
...
+            var producerConfig = new Dictionary<string, object>
+            {
+                { "bootstrap.servers", bootstrapServers },
+                // Note: you can specify more than one schema registry url using the
+                // schema.registry.url property for redundancy (comma separated list).
+                // The property name is not plural to follow the convention set by
+                // the Java implementation.
+                { "schema.registry.url", schemaRegistryUrl },
+                // optional schema registry client properties:
+                { "schema.registry.connection.timeout.ms", 5000 },
```
edenhill: People will most likely copy these values as-is, so it might be better to comment them out and rely on defaults.

examples/AvroGeneric/Program.cs
```diff
@@ -0,0 +1,121 @@
...
+            };
+
+            var consumerConfig = new Dictionary<string, object>
+            {
+                { "bootstrap.servers", bootstrapServers },
+                { "group.id", Guid.NewGuid() },
+                { "schema.registry.url", schemaRegistryUrl }
+            };
+
+            var s = (RecordSchema)Schema.Parse(
```
edenhill: Add a comment on how to read this from file as an alternative

examples/AvroGeneric/Program.cs
```diff
@@ -0,0 +1,121 @@
...
+
+                consumer.OnConsumeError += (_, e)
+                    => Console.WriteLine("Consume error: " + e.Error.Reason);
+
+                consumer.Subscribe(topicName);
+
+                CancellationTokenSource cts = new CancellationTokenSource();
+                var consumeTask = Task.Factory.StartNew(() =>
+                {
+                    while (!cts.Token.IsCancellationRequested)
```
edenhill: Rumour on the street is that this should be in a try-block due to AggregationExceptions ;)

examples/AvroGeneric/Program.cs
```diff
@@ -0,0 +1,121 @@
...
+                Console.WriteLine($"{producer.Name} producing on {topicName}. Enter user names, q to exit.");
+
+                int i = 0;
+                string text;
+                while ((text = Console.ReadLine()) != "q")
+                {
+                    var record = new GenericRecord(s);
+                    record.Add("name", text);
+                    record.Add("favorite_number", i++);
+                    record.Add("favorite_color", "blue");
```
edenhill: so many u's missing hereI don't even know where to start :)

examples/AvroGeneric/Program.cs
```diff
@@ -0,0 +1,121 @@
...
+
+                    var deliveryReport = producer.ProduceAsync(topicName, text, record).Result;
```
edenhill: Sync producing is an anti-pattern, can you change this to a ContinueWith thing instead?

examples/AvroSpecific/Program.cs
```diff
@@ -115,7 +117,7 @@ static void Main(string[] args)
                 {
                     User user = new User { name = text, favorite_color = "green", favorite_number = i++ };
                     var deliveryReport = producer.ProduceAsync(topicName, text, user).Result;
```
edenhill: dito

src/Confluent.Kafka.Avro/GenericDeserializerImpl.cs
```diff
@@ -0,0 +1,94 @@
+﻿// Copyright 2016-2018 Confluent Inc.
```
edenhill: 2018

src/Confluent.Kafka.Avro/GenericDeserializerImpl.cs
```diff
@@ -0,0 +1,94 @@
...
+            using (var stream = new MemoryStream(array))
+            using (var reader = new BinaryReader(stream))
+            {
+                var magicByte = reader.ReadByte();
+                if (magicByte != Constants.MagicByte)
+                {
+                    // may change in the future.
+                    throw new InvalidDataException($"magic byte should be 0, not {magicByte}");
+                }
+                var writerIdBigEndian = reader.ReadInt32();
```
edenhill: this variable isn't used after NetworkToHostOrder, why not pass reader.ReadInt32() directly to NetworkToHostOrder?

src/Confluent.Kafka.Avro/GenericDeserializerImpl.cs
```diff
@@ -0,0 +1,94 @@
...
+                var writerId = IPAddress.NetworkToHostOrder(writerIdBigEndian);
+
+                DatumReader<GenericRecord> datumReader = null;
+
+                lock (datumReaderLock)
+                {
+                    datumReaderBySchemaId.TryGetValue(writerId, out datumReader);
+                    if (datumReader == null)
+                    {
+                        // TODO: If any of this cache fills up, this is probably an
```
edenhill: yes, this should really be an LRU

src/Confluent.Kafka.Avro/GenericDeserializerImpl.cs
```diff
@@ -0,0 +1,94 @@
...
+                        // indication of misuse of the deserializer. Ideally we would do 
+                        // something more sophisticated than the below + not allow 
+                        // the misuse to keep happening without warning.
+                        if (datumReaderBySchemaId.Count > schemaRegistryClient.MaxCachedSchemas)
+                        {
+                            datumReaderBySchemaId.Clear();
+                        }
+
+                        // we retain the lock over this blocking call since there may be concurrent
```
edenhill: I'm not sure how common multi-threaded consumption is in .NET clients, but an alternative would be to keep a separate lock on each cached schema and maintain a state (resolving, resolved) that other threads could wait on for a specific schema rather than being blocked by the entire schemaRegistryClient.

src/Confluent.Kafka.Avro/GenericSerializerImpl.cs
```diff
@@ -0,0 +1,137 @@
+﻿// Copyright 2016-2018 Confluent Inc.
```
edenhill: 2018

src/Confluent.Kafka.Avro/GenericSerializerImpl.cs
```diff
@@ -0,0 +1,137 @@
...
+
+namespace Confluent.Kafka.Serialization
+{
+    internal class GenericSerializerImpl : IAvroSerializerImpl<GenericRecord>
+    {
+        private ISchemaRegistryClient schemaRegistryClient;
+        private bool autoRegisterSchema;
+        private int initialBufferSize;
+        private bool isKey;
+
```
edenhill: This class seems to lack locking?

src/Confluent.Kafka.Avro/SpecificDeserializerImpl.cs
```diff
@@ -0,0 +1,139 @@
...
+            using (var stream = new MemoryStream(array))
+            using (var reader = new BinaryReader(stream))
+            {
+                var magicByte = reader.ReadByte();
+                if (magicByte != Constants.MagicByte)
+                {
+                    // may change in the future.
+                    throw new InvalidDataException($"magic byte should be 0, not {magicByte}");
+                }
+                var writerIdBigEndian = reader.ReadInt32();
```
edenhill: dito used-once

src/Confluent.Kafka.Avro/SpecificDeserializerImpl.cs
```diff
@@ -0,0 +1,139 @@
...
+                        var writerSchema = Avro.Schema.Parse(writerSchemaJson);
+
+                        datumReader = new SpecificReader<T>(writerSchema, ReaderSchema);
+                        datumReaderBySchemaId[writerId] = datumReader;
+                    }
+                }
+
+                return datumReader.Read(default(T), new BinaryDecoder(stream));
+            }
+        }
```
edenhill: this code looks almost identical to GenericDeserializerImpl, maybe refactor into one common base class?

src/Confluent.Kafka.Avro/SpecificSerializerImpl.cs
```diff
@@ -0,0 +1,138 @@
...
+            {
+                stream.WriteByte(Constants.MagicByte);
+
+                writer.Write(IPAddress.HostToNetworkOrder(writerSchemaId.Value));
+                avroWriter.Write(data, new BinaryEncoder(stream));
+
+                // TODO: maybe change the ISerializer interface so that this copy isn't necessary.
+                return stream.ToArray();
+            }
+        }
```
edenhill: this class seems to lack locking

examples/AvroGeneric/Program.cs
mhowlett: haha

examples/AvroGeneric/Program.cs
mhowlett: i'm numb to this by now

examples/AvroGeneric/Program.cs
mhowlett: ok, though I think this anti-pattern has it's place. note that after #431 is merged you'll be able to do this: `producer.Produce(dr => Console.WriteLine("blah"), topicName, text, record);`.

src/Confluent.Kafka.Avro/GenericDeserializerImpl.cs
mhowlett: more important though is the behavior if it does fill up, since adding an LRU will be just adding an optimization for a situation that generally shouldn't occur in practice.

src/Confluent.Kafka.Avro/SpecificDeserializerImpl.cs
mhowlett: you're right, though the code is relatively short, I think annoying to refactor (since the difference is in typing), and introducing a more complex structure (especially one that includes a base class) comes with the downside in making the code more difficult to follow (refer to the java implementation!). Also "rule of three" https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)

src/Confluent.Kafka.Avro/GenericDeserializerImpl.cs
mhowlett: mow you mention it, I guess this would be v. uncommon and my comment about retaining the lock over the blocking call is unimportant. The change @DavidBetteridge suggests in #449 is appropriate I think (simply use thread safe collections). I'll make that change.

edenhill: From what I understand that proposed solution will not work because of race conditions.

src/Confluent.Kafka.Avro/GenericSerializerImpl.cs
mhowlett: yes, good point. in the serializer case, locking would seem very important... using thread safe collections will suffice.

src/Confluent.Kafka.Avro/SpecificDeserializerImpl.cs
mhowlett: (so leaving as is)

src/Confluent.Kafka.Avro/GenericDeserializerImpl.cs
```diff
@@ -58,31 +56,23 @@ public GenericRecord Deserialize(string topic, byte[] array)
                 }
                 var writerId = IPAddress.NetworkToHostOrder(reader.ReadInt32());
 
-                DatumReader<GenericRecord> datumReader = null;
-
-                lock (datumReaderLock)
+                datumReaderBySchemaId.TryGetValue(writerId, out DatumReader<GenericRecord> datumReader);
```
edenhill: There is a race condition between this TryGetValue and the insert below.

mhowlett: I think it's benign. We don't care if we have make more than one datumReader - any of them will be valid - as long as we definitely have a valid instance to do the read (which we do). Ideally we wouldn't ever create more than one datumReader for a given writerId (hence what I was trying to do with the longer lasting lock), but it doesn't matter if we do (except possibly from a perf perspective), and as you point out, concurrent consuming won't be very common. Another thing that can happen (will be very uncommon) is the cache is cleared that includes a datumReader that's going to be used by another thread, but again that doesn't matter - the only thing that really matters is we have a valid datumReader instance when we go to use it.

src/Confluent.Kafka.Avro/GenericDeserializerImpl.cs
mhowlett: I think it's ok (see note below). the cache is just trying to avoid constructing `datumReader` instances unnecessarily, it doesn't actually matter if this happens.

src/Confluent.Kafka.Avro/GenericDeserializerImpl.cs
edenhill: If this is indeed intentional make a very clear comment about it, because this looks like a mistake/bug.

src/Confluent.Kafka.Avro/GenericDeserializerImpl.cs
mhowlett: there's no tech debt issue with the threading. I think not warning the user if the cache fills up is an issue (I think fail fast is better + force the user to manually flag they're doing something that is probably wrong to continue, but everyone else disagreed with me...), but I also don't think we need to track it. if people hit it, we'll find out about it and be reminded fix it.

src/Confluent.Kafka.Avro/GenericDeserializerImpl.cs
mhowlett: ok, will add a note.

src/Confluent.Kafka.Avro/GenericDeserializerImpl.cs
edenhill: After giving this some more thought...

There are two scenarios:
 1. consumers are used from a single thread only - this function will only be called by one thread at any given time. No blocking.
 2. consumers are possibly used from multiple threads - this function will be called simultaneously from multiple threads, possibly deserializing messages with the same schema-id, which will result in multiple parallell requests to the schema-registry. Multiply this by the number of consumer instances, apps, etc, and there will be an unnecessary spike of HTTP requests to the schema-registry.

Now, with the two different solutions:
 a. only lock datumReaderBySchemaId (race condition)
 b. lock entire Lookup/QuerySR/Insert cycle (no race condition)

For scenario 1 there is no difference between **a** and **b**.
For scenario 2:
 * with solution **a** there will be multiple outstanding schema-registry requests which we all can agree is a bad thing in large deployments. With SR serving each request in order it means there will be no speed-up, rather the opposite, for the awaiting consumer code wanting its messages deserialized.
 * with solution **b** there will be just one schema-registry request and the other consumer threads will block waiting for the schema to be queried and inserted. The wait time of the blocked threads will be no longer than the query+insert time of the first thread.

With this in mind I strongly vote for going back to the initial implementation with no race conditions (solution b).

src/Confluent.Kafka.Avro/GenericSerializerImpl.cs
```diff
@@ -31,9 +32,17 @@ internal class GenericSerializerImpl : IAvroSerializerImpl<GenericRecord>
         private int initialBufferSize;
         private bool isKey;
 
-        private Dictionary<Avro.RecordSchema, string> knownSchemas = new Dictionary<Avro.RecordSchema, string>();
-        private HashSet<KeyValuePair<string, string>> registeredSchemas = new HashSet<KeyValuePair<string, string>>();
-        private Dictionary<string, int> schemaIds = new Dictionary<string, int>();
+        private ConcurrentDictionary<Avro.RecordSchema, string> knownSchemas
```
edenhill: Same discussion here as for the Deserializer, but it is worse here because there are three different dicts that should be in sync but the lack of common locking will lead to the opposite.

mhowlett: the dicts don't need to always be in sync. there is a lock that covers the only case where this is important. 

but on reflection, i re-agree that preventing a spike in access to SR is important, particularly in the serializer case (going back to my initial implementation).

mhowlett:
added large scale locking to all deserializer / serializer implementation methods ( thanks for persisting @edenhill ) + added `.ConfigureAwait(false)` to all SR requests. All Unit and Integration tests pass. Should be good to go. 

@edenhill

src/Confluent.Kafka/Producer.cs
```diff
@@ -56,7 +56,7 @@ private Task StartPollTask(CancellationToken ct)
                     {
                         while (true)
                         {
-                            ct.ThrowIfCancellationRequested();
+                            ct.ThrowIfCancellationRequested(); d
```
edenhill: build much? ;)

src/Confluent.Kafka.Avro/GenericDeserializerImpl.cs
```diff
@@ -56,30 +58,29 @@ public GenericRecord Deserialize(string topic, byte[] array)
...
 
-                    datumReader = new GenericReader<GenericRecord>(writerSchema, writerSchema);
-                    datumReaderBySchemaId[writerId] = datumReader;
-                }
+                        datumReader = new GenericReader<GenericRecord>(writerSchema, writerSchema);
+                        datumReaderBySchemaId[writerId] = datumReader;
+                    }
 
-                return datumReader.Read(default(GenericRecord), new BinaryDecoder(stream));
+                    return datumReader.Read(default(GenericRecord), new BinaryDecoder(stream));
```
edenhill: shouldnt this be outside the lock?

src/Confluent.Kafka.Avro/SpecificDeserializerImpl.cs
```diff
@@ -105,26 +107,25 @@ public T Deserialize(string topic, byte[] array)
...
 
-                    datumReader = new SpecificReader<T>(writerSchema, ReaderSchema);
-                    datumReaderBySchemaId[writerId] = datumReader;
-                }
+                        datumReader = new SpecificReader<T>(writerSchema, ReaderSchema);
+                        datumReaderBySchemaId[writerId] = datumReader;
+                    }
 
-                return datumReader.Read(default(T), new BinaryDecoder(stream));
+                    return datumReader.Read(default(T), new BinaryDecoder(stream));
```
edenhill: dito, outside lock

src/Confluent.Kafka/Producer.cs
mhowlett: ahh. that was a merge done in github.

src/Confluent.Kafka.Avro/GenericDeserializerImpl.cs
mhowlett: you're right, it's better outside the lock, though doesn't need to be for correctness.

PR #449, opened 2018-03-05
I was finding that the consumer was deadlocking when called from a WinForm client

DavidBetteridge:
I was finding that the consumer was deadlocking when called from a WinForm client.  This was due to three methods in the call chain not having ConfigureAwait(false) set.

Additional, I've also removed the need for the lock constructs by replacing the Dictionaries with ConcurrentDictionaries. This means that these methods can be marked with the async keyword.

### DIFF
--- a/src/Confluent.Kafka.Avro/AvroDeserializer.cs
+++ b/src/Confluent.Kafka.Avro/AvroDeserializer.cs
@@ -24,7 +24,7 @@ using Avro.Specific;
 using Confluent.SchemaRegistry;
 using System.Reflection;
 using System;
-
+using System.Collections.Concurrent;
 
 namespace Confluent.Kafka.Serialization
 {
@@ -47,7 +47,7 @@ namespace Confluent.Kafka.Serialization
         ///     A datum reader cache (one corresponding to each write schema that's been seen) 
         ///     is maintained so that they only need to be constructed once.
         /// </remarks>
-        private readonly Dictionary<int, DatumReader<T>> datumReaderBySchemaId = new Dictionary<int, DatumReader<T>>();
+        private readonly ConcurrentDictionary<int, DatumReader<T>> datumReaderBySchemaId = new ConcurrentDictionary<int, DatumReader<T>>();
         private object datumReaderLock = new object();
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


DavidBetteridge:
[clabot:check]

mhowlett:
great, thanks @DavidBetteridge - important stuff, should be able to incorporate before the release. Will likely need to be a different PR because the generic avro serdes pr changes the class structure a bit.

src/Confluent.Kafka.Avro/AvroDeserializer.cs
```diff
@@ -173,24 +173,19 @@ public T Deserialize(string topic, byte[] array)
...
-                        datumReaderBySchemaId[writerId] = datumReader;
-                    }
+                    // it's a good idea to retain the lock over this blocking call since there
+                    // may be concurrent deserialize calls for the same T, and in that case it's
+                    // best not to hit Schema Registry more than once for the same information.
+                    var writerSchemaJson = SchemaRegistryClient.GetSchemaAsync(writerId).ConfigureAwait(false).GetAwaiter().GetResult();
+                    var writerSchema = Avro.Schema.Parse(writerSchemaJson);
+
+                    datumReader = new SpecificReader<T>(writerSchema, ReaderSchema);
+                    datumReaderBySchemaId[writerId] = datumReader;
```
edenhill: There is a race condition here between TryGetValue and the insert, the original lock is needed (which makes the use of Concurrent dict excessive).

DavidBetteridge: Good point but I was thinking that the race condition wouldn't matter that much as it just means there is a single additional call to the schema registry.

The more important part of the change,  is the inclusion of the .ConfigureAwait(false) as that helps prevent the deadlocks caused by context captures.

mhowlett: I added `ConfigureAwait(false)` to all SR method calls.

src/Confluent.SchemaRegistry/Rest/RestService.cs
```diff
@@ -189,48 +189,48 @@ private HttpRequestMessage CreateRequest(string endPoint, HttpMethod method, par
...
+                    $"/subjects/{subject}/versions/latest",
                     HttpMethod.Get);
 
         public async Task<int> RegisterSchemaAsync(string subject, string schema)
             => (await RequestAsync<SchemaId>(
-                    $"/subjects/{subject}/versions", 
-                    HttpMethod.Post, 
+                    $"/subjects/{subject}/versions",
+                    HttpMethod.Post,
                     new SchemaString(schema))).Id;
```
andreycha: Should also use `ConfigureAwait(false)`.

src/Confluent.SchemaRegistry/Rest/RestService.cs
```diff
@@ -255,34 +255,34 @@ public Task<Schema> CheckSchemaAsync(string subject, string schema)
 
         public async Task<Compatibility> GetGlobalCompatibilityAsync()
             => (await RequestAsync<Config>(
-                    "/config", 
+                    "/config",
                     HttpMethod.Get)).CompatibilityLevel;
```
andreycha: `ConfigureAwait(false)`

src/Confluent.SchemaRegistry/Rest/RestService.cs
```diff
@@ -255,34 +255,34 @@ public Task<Schema> CheckSchemaAsync(string subject, string schema)
...
 
         public async Task<Compatibility> GetCompatibilityAsync(string subject)
             => (await RequestAsync<Config>(
-                    $"/config/{subject}", 
+                    $"/config/{subject}",
                     HttpMethod.Get)).CompatibilityLevel;
```
andreycha: `ConfigureAwait(false)`

mhowlett:
Thanks  @DavidBetteridge, @andreycha - adding in `ConfigureAwait` as appropriate throughout the library now. IIRC other aspects were addressed in #447

PR #465, opened 2018-03-16
Avro serde support for .net standard 2.0

mhowlett:
this is dependent on merging https://github.com/confluentinc/avro/pull/5 . opening now to get a public CI build so people eager for this functionality can start testing it.

### DIFF
--- a/examples/AvroGeneric/AvroGeneric.csproj
+++ b/examples/AvroGeneric/AvroGeneric.csproj
@@ -1,71 +1,12 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
   <PropertyGroup>
-    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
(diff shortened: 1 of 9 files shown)


src/Confluent.Kafka/Serialization/Extensions/Message.cs
```diff
@@ -50,7 +50,7 @@ public static class MessageExtensions
             }
             catch (Exception ex)
             {
-                throw new KafkaException(new Error(ErrorCode.Local_KeyDeserialization, ex.Message), ex);
+                throw new KafkaException(new Error(ErrorCode.Local_KeyDeserialization, ex.ToString()), ex);
```
mhowlett: this provides a full stack trace rather than meaningless overview.

mhowlett:
Depends on https://github.com/confluentinc/avro/pull/5

examples/AvroGeneric/AvroGeneric.csproj
```diff
@@ -1,71 +1,12 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
edenhill: This is nice, but is it safe with older .NET frameworks?

mhowlett: yes, though probably not old versions of visual studio. but same constraints as other projects in the solution.

note: I'll do general verification of cross framework compatibility again when we do an RC.

PR #474, opened 2018-03-23
removed obsolete methods

mhowlett:


src/Confluent.Kafka/Consumer.cs
```diff
@@ -308,11 +308,6 @@ public void Poll(int millisecondsTimeout)
             }
         }
 
-        /// <include file='include_docs_consumer.xml' path='API/Member[@name="Poll"]/*' />
-        [Obsolete("Use an overload of Poll with a finite timeout.", false)]
-        public void Poll()
```
mhowlett: I propose later adding a method `Poll(CancellationToken)`, which is idiomatic C# that is implemented by calling `Poll` with a small timeout in a loop. We shouldn't have methods that block indefinitely without allowing the application to exit cleanly.

src/Confluent.Kafka/ISerializingProducer.cs
```diff
@@ -49,42 +49,5 @@ public interface ISerializingProducer<TKey, TValue>
         /// <include file='include_docs_producer.xml' path='API/Member[@name="ProduceAsync_string_Partition_TKey_TValue_Timestamp_IEnumerable"]/*' />
         /// <include file='include_docs_producer.xml' path='API/Member[@name="Produce_Action"]/*' />
         void Produce(Action<Message<TKey, TValue>> deliveryHandler, string topic, Partition partition, TKey key, TValue val, Timestamp timestamp, IEnumerable<Header> headers);
-
-#region obsolete produce methods
-
-        /// <include file='include_docs_producer.xml' path='API/Member[@name="ProduceAsync_Obsolete"]/*' />
-        [Obsolete("Variants of ProduceAsync that include a blockIfQueueFull parameter are depreciated - use the " + Producer.BlockIfQueueFullPropertyName + " configuration property instead.")]
-        Task<Message<TKey, TValue>> ProduceAsync(string topic, TKey key, TValue val, int partition, bool blockIfQueueFull);
```
mhowlett: rather than continue with this mess past 1.0.x, I think we should support 0.11.x for longer.

src/Confluent.Kafka/Producer.cs
```diff
@@ -392,28 +392,14 @@ Action<Message> deliveryHandler
...
-        public Producer(IEnumerable<KeyValuePair<string, object>> config, bool manualPoll, bool disableDeliveryReports)
+        private Producer(IEnumerable<KeyValuePair<string, object>> config)
         {
             // TODO: Make Tasks auto complete when EnableDeliveryReportsPropertyName is set to false.
             // TODO: Hijack the "delivery.report.only.error" configuration parameter and add functionality to enforce that Tasks 
             //       that never complete are never created when this is set to true.
 
             LibRdKafka.Initialize(null);
 
-            this.manualPoll = manualPoll;
```
mhowlett: to save you checking... these are set appropriately with field initializers.

gccmbr:
I am liking this "confluentinc:1.0-experimental" branch!  Thank you @mhowlett

mhowlett:
@gccmbr here's a sneak peak at the api changes mentioned elsewhere: https://github.com/mhowlett/confluent-kafka-dotnet/tree/newapi

the gist of it is `Message` now refers to the payload only and there are separate `DeliveryReport` and `ConsumerRecord` classes (that contain a `Message`). Producer takes a where (topic or topic/partition) and a what (message).

not quite done, but I hope to get v1.0-experimental-1 on nuget within a few days.

all api changes subject to change (but this is feeling pretty good, and has broad agreement)

gccmbr:
Awesome.  This is huge.  I understand the motivation now and how this would need to break previous logic / constructors.  I will nuget pre-release refresh every few hours. LOL.   Really exciting.  

As mentioned to @edenhill, I have a small PowerEdge 9-node Kafka cluster running from inside my closet and a 24/7 load to run on it.  I will do my best to provide meaningful feedback for bugs/performance/etc.

PR #484, opened 2018-03-29
fix load librdkafka.dll

pengweiqhca:


pengweiqhca:
[clabot:check]

mhowlett:
LGTM. I haven't thought about why this is necessary, but the extra search path can't hurt. Putting this into 1.0-experimental - there will be another nuget artifact generated from that before too long no doubt.

adison88:
@mhowlett looks like this is missing from 0.11.6.x. Can we add this as well?

PR #489, opened 2018-04-02
avrogen instructions -> readme

mhowlett:
@edenhill

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,6 +141,16 @@ public class Program
 }
 ```
 
+### AvroGen tool
+
+The Avro serializer and deserializer provided by `Confluent.Kafka.Avro` can be used with the `GenericRecord` class
+or with specific classes generated using the `avrogen` tool 
+(available [here](https://github.com/confluentinc/avro/releases/download/v1.7.7.4/avrogen.zip)). Usage:
+
+```
+dotnet /path/to/avrogen.dll -s your_schema.asvc .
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -141,6 +141,16 @@ public class Program
 }
 ```
 
+### AvroGen tool
+
+The Avro serializer and deserializer provided by `Confluent.Kafka.Avro` can be used with the `GenericRecord` class
```
edenhill: Maybe make Confluent.Kafka.Avro a link to the NuGet package?

README.md
```diff
@@ -141,6 +141,16 @@ public class Program
...
+or with specific classes generated using the `avrogen` tool 
+(available [here](https://github.com/confluentinc/avro/releases/download/v1.7.7.4/avrogen.zip)). Usage:
+
+```
+dotnet /path/to/avrogen.dll -s your_schema.asvc .
```
edenhill: Do we need to explain what is being generated and how to include them in the project (if that's needed?)?

README.md
mhowlett: people find this via their ide, and it's linked above, so i think no need.

README.md
mhowlett: yep, this would be better, but this was just a quick pr to get a link to the tool up here since it's not easy to find otherwise.

PR #503, opened 2018-04-09
added confluent cloud example

mhowlett:
An example demonstrating use of the client with confluent cloud.

### DIFF
--- a/examples/ConfluentCloud/ConfluentCloud.csproj
+++ b/examples/ConfluentCloud/ConfluentCloud.csproj
@@ -1,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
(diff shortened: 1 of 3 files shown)


examples/ConfluentCloud/ConfluentCloud.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp2.0</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Confluent.Kafka" Version="0.11.4" />
```
mhowlett: the other examples reference the Confluent.Kafka project in the current solution, which is arguably better, however I'm thinking of changing them all over to reference the published  nuget.org package like i'm doing here since this is how people will reference them their own programs, and they would prefer to be able to just copy/paste without thinking.

Issue #508, opened 2018-04-11
null reference exception due to non-deterministic finalization.


mhowlett:
Description
===========

```
Exception: System.NullReferenceException

Message: Object reference not set to an instance of an object.

StackTrace:   at Confluent.Kafka.Impl.NativeMethods.NativeMethods.rd_kafka_destroy(IntPtr rk)
  at Confluent.Kafka.Impl.SafeKafaHandle.ReleaseHandle()
  at System.Runtime.InteropServices.SafeHandle.InternalFinalize()
  at System.Runtime.InteropServices.SafeHandle.Finalize();
```

`SafeKafaHandle.ReleaseHandle` is trying to destroy the librdkafka handle, but to do so it needs a reference to `NativeMethods`, which may have already been cleaned up.

To fix, I think we should make `NativeMethods` a safe handle, then make use of the `DangerousAddRef` / `DangerousRelease` methods.

How to reproduce
================

Intermittent

Checklist
=========
Please provide the following information:

 - [ ] Confluent.Kafka nuget version: 0.11.4
 - [ ] Apache Kafka version: 
 - [ ] Client configuration:
 - [ ] Operating system:
 - [ ] Provide logs (with "debug" : "..." as necessary in configuration)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

PR #509, opened 2018-04-16
protect rd_kafka_destroy references from GC

mhowlett:
@edenhill - this addresses #508 

It seems clear that the issue is the order of object garbage collection is undefined, so it is unsafe to call through to rd_kafka_destroy in the finalizer of a producer/consumer. It's not completely clear to me from the stack trace in the issue exactly what object is being referenced, and it's difficult for me to replicate the issue, so I'm playing it safe and assuming it could be one of two. The `GCHandle.Alloc` call basically just marks an object as being referenced, so the GC will never collect it (it's not pinned though, so the object will move to generation 2 and not interfere with GC efficiency). I could reference count producer / consumer construction / destruction, and do a corresponding `GCHandle.Free`, but this would require substantial effort/complexity, and I see little point.

### DIFF
--- a/src/Confluent.Kafka/Impl/LibRdKafka.cs
+++ b/src/Confluent.Kafka/Impl/LibRdKafka.cs
@@ -131,7 +131,6 @@ namespace Confluent.Kafka.Impl
             _topic_conf_set_partitioner_cb = (Action<IntPtr, PartitionerDelegate>)methods.Where(m => m.Name == "rd_kafka_topic_conf_set_partitioner_cb").Single().CreateDelegate(typeof(Action<IntPtr, PartitionerDelegate>));
             _topic_partition_available = (Func<IntPtr, int, bool>)methods.Where(m => m.Name == "rd_kafka_topic_partition_available").Single().CreateDelegate(typeof(Func<IntPtr, int, bool>));
             _new = (Func<RdKafkaType, IntPtr, StringBuilder, UIntPtr, SafeKafkaHandle>)methods.Where(m => m.Name == "rd_kafka_new").Single().CreateDelegate(typeof(Func<RdKafkaType, IntPtr, StringBuilder, UIntPtr, SafeKafkaHandle>));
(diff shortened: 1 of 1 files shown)


mhowlett:
thanks. the cost-benefit of trying to construct such a test isn't worth it imo. if the reporter verifies this as resolved, I think that's good enough. I won't merge this immediately, will wait for confirmation.

mhowlett:
it's difficult to say. it could be a day or more and require us maintaining a second test environment (it's only ever been reported as happening in IIS).

edenhill:
Maybe IIS is not required but only surfaces the problem more often. I would like to see some effort of reproducing these types of bugs which are not crystal clear to reason about.

mhowlett:
of course, and not arguing it's not useful work, just prioritization. e.g. there are many high priority bugs to get fixed that people are actually run into a lot before putting in time to find other problems that may or may not exist.

pankajtakawale:
@mhowlett just faced this issue on the Confluent.Kafka `v0.11.4`
```Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
   at Confluent.Kafka.Impl.NativeMethods.NativeMethods_Debian9.rd_kafka_destroy(IntPtr rk)
   at Confluent.Kafka.Impl.SafeKafkaHandle.ReleaseHandle()
   at System.Runtime.InteropServices.SafeHandle.InternalFinalize()
   at System.Runtime.InteropServices.SafeHandle.Finalize()
```

mhowlett:
thanks for the feedback

this PR was merged after 0.11.4 (so not in a release yet).

generally, it's a bad idea to be relying on finalizers to clean up - best to make sure you're calling `Dispose` explicitly.

eiriktsarpalis:
I'm getting this error deterministicly when running nunit tests in dotnet core (has never triggered in net4x). Any chance of releasing this fix soon?

eiriktsarpalis:
Can reproduce the issue with `1.0.0-experimental-3` as well. Has that release incorporated the fix?

adison88:
I also experienced the same bug in 0.11.6 version. Is the fix part of it?

PR #543, opened 2018-06-02
Validate config values for 'not-null'

namelessvoid:
Recently, we provoked a generic NullReferenceException by accidentally passing a `null` value as configuration to the consumer constructor which took us some time to identify the parameter causing the trouble.

Therefore, this PR suggests to apply validation to the configuration passed to consumers and producers, checking all key-value-pairs' values for not null. In case the validation fails, an ArgumentExcpetion is thrown containing the configuration key name to make debugging more easy.

I've implemented the validation for the `1.0-experimental` branch. However - if more appropriate - I can supply the patch for another branch as well.

Looking forward to your feedback!

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -151,7 +151,7 @@ namespace Confluent.Kafka
             modifiedConfig
                 .ToList()
                 .ForEach((kvp) => {
-                    if(kvp.Value == null) throw new ArgumentException($"'{kvp.Key}' configuration parameter must not be null.");
+                    if (kvp.Value == null) throw new ArgumentException($"'{kvp.Key}' configuration parameter must not be null.");
                     configHandle.Set(kvp.Key, kvp.Value.ToString());
                 });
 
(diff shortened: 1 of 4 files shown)


namelessvoid:
[clabot:check]

mhowlett:
yep, i like this, thanks!

the alternative to throwing an exception would be to consider a `null` value to mean 'not set'. The benefit of that is it would allow inline expressions that return a config parameter, or indication that the default should apply (so allow terser code in some cases). The downside is that more often than not, a `null` passed as a parameter is likely to be something the developer didn't mean to do, so it's better to fail fast just in case. The latter argument wins for me.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -150,15 +150,21 @@ public Consumer(IEnumerable<KeyValuePair<string, object>> config)
             var configHandle = SafeConfigHandle.Create();
             modifiedConfig
                 .ToList()
-                .ForEach((kvp) => { configHandle.Set(kvp.Key, kvp.Value.ToString()); });
+                .ForEach((kvp) => {
+                    if(kvp.Value == null) throw new ArgumentException($"'{kvp.Key}' configuration parameter must not be null.");
```
mhowlett: space after `if` (other) places as well

mhowlett:
one small formatting thing to fix, then I'm good to merge. a PR against `master` would also be awesome :-) - if not i'll take care of it. it may be easy to cherry-pick this, but I can't remember.

namelessvoid:
Thanks for your feedback! I'll fix formatting and provide a PR against `master` as soon as possible.

PR #544, opened 2018-06-04
Validate config values for 'not-null' (on `master`)

namelessvoid:
Same feature as #543, cherry-picked to `master`.


mhowlett:
LGTM

PR #551, opened 2018-06-16
Introduced SyslogLevel enumeration for LogMessage.

kuskmen:
Hi, 
I've been using the client in our company and since I am constantly looking for open source projects to hop  in I said myself I might as well help you guys with whatever your need.

There is not labels for issues like (help-wanted, good-first-issue) even though I still think I need way more time to understand the whole project in order to start help with issues. 

So I thought I might as well look the code and perhaps find something that can be done which should not take much experience in the project (like a initial help so to say).

I noticed there are a lot of TODO's so I've started with this simple one, changing log level to enumeration rather than int.

However as simple as it looks there might be a lot of mistakes that I did while doing this, hence there are couple of questions I want to ask before finalizing the PR.

1. http://man7.org/linux/man-pages/man3/syslog.3.html this was the source I used when creating the enumeration (values for level) is this sufficient/correct?
2. Should we map to some explicit numbers each log level in the enumeration?
3. I put error log in the places where level is expected because that's what it looked to me, but I am almost certain that it is not right so what will be correct level?
4. Is this the correct branch for pre-release development? (There is no code of conduct so I decided that this might be the correct branch, but I am not 100% sure)

Regards,
kuskmen


kuskmen:
[clabot:check]

src/Confluent.Kafka/LogLevel.cs
```diff
@@ -0,0 +1,41 @@
...
+{
+    /// <summary>
+    ///  Represents enumeration with levels comming from syslog(3)
+    /// </summary>
+    public enum LogLevel
+    {
+        /// <summary>
+        ///  debug-level message
+        /// </summary>
+        Debug,
```
edenhill: these need to be explicitly mapped to syslog error levels, e.g. Debug = 7, ..

kuskmen: Hi @edenhill , thanks for the review. I will map them accordingly I assume that this is the desired mapping?
```
0 | EMERGENCY
1 | ALERT 
2 | CRITICAL 
3 | ERROR 
4 | WARNING
5 | NOTICE 
6 | INFORMATIONAL
7 | DEBUG
```

edenhill: Yep, looks good

mhowlett:
thanks @kuskmen - yep we should do this. It's a pity Microsoft log levels don't line up with syslog. I think the best name for the enum would be `SyslogLevel`, since it adds clear differentiation with `TraceLevel` in `System.Diagnostics` and `LogLevel` in `Microsoft.Extensions.Logging` which it might otherwise be confused with, and conveys something useful. I don't think there is an appropriate existing enum for this that we can use, and we probably wouldn't want to anyway in an effort to keep `Confluent.Kafka` dependencies as few as possible.

src/Confluent.Kafka/SysLogLevel.cs
```diff
@@ -3,39 +3,39 @@ namespace Confluent.Kafka
     /// <summary>
     ///  Represents enumeration with levels comming from syslog(3)
     /// </summary>
-    public enum LogLevel
+    public enum SysLogLevel
```
edenhill: nit: Should probably be "SyslogLevel", since syslog is referred to as "syslog", not "SysLog".

src/Confluent.Kafka/SysLogLevel.cs
```diff
@@ -0,0 +1,41 @@
+namespace Confluent.Kafka
+{
+    /// <summary>
+    ///  Represents enumeration with levels comming from syslog(3)
+    /// </summary>
+    public enum SysLogLevel
```
edenhill: nit: Since "syslog" is a well known entity and is always spelled as "syslog" (or sometimes "Syslog") but never "SysLog", I think this should be changed to "SyslogLevel".

src/Confluent.Kafka/Consumer.cs
```diff
@@ -67,11 +67,11 @@ private void LogCallback(IntPtr rk, int level, string fac, string buf)
             if (OnLog == null)
             {
                 // A stderr logger is used by default if none is specified.
-                Loggers.ConsoleLogger(this, new LogMessage(name, level, fac, buf));
+                Loggers.ConsoleLogger(this, new LogMessage(name, SyslogLevel.Error, fac, buf));
```
mhowlett: The level needs to be transferred all the way through from librdkafka, not replaced by error here. Also `int` is besty replaced by `SyslogLevel` all the way back to the p/invoke code (i.e. including in `LibRdKafka.LogDelegate`).

kuskmen: I hope I understood you correct, you want me to change `LogDelegate` definition to use `SyslogLevel` and platform invoke code will map int level from native to `SyslogLevel` enumeration.

Issue #406, opened 2018-01-12
ProduceAsync returns a task which never completes


badyam:
Description
===========
When Producer constructed with disableDeliveryReports = true option,
then ProduceAsync returns a task which never completes.
That may lead to a blocking of code execution and, potentially, to memory leaks.
This behavior is mentioned in docs on disableDeliveryReports parameter in Producer's constructor:

> If true, disables delivery report notification. Note: if set to true and you use a ProduceAsync variant that returns  a Task, the Tasks will never complete. Typically you should set this parameter to false. Set it to true for "fire and forget" semantics and a small boost in performance.

Despite of the fact this behavior is being described in docs, it seems weird. 
What is the reasoning behind returning a never-completed task?

I believe **in case of disabled delivery reports ProduceAsync should return completed task**.
Result of completed task should be a Message which has -1 in Partition and Offset properties  and null in Error, because these properties cannot be assigned with actual values without delivery report. Other properties of message can be actual.

Such behavior makes client code, that utilizes this method, a way more consistent and robust and prevents thread blocking and memory leaks.
Besides, such approach doesn't break an existing client code and doesn't violate "fire-and-forget" pattern.

As for impact of returning a completed task to performance, it seems to be inconsiderable.

To support my opinion I'm referencing to Stephen Toub's blog post.
https://blogs.msdn.microsoft.com/pfxteam/2011/10/02/dont-forget-to-complete-your-tasks/

How to reproduce
================
```
var producer = new Producer(new Dictionary<string, object> {{ "bootstrap.servers", "localhost:9092" }}, false, true);
var key = new byte[] { 16 };
var value = new byte[] { 42 };
var task = producer.ProduceAsync("test-topic", key, value);
var sentMessage = task.GetAwaiter().GetResult();
Console.WriteLine("Message sent");
```
"Message sent" wont be printed until program is forcibly terminated.

Checklist
=========
Please provide the following information:

 - [x] Confluent.Kafka nuget version: 0.11.3
 - [x] Apache Kafka version: 0.10.1.0
 - [ ] Client configuration:
 - [x] Operating system: Windows 10
 - [ ] Provide logs (with "debug" : "..." as necessary in configuration)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

mhowlett:
thanks @badyam - completely agree. this should be a quick win - feel free to submit a PR :-) or I will get to implementing it myself at some point.

badyam:
[...]

Issue #545, opened 2018-06-05
ProduceAsync blocks indefinitely


kuskmen:
Hi, 
this is more like a question since I don't have invitation in Slack I realized this would be the best place.

I need to make one simple producer for our automation tests and make sure message is successfully pushed before I can continue with test execution, but sadly producer blocks for indefinite amount of time.
```
public async Task PublishPurchases(IEnumerable<BetPurchase> purchases)
{
      foreach (var purchase in purchases)
      {
          await Producer.ProduceAsync("incoming-purchases", purchase.GetHashCode().ToString(), purchase);
      }
}
```

### What I noticed so far:

- I am using burrow dashboard as monitor for kafka, where I see messages are successfully pushed, but call is still blocked
- I noticed remarks comment on the `Producer<>.ProduceAsync` and subscribed for `Producer<>.OnError` event to check if the send queue is full, but no error was raised there.

This comes from remarks as well:
> Warning: if background polling is disabled and Poll is not being called in another thread, this will block indefinitely.

Sadly I am not sure what it means exactly? Which Poll method needs to be called in another thread? On `Consumer<>`? If yes, why would producer be blocked until consumer polls? How can I enable background polling?

mhowlett:
the producer has a background thread that polls for delivery reports - you can turn this off with a constructor argument, but generally you shouldn't.

what is your producer config?
how long did you wait? (how long is indefinite?).
does `Producer.ProduceAsync("incoming-purchases", purchase.GetHashCode().ToString(), purchase).Result` work? (i.e. a synchronous call).

kuskmen:
Hi @mhowlett , 

1. What is your producer config? - They are all default, I am only adding "bootstrap.servers" configuration.
2. How long did you wait? - As far as I recall I gave up after 1 minute of waiting.
3. does `Producer.ProduceAsync("incoming-purchases", purchase.GetHashCode().ToString(), purchase).Result` work? - Yes it does, but I'd really want to omit `.Result` as it might cause deadlocks https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html

marpit:
I think that it is related with this issue https://github.com/confluentinc/confluent-kafka-dotnet/issues/406

andreycha:
[...]

PR #553, opened 2018-06-18
ProducerAsync return task which never completes

andreycha:
Quick fix for https://github.com/confluentinc/confluent-kafka-dotnet/issues/406.

Several points:
- Decided to return `NoError` instance for error instead of null, because this property is always filled
- Not sure about `Timestamp`, so used about `NotAvailable` type

### DIFF
--- a/src/Confluent.Kafka/ErrorCode.cs
+++ b/src/Confluent.Kafka/ErrorCode.cs
@@ -415,7 +415,7 @@ namespace Confluent.Kafka
         IllegalSaslState = 34,
 
         /// <summary>
-        ///     Unuspported version
+        ///     Unsupported version
         /// </summary>
         UnsupportedVersion = 35
     };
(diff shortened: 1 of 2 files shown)


mhowlett:
thanks @andreycha - is this the problem reported in #545 as well? I was aware of the issue that you've resolved in this PR, what I'm trying to work out is if there is something else going on as well.

andreycha:
@mhowlett no, these are different scenarios. In #545 delivery **is** enabled. Here we have rather logical mistake, and in #545 I believe it's something related to sync context, because it only hangs in unit tests.

lalitbisht-Rg:
I am also facing this issue when running my code in EKS some times ProducerAsync return task which never completes
but same code running fine in Windows environment. below is my Producer config.

var config = new ProducerConfig
                {
                    //   GroupId = "" + ConfigurationManager.AppSettings["GroupId"],// "ARIFILES",
                    BootstrapServers = "" + ConfigurationManager.AppSettings["bootstrapserver"],
                    QueueBufferingMaxMessages = 300,
                    BatchNumMessages = 10,
                    CompressionLevel = 2,
                    Acks = Acks.All,
                    MessageTimeoutMs = 5000
                };

PR #555, opened 2018-06-20
Some code cleanup

andreycha:
Several minor things which I noticed while strolling though the codebase:

- Use C#7 features (inline variables, `is null`)
- Use `readonly` where applicable
- Rename two methods according to standard (CamelCase)
- Remove some extra empty lines


mhowlett:
these look good but I'm only going to merge bug fixes into the master branch at present, can you open a similar pr against 1.0-experimental?

mhowlett:
yep, that's what i meant to write.

andreycha:
Done.

PR #556, opened 2018-06-20
Correct assertions in tests

andreycha:
- Put expected values on correct place (first argument), so that assert messages will be correct
- Use `Assert.Contains` instead of `Assert.True` + `Contains`


mhowlett:
thanks @andreycha - but I beat you to it, this has already been corrected in the experimental branch. we'll hopefully get this merged up before too long.

andreycha:
@mhowlett oh, ok :). You might leave this till experimental branch is merged to see if there are any diffs left.

mhowlett:
the experimental branch is getting pretty hard to diff... basically i just fixed up all the compiler warnings, which I suspect is what you've done as well.

andreycha:
I see. Closing this one then.

Issue #498, opened 2018-04-05
Application breaks due to an exception if a consumer is polling when it is disposed


mindcrash:
Description
================

When a consumer is polling while being disposed an exception is thrown:

````
Unhandled Exception: System.ObjectDisposedException: kafka handle is closed
   at Confluent.Kafka.Impl.SafeHandleZeroIsInvalid.ThrowIfHandleClosed()
   at Confluent.Kafka.Impl.SafeKafkaHandle.ConsumerPoll(Message& message, IntPtr millisecondsTimeout)
   at Confluent.Kafka.Consumer.Consume(Message& message, Int32 millisecondsTimeout)
   at Confluent.Kafka.Consumer`2.Consume(Message`2& message, Int32 millisecondsTimeout)
   at Confluent.Kafka.Consumer`2.Poll(Int32 millisecondsTimeout)
   at Cloud.StreamingServices.Program.StatisticsService.<StartAsync>b__2_1() in C:\repos\streamingservices\statistics\src\Cloud.StreamingServices\Program.cs:line 367
   at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot)
````

Which, ofcourse, is completely understandable.

However, the only current way to make sure that disposing the consumer does not break the application which is hosting the consumer is by putting it into a very costly exception handler.

This could be addressed by exposing a public IsDisposed property which is flagged when Dispose is called, which makes the code mentioned below look like this instead:

```
                var config = new Dictionary<string, object>()
                {
                    { "bootstrap.servers", "127.0.0.1:29092" },
                    { "group.id", "consumers" }
                };                

                consumer = new Consumer<Ignore, byte[]>(config, null, new ByteArrayDeserializer());
		consume.Subscribe("topic");

                consumer.OnMessage += (o, m) => {
                    Console.WriteLine("Message received!");
                };

                await Task.Factory.StartNew(() => {
                    while(!consumer.IsDisposed)
                       consumer.Poll(10);
                });
            }
````

How to reproduce
================

Disposing the consumer (e.g. when responding to the CancelKeyPress event) when the following code is running will trigger the exception mentioned above

```
                var config = new Dictionary<string, object>()
                {
                    { "bootstrap.servers", "127.0.0.1:29092" },
                    { "group.id", "consumers" }
                };                

[...]

PR #558, opened 2018-06-21
Added IsDisposed public property to consumer and producer classes.

kuskmen:
* Capitalized on some newer C# language syntax.
* Removed unused namespaces

This is regarding enhancement that was requested here #498.
What do you think is it sufficient enough or I am missing something?


mhowlett:
`IsDisposed` is not something that is generally provided (is not part of `IDisposable`), so i think we want good reason to add this, otherwise it will promote ill conceived usage (e.g. #498). See note in #498.

mhowlett:
some nice modern C# syntax changes though - i'll merge so as to include them (but will remove the `IsDisposed` properties).

PR #568, opened 2018-07-12
Span<T> serializers

xqrzd:
Took a stab at adding `Span<T>` serializers. It probably needs some work still. If you already have plans to add these, feel free to close this PR.


src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -224,91 +225,51 @@ private IntPtr marshalHeaders(IEnumerable<Header> headers)
...
-                pKey = Marshal.UnsafeAddrOfPinnedArrayElement(key, keyOffset);
-            }
-
-            IntPtr headersPtr = marshalHeaders(headers);
+                    var errorCode = Librdkafka.producev(
+                        handle,
+                        topic,
+                        partition,
+                        (IntPtr)(MsgFlags.MSG_F_COPY | (blockIfQueueFull ? MsgFlags.MSG_F_BLOCK : 0)),
+                        Unsafe.AsRef<byte>(valPtr), (UIntPtr)val.Length,
```
xqrzd: I changed the signature of this function from `IntPtr` to `in byte`. If that doesn't make sense, I can change it back, and
`Unsafe.AsRef<byte>(valPtr)`
can be replaced with
`(IntPtr)valPtr`

mhowlett:
timely! I planned to try to update the serializer api later this week or next week. I haven't completely thought this through, but was thinking we want something similar to the new pipelines API which has a `GetMemory(int) ` method. i.e. avoid allocating memory on serialization if at all possible. Thread safety needs to be maintained as well, which will probably complicate that. The serialization case is more complex than the deserialization case. 

https://blogs.msdn.microsoft.com/dotnet/2018/07/09/system-io-pipelines-high-performance-io-in-net/

xqrzd:
Interesting, that's way more ambitious than my PR, I'm looking forward to it. I definitely agree it's complex. In particular, I wonder how to handle the case where the message size is unknown (eg. avro). Would the buffer be expandable, or would you ask for a bigger one, or would you ask for another buffer.

xqrzd:
@mhowlett I was thinking about this again, the internal (non-generic) Producer allows ideal memory pooling with this PR, here's some example pseudo code,

```csharp
// The internal producer
IProducer producer...

// RecyclableMemoryStream uses ArrayPool<byte>.Shared
// so it's able to resize if needed, and avoid allocations.
using (var keyStream = new RecyclableMemoryStream())
using (var valueStream = new RecyclableMemoryStream())
{
	SerializeKey(key, keyStream);
	SerializeValue(value, valueStream);

	// AsSpan() slices the underlying buffer,
	// so no extra allocations occur here.
	ReadOnlySpan<byte> keySpan = keyStream.AsSpan();
	ReadOnlySpan<byte> valueSpan = valueStream.AsSpan();

	var task = producer.ProduceAsync(..., keySpan, valueSpan, ...);
}

return task;
```

I'm going to see if I can get a more concrete example working in the generic producer. It's more tricky, since the user has to return a buffer, but doesn't have a way to dispose/free the buffer afterwards if it's needed.

mhowlett:
@xqrzd - sounds like this is likely on the right track (without having fully thought it through). Maybe we could have `ISerializer.Serialize(RecyclableMemoryStream stream, string topic, T data)`.

Another potential point of optimization for producing may be to avoid using `MsgFlags.MSG_F_COPY` on the `producev` call. This would mean pinning the key/value memory for longer until the delivery report is available. We need to pin anyway - in many cases there may not be downside to pinning longer.

[this task you've chosen is a tricky thing!]

PR #571, opened 2018-07-17
Add Consume overloads that always return message

rkoeninger:
Added overloads of the `Consumer.Consume` method that return the message from `kafkaHandle.ConsumerPoll` directly.

This allows client code to assess error codes for itself and respond to messages/errors/EOF immediately instead of via event handlers, which can become convoluted.

Returning `null` to signal failure is odd, but overloads that use an `out` parameter and return `bool` are already taken. Could also call these overloads by a different name like `Fetch`? More work could be done to determine why the underlying `Consume` function returned false and return a `Message` object with an appropriate error code instead of `null`.


rkoeninger:
@ConfluentCLABot [clabot:check]

mhowlett:
I think we'll move to something like:

```
                while (true)
                {
                    try
                    {
                        var cr = await consumer.ConsumeAsync(TimeSpan.FromSeconds(1));
                        if (cr.Message != null)
                        {
                            Console.WriteLine($"Message at {cr.TopicPartitionOffset}: ${cr.Value}");
                        }
                        if (cr.IsPartitionEOF) // checking this is optional. done inline.
                        {
                            // do stuff.
                        }
                    }
                    catch (ConsumeException e)
                    {
                        Console.WriteLine($"Consume error: {e.Error}");
                    }
                }
```

have a look at the `1.0-experimental` branch for the `ConsumerRecord` class definition

Issue #501, opened 2018-04-08
Slow AdvancedProducer provided by Confluent Kafka .NET


PingPongSet:
Description
===========

The code in the link below wass modified to send 50 messages of 100 characters. It takes around 50 seconds, which is about 200+ times slower than other brokers running on the same environment.

https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/examples/AdvancedProducer/Program.cs

How to reproduce
================

Basically, the code below is repeated to run 50 times

var deliveryReport = producer.ProduceAsync(topicName, "", messageWith100Characters);
var result = deliveryReport.Result; // synchronously waits for message to be produced.

Client: Confluent.Kafka nuget version: 0.11.4  runs on Windows 7
Server: RHEL 7+ via VirtualBox
Default settings from both Kafka and ZooKeeper are used

Checklist
=========
Please provide the following information:

 - [x] Confluent.Kafka nuget version: 0.11.4 https://www.nuget.org/packages/Confluent.Kafka/0.11.4
 - [x] Apache Kafka version: kafka_2.11-1.1.0.tgz
 - [x] Client configuration: .NET CORE 2
 - [x] Operating system: Redhat 7+
 - [ ] Provide logs (with "debug" : "..." as necessary in configuration)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

mhowlett:
this is extremely slow, even for synchronous producing (which is not recommended). I'm unsure why  based on the information provided.

mhowlett:
the 1s delay between send requests suggests you have either changed the settings away from the defaults (used in the example you cite) or you are somehow inserting a delay between produce requests. please provide a complete example demonstrating the problem, otherwise all we can do is speculate ...

PingPongSet:
I am not aware of changing default settings:

Main code below:

```
public class Producer: ProducerBase
    {
        public Producer(ILogger<MqBase> logger, IAppSetting appSetting) : base(logger, appSetting)
        {
        }

        public void Run()
        {
            MeasureTotalRunTime(() => {
                var brokerList = _appSetting.MessagingServerSetting.ServerName;
                var topic =  QUEUE_NAME ;
                Produce(brokerList, topic);
            });
        }
        public void Produce(string brokerList, string topic)
        {                             
            using (var producer = new Producer<string, string>(constructConfig(brokerList), new StringSerializer(Encoding.UTF8), new StringSerializer(Encoding.UTF8)))
[...]

PR #574, opened 2018-07-23
added note about socket.blocking.max.ms to readme

mhowlett:


mhowlett:
i think issue #501 is important enough to point out on the front page since many people are likely to run into it.

rnpridgeon:
LGTM!

PR #576, opened 2018-07-24
Simplified LINQ for LibRdKafka delegates initialization.

kuskmen:
Its not much of a improvement although I suspect tiny bit, but it should be better now simplified.


mhowlett:
thanks @kuskmen - good point.

PR #602, opened 2018-08-22
[WIP] Add support for per-partition batch consume

xqrzd:
I did the bare minimum to show an example, but basically this solves 2 issues,

- The batching removes the need to build queues on top of Confluent.Kafka
- The per-partition consume lets consumers read partitions independently, so a slow partition won't slow down others

I'm not sure if this is compatible with autocommit. I'm terrible at API design, and I'm not sure how this should be implemented. This PR isn't to be merged, it's just a discussion if per-partition consuming is worth adding. If it is, I'm happy to update this PR with a new design.

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -50,7 +50,7 @@ namespace Confluent.Kafka
         private IDeserializer<TKey> KeyDeserializer { get; }
         private IDeserializer<TValue> ValueDeserializer { get; }
 
-        private readonly SafeKafkaHandle kafkaHandle;
+        internal readonly SafeKafkaHandle kafkaHandle;
 
         private static readonly byte[] EmptyBytes = new byte[0];
 
@@ -367,6 +367,12 @@ namespace Confluent.Kafka
         internal ConsumeResult<TKey, TValue> Consume(int millisecondsTimeout)
         {
             var msgPtr = kafkaHandle.ConsumerPoll(enableTimestampMarshaling, enableHeaderMarshaling, (IntPtr)millisecondsTimeout);
+            return GetMessage(msgPtr);
+        }
+
+
+        internal ConsumeResult<TKey, TValue> GetMessage(IntPtr msgPtr)
+        {
             if (msgPtr == IntPtr.Zero)
             {
                 return new ConsumeResult<TKey, TValue>
@@ -413,7 +419,7 @@ namespace Confluent.Kafka
                     Librdkafka.message_headers(msgPtr, out IntPtr hdrsPtr);
                     if (hdrsPtr != IntPtr.Zero)
                     {
-                        for (var i=0; ; ++i)
+                        for (var i = 0; ; ++i)
                         {
                             var err = Librdkafka.header_get_all(hdrsPtr, (IntPtr)i, out IntPtr namep, out IntPtr valuep, out IntPtr sizep);
                             if (err != ErrorCode.NoError)
@@ -482,7 +488,7 @@ namespace Confluent.Kafka
                     unsafe
                     {
                         val = this.ValueDeserializer.Deserialize(
-                            topic, 
+                            topic,
                             msg.val == IntPtr.Zero ? EmptyBytes : new ReadOnlySpan<byte>(msg.val.ToPointer(), (int)msg.len),
                             msg.val == IntPtr.Zero);
                     }
@@ -505,8 +511,8 @@ namespace Confluent.Kafka
                     );
                 }
 
-                return new ConsumeResult<TKey, TValue> 
-                { 
+                return new ConsumeResult<TKey, TValue>
+                {
                     TopicPartitionOffset = new TopicPartitionOffset(topic, msg.partition, msg.offset),
                     Message = new Message<TKey, TValue>
                     {
@@ -749,17 +755,13 @@ namespace Confluent.Kafka
         ///     from and will replace any previous assignment.
         /// </summary>
         /// <param name="partitions">
-        ///     The set of partitions to consume from. If an offset value of
-        ///     Offset.Invalid (-1001) is specified for a partition, consumption
-        ///     will resume from the last committed offset on that partition, or
-        ///     according to the 'auto.offset.reset' configuration parameter if
-        ///     no offsets have been committed yet.
+        ///     The set of partitions to consume from. Consumption will resume
+        ///     from the last committed offset on each partition, or according
+        ///     to the 'auto.offset.reset' configuration parameter if no offsets
+        ///     have been committed yet.
         /// </param>
-        public void Assign(IEnumerable<TopicPartitionOffset> partitions)
-        {
-            assignCalled = true;
-            kafkaHandle.Assign(partitions.ToList());
-        }
+        public void Assign(IEnumerable<TopicPartition> partitions)
+            => Assign(partitions.Select(p => new TopicPartitionOffset(p, Offset.Invalid)).ToList());
 
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


mhowlett:
@xqrzd thanks for the pr! i'm definitely interested in a batching `Consume` method for a different reason - increased performance.

it would great if you could go into some more detail on your motivation here - give a concrete example demonstrating how a batching interface would simplify application code and also an example of where restricting the messages returned by `Consume` to be from a specific partition is useful.

mhowlett:
something relevant here is that you can use `Pause`/`Resume` to control which partitions in an assigned set messages will be consumed from. not exactly the same as what you're proposing, but similar (and the same as what is offered by the java consumer).

xqrzd:
Sorry for the late reply. I'll try and come up with a more concrete example when I have time, in the meantime I can give a quick overview.

I had looked into pause/resume, which accomplishes the same goal as per-partition consuming when it comes to dealing with slow partitions, however according to point `#4` here, https://github.com/edenhill/librdkafka/issues/1849#issuecomment-397763904 , librdkafka's internal message queue is flushed when a partition is paused. I was worried this could lead to more network traffic, as every time pause is called, all the flushed messages will have to be fetched again.

As for batching, some of our consumers simply read data from kafka, and store the results in a database. It's more efficient to insert multiple rows at a time, rather than doing one-off inserts. To accomplish this, I just built a simple queue on top of Confluent.Kafka, using `System.Threading.Channels`, that fires whenever a size or duration threshold is hit. librdkafka already has a function to do exactly this (`rd_kafka_consume_batch_queue`) this, and it's more efficient than building another queue on top of it.

Basically, I just want something like,
```
while (true) 
{
    var messages = consumer.ConsumeBatch(maxMessages: 1000, timeout: TimeSpan.FromSeconds(5));
    UpsertMessages(messages);
}
```

A real async consume would be pretty cool, since you could simultaneously consume many partitions with only a few threads (rather than a dedicated polling thread per partition with `rd_kafka_consume_batch_queue`), but I didn't see anything like that available in librdkafka.

ashishbhatia22:
@xqrzd - We have a similar need to consume messages in batch. Primary motivation is to perform database operations in batch. Can you please share the example that you used to build a queue on top of Confluent.Kafka? Thanks,

chadjefferies:
Curious if an idea like this has any chance of making it into the 1.0 release? It most certainly would be useful for all the reasons already mentioned.

mhowlett:
batch consume won't make it into 1.0 - absolute minimal scope is taking too long as it is! since you can implement this functionality on top of the api we provide, it's lower priority, though for the reasons provided in this thread, I think it's probably worth having.

@xqrzd - thanks for listing your use case. Also, thanks for the motivation for a real async consume method - we've been talking about the possibility of adding this. At this stage, we're mostly concerned with making sure the 1.0 api can grow into this if we want.

AndyPook:
As librdkafka pre-loads I don't _think_ there's much perf benefit to just reading until you have a batch ??

A real simple way to get batching is to use TPL-DataFlow and a BatchBlock (ensure you include a timeout to prevent stalling if the batch size is 100 and there's only 99 on the partitions you have assigned)

KrzysztofBranicki:
Part of this proposal that was described as 

> The per-partition consume lets consumers read partitions independently

would be very beneficial for the system my team is building. We us Kafka as Pub-sub to send domain events in the event-driven microservices scenario. Currently in order to be able to process events from multiple partitions in parallel we need to scale particular microservice horizontally, but this is wasteful since one instance has the capacity to process multiple events in parallel.

plachor:
In addition to above .. callback approach for consume method would be ideal for per-partition consume. So that library would take care of preservation of order within single partition but single consumer instance could process multiple consumes in parallel.

AndyPook:
Just me 2p/2c...

The "problem" with consuming many messages to be processed in parallel is offset tracking. Somehow you have to figure out when to commit the offset. Depending on your use case this may be super important.Are your messages idempotent? If it's ok to "lose" messages (ie not guarantee that every one is processed/handled) then not so much. Or, via some retry policy, failures could re-publish the message to be handled later. But then maybe ordering is important.

There are a load of techniques laid out in the kafka docs around offset management. But can be super complex to handle all the edge cases. Concurrency, ordering, finding the "holes" in the sequence...
If this is important, it is _much_ simpler to just partition to the level of parallelism you desire and run that number of consumers. These can easily be run in the same process if that's how you want it.
If you are running in docker/swarm/k8s then just run as many containers up to the number of partitions.
Yes, it may not seem "efficient" but it is so much easier to reason about for very little resource expense.

The scale out of consumers could maybe be handled with inspection through the admin client or maybe the partition assignment events.

Lastly, instrument all the things. A huge factor here is the median time to process a message and the required throughput. Factor the number of partitions/consumers/workers accordingly.

A lot of this assumes random (ish) distribution of messages across partitions. If you have an explicit strategy (ie priority, client id ...) then who knows :) but maybe use topic + naming convention + dynamic clients with a "manager" using admin client.

Anyway, YMMV

KrzysztofBranicki:
@AndyPook I don't see how offset tracking is an issue in parallel message processing as long as those messages that are processed in parallel are coming from different partitions (from perspective of single partition messages are still processed sequentially).

Just to answer some of the questions about the context - as I said we use kafka to send domain events e.g. `ReservationCreated`, `ReservationCancelled` etc. so we can't lose messages. What we are delivering is at least once delivery guarantees. Because we have at least once delivery guarantees all our event handler are idempotent. Ordering is also important but only within single partition so we use whatever makes sense form domain perspective as the partitioning key e.g. `TenantId`. Because we have at least once delivery guarantees we can leverage offset autocommit functionality without any problems (we just mark message after it is processed successfully and kafka client will commit the offset after appropriate batch size is reached or timeout).

We know that we can scale instances in Kubernetes horizontally, this is exactly what we are doing today to be able to process events from different partitions in parallel. But this is wastefully because one microservice instance can only process one message at the time. This is especially wastefully when you consider that we are using C# `async await` in our event handlers for all the IO/network bound operations. `async await` gives back the thread to the threadpool when it is waiting for the IO bound operation to complete but in this case no one will profit from this thread being returned to the pool because you are not able to start processing next message until you finish with the current one. 

I understand that your recommendation is to spin up multiple consumers inside one microservice process. Are there any cons to this approach, I thought that general recommendation was to have one consumer per process?

AndyPook:
:) yup we're on the same wavelength. Pardon me for making assumptions. Quite often, when peeps talk about "process messages in parallel" they mean read a bunch of messages then Parallel.Foreach or TPL-DataFlow = the issues you've already figured out.

wrt threads, async etc I think the effects are actually smaller than  certainly I originally thought...
Yes, Tasks are awaited, releasing Threads (and stuff). But, with the "virtualization" via containers all the Tasks across all the containers will be scheduled across the available hardware threads.

You  certainly can run multiple consumers from a single process. But then you lose the ability to scale across the cluster as they are effectively locked to a node within a single container.
You def do "lose" some memory as every process/container will need some before actually handling a message.
I'm currently of the opinion that the trade-off with memory and a little cpu vs  flexibility/scale/resilience is very much worth it.

wdamron:
The ability to independently consume from assigned partitions (using consumer groups) can be helpful when a consumer is doing some latency-bound work for each message and/or uses retry logic (common for at-least-once processing).

With the current shared queue, a slow/blocking consumer causes head-of-line blocking for all its assigned partitions. If the consumer blocks for long enough, of course the max poll interval will be exceeded and the consumer will leave the group.

Partition-level consumers are supported by librdkafka [(reference)](https://github.com/edenhill/librdkafka/wiki/FAQ#what-are-partition-queues-and-why-are-some-partitions-slower-than-others). The existing alternative of using `.Pause` and `.Resume` to handle "slow" partitions has its own drawbacks with respect to flushing the prefetch queue.

---

A basic use-case could be an at-least-once consumer which calls out to an external service or database for each message, retrying until the operation succeeds. While a message is being retried, the consumer would be blocked for all its assigned partitions despite only 1 partition being blocked by that message, the `.Consume` method would not be called, and callbacks would not be triggered.

Partition-level consumers would work around this somewhat, allowing the "slow" partition to not block the others, and also allowing the main `.Consume` method to be called more often than the occasionally-blocked handler can call it (ensuring callbacks are handled). A "slow" partition could be caused by data-dependent control flow, especially if messages sharing the same key/partition happen to cause the same slowness.

Building an in-memory queueing solution on top of this library to allow some form of windowed out-of-order processing with respect to separate partitions (or message keys) is pretty do-able, but can require quite a bit of bookkeeping depending on the requirements, and still may block the `.Consume` method from being called often enough when the window can't be advanced.

---

In C, the setup might be something roughly like the following (which is implemented in this PR):

```
// get partition queue:
rkqu = rd_kafka_queue_get_partition(rk, topic, partition);
// disable forwarding to main queue:
rd_kafka_queue_forward(rkqu, NULL);

// consume partition:
rkm = rd_kafka_consume_queue(rkqu, timeout);

// elsewhere, call poll to ensure callbacks are handled:
rkm = rd_kafka_consumer_poll(rk, timeout); // will always return null
```

Also, there are some related issues in the tracker:
* https://github.com/confluentinc/confluent-kafka-dotnet/issues/790
* https://github.com/confluentinc/confluent-kafka-dotnet/issues/812
* https://github.com/confluentinc/confluent-kafka-dotnet/issues/918

tomasz-podolak:
@mhowlett, are there any plans to expose per partition consumers/queues in the nearest future?

Issue #614, opened 2018-08-30
1.0 API Feedback


mhowlett:
This issue is to track community feedback on the new 1.0 API and document changes that are likely to be made before it's release. We'll continue to update this first comment to reflect open changes currently under consideration & progress.

For problems, please open another issue.

Comments / suggestions? Add them below!

Current changes under consideration for 1.0:

- add `async` serializers/deserializers (serdes that return `Task`s) in addition to simple serde types.
- `ProduceAsync` - correctly implement a `CancellationToken` parameter.
- `AdminClient.GetMetadata` - remove this method and add methods analogous to the Java client.
- `ListGroups` - parse the group data and change the return type.
- `Span<T>` serializers.
- return types from `AdminClient` methods. Even though it's not useful now, a return type from e.g. `CreateTopicsAsync` would allow for easier future extensibility.
- how should we communicate that a subscription didn't match any topics? 
- Exception types corresponding to more `ErrorCode`s.
- Investigate making `Consumer<,>` implement `IEnumerable`.
- pass message headers to deserializers
- `Commit` to throw `CommitException`, not `TopicPartitionOffsetException`.

Review TODOs:

- Add `SCRAM-SHA-1` sasl mechanism.
- Avro serializers should have adaptive buffer sizing.
- Reconsider various strongly typed config enum names. in particular: `Sasl_Plaintext`, `Sasl_Ssl`, `Roundrobin`, `Ssl`.
- Ordering of configuration is important to librdkafka. this isn't compatible with use of `Dictionary<K,V>`.
- Check that the float and double serdes are IEEE-754 compatible. Consider using `IPAddress.HostToNetworkOrder`.

xqrzd:
Super exciting to see it getting close to completion! Speaking of `Span<T>` serializers, I still haven't come up with something I'm happy with, but recently I prototyped using `IMemoryOwner<byte>` for the return type of `ISerializer<T>` here https://github.com/xqrzd/confluent-kafka-dotnet/commit/0686dcd58188bc0d8ae4233a053222ece422be50. The serializer would be responsible for allocating an `IMemoryOwner<byte>`, and would transfer ownership to Confluent.Kafka, according to this rule, https://gist.github.com/GrabYourPitchforks/4c3e1935fd4d9fa2831dbfcab35dffc6#rule-7-if-you-have-an-imemoryownert-reference-you-must-at-some-point-dispose-of-it-or-transfer-ownership-but-not-both
[...]

PR #617, opened 2018-08-31
added basic auth capability to schema registry client

mhowlett:


test/Confluent.SchemaRegistry.IntegrationTests/docker-compose.yaml
```diff
@@ -0,0 +1,34 @@
+version: '3'
```
mhowlett: this docker compose file brings up a zookeeper and kafka instance just for the basic auth schema registry [copy/pasted from ryan]. I need to update the integration tests for .net to use docker more generally. I'll incorporate this into that when it's finally done a different PR. Also, I chose not to bother with SSL as i think that adds very marginal benefit for the effort required.

src/Confluent.Kafka/Config.cs
```diff
@@ -0,0 +1,903 @@
...
+        /// </summary>
+        Plaintext,
+
+        /// <summary>
+        ///     Ssl
+        /// </summary>
+        Ssl,
+
+        /// <summary>
+        ///     Sasl_Plaintext
```
edenhill: Maybe the underscore should be removed since we already have camelcase to separate the words

src/Confluent.Kafka/Config.cs
```diff
@@ -0,0 +1,903 @@
...
+    {
+        /// <summary>
+        ///     Range
+        /// </summary>
+        Range,
+
+        /// <summary>
+        ///     Roundrobin
+        /// </summary>
+        Roundrobin
```
edenhill: This needs to be RoundRobin for me to remain sane

src/Confluent.Kafka/Config.cs
```diff
@@ -0,0 +1,903 @@
...
+        ///     an existing key/value pair collection.
+        /// </summary>
+        public ClientConfig(IEnumerable<KeyValuePair<string, string>> config) { this.properties = new Dictionary<string, string>(config.ToDictionary(a => a.Key, a => a.Value)); }
+
+        /// <summary>
+        ///     Set a configuration property using a string key / value pair.
+        /// </summary>
+        /// <remarks>
+        ///     Two scenarios where this is useful: 1. For setting librdkafka
+        ///     plugin config properties. 2. You are using a different version of
```
edenhill: Make sure this dict retains ordering

src/Confluent.Kafka/Config.cs
```diff
@@ -0,0 +1,903 @@
...
+    /// </summary>
+    public enum SecurityProtocolType
+    {
+        /// <summary>
+        ///     Plaintext
+        /// </summary>
+        Plaintext,
+
+        /// <summary>
+        ///     Ssl
```
edenhill: Is this decapalization standard in .NET?

src/Confluent.Kafka/Deserializers.cs
```diff
@@ -0,0 +1,279 @@
...
+        };
+
+        /// <summary>
+        ///     Deserializes a null value to a null value.
+        /// </summary>
+        public static Deserializer<Null> Null = (topic, data, isNull) =>
+        {
+            if (!isNull)
+            {
+                throw new System.ArgumentException("Deserializer<Null> may only be used to deserialize data that is null.");
```
edenhill: As the argument isn't passed by the user but by a consumed message I think perhaps ValueException (or whatever) might be better

src/Confluent.Kafka/Deserializers.cs
```diff
@@ -0,0 +1,279 @@
...
+        ///     True if the data is null, false otherwise.
+        /// </param>
+        /// <returns>
+        ///     The deserialized <see cref="System.Int64"/> value.
+        /// </returns>
+        public static long Long(string topic, ReadOnlySpan<byte> data, bool isNull)
+        {
+            if (isNull)
+            {
+                throw new ArgumentException($"Arg [{nameof(data)}] is null");
```
edenhill: ditolito

src/Confluent.Kafka/Deserializers.cs
```diff
@@ -0,0 +1,279 @@
...
+            // network byte order -> big endian -> most significant byte in the smallest address.
+            return
+                (((int)data[0]) << 24) |
+                (((int)data[1]) << 16) |
+                (((int)data[2]) << 8) |
+                (int)data[3];
+        }
+
+        /// <summary>
+        ///     Deserializes a big endian encoded (network byte ordered) System.Single value from a byte array.
```
edenhill: Is this IEEE-754 compatible? It really should be

src/Confluent.Kafka/Deserializers.cs
```diff
@@ -0,0 +1,279 @@
...
+        /// <param name="data">
+        ///     A byte array containing the serialized System.Double value (big endian encoding).
+        /// </param>
+        /// <param name="isNull">
+        ///     True if the data is null, false otherwise.
+        /// </param>
+        /// <returns>
+        ///     The deserialized System.Double value.
+        /// </returns>
+        public static double Double(string topic, ReadOnlySpan<byte> data, bool isNull)
```
edenhill: dito IEEE-754

src/Confluent.Kafka/Producer.cs
```diff
@@ -39,30 +63,54 @@ public class Producer<TKey, TValue> : IProducer<TKey, TValue>
...
             }
 
             if (valueSerializer == null)
             {
-                if (typeof(TValue) != typeof(Null))
+                if (typeof(TValue) == typeof(Null))
+                {
+                    this.valueSerializer = Serializers.Null as Serializer<TValue>;
+                }
+                else if (typeof(TValue) == typeof(byte[]))
```
edenhill: should this be isinstanceof rather than typeof?

src/Confluent.Kafka/Serializers.cs
```diff
@@ -0,0 +1,196 @@
...
+        /// </returns>
+        public static byte[] Double(string topic, double data)
+        {
+            if (BitConverter.IsLittleEndian)
+            {
+                unsafe
+                {
+                    byte[] result = new byte[8];
+                    byte* p = (byte*)(&data);
+                    result[7] = *p++;
```
edenhill:
```

      if (BitConverter.IsLittleEndian)
         Array.Reverse(bytes);
```

src/Confluent.Kafka/Serializers.cs
```diff
@@ -0,0 +1,196 @@
...
+        ///     The <see cref="System.Int32"/> value <paramref name="data" /> encoded as a byte array of length 4 (network byte order).
+        /// </returns>
+        public static byte[] Int32(string topic, int data)
+        {
+            var result = new byte[4]; // int is always 32 bits on .NET.
+            // network byte order -> big endian -> most significant byte in the smallest address.
+            // Note: At the IL level, the conv.u1 operator is used to cast int to byte which truncates
+            // the high order bits if overflow occurs.
+            // https://msdn.microsoft.com/en-us/library/system.reflection.emit.opcodes.conv_u1.aspx
+            result[0] = (byte)(data >> 24);
```
edenhill: there are builtin methods for int swapping:
https://docs.microsoft.com/en-us/dotnet/api/system.net.ipaddress.hosttonetworkorder?view=netframework-4.7.2

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -66,25 +58,76 @@ public int MaxCachedSchemas
...
+            string username = null;
+            string password = null;
 
-            var identityMapCapacityMaybe = config.Where(prop => prop.Key.ToLower() == SchemaRegistryMaxCachedSchemasPropertyName).FirstOrDefault();
-            this.identityMapCapacity = identityMapCapacityMaybe.Value == null ? DefaultMaxCachedSchemas : (int)identityMapCapacityMaybe.Value;
+            if (basicAuthSource == "USER_INFO")
+            {
+                if (basicAuthInfo == "")
+                {
+                    throw new ArgumentException($"CachedSchemaRegistryClient: {SchemaRegistryConfig.PropertyNames.SchemaRegistryBasicAuthCredentialsSource} was set to 'USER_INFO', but no value was specified for {SchemaRegistryConfig.PropertyNames.SchemaRegistryBasicAuthUserInfo}.");
```
edenhill: you may drop the "was":es if you like

src/Confluent.SchemaRegistry/Rest/RestService.cs
```diff
@@ -137,6 +161,9 @@ private async Task<HttpResponseMessage> ExecuteOnOneInstanceAsync(HttpRequestMes
                 }
                 catch (HttpRequestException e)
                 {
+                    // don't retry error responses originating from Schema Registry.
+                    if (e is SchemaRegistryException) { throw; }
+
```
edenhill: where do the below exceptions bubble up to the user? Produce()? Poll?

test/Confluent.Kafka.Avro.UnitTests/Configuration.cs
```diff
@@ -48,10 +48,10 @@ public void SerializerConfigure()
         {
             var avroSerializer = new AvroSerializer<int>(schemaRegistryClient);
 
-            var config = new Dictionary<string, object>
+            var config = new Dictionary<string, string>
             {
-                { "avro.serializer.buffer.bytes", 42 },
-                { "avro.serializer.auto.register.schemas", false }
+                { "avro.serializer.buffer.bytes", "42" },
```
edenhill: the auto conversion of types was pretty nice....

examples/AvroBlogExamples/Program.cs
```diff
@@ -35,13 +35,10 @@ class Program
...
         {
-            var config = new Dictionary<string, object>
-            {
-                { "bootstrap.servers", bootstrapServers },
-                { "schema.registry.url", schemaRegistryUrl }
-            };
+            var producerConfig = new ProducerConfig { BootstrapServers = bootstrapServers };
 
-            using (var producer = new Producer<Null, GenericRecord>(config, null, new AvroSerializer<GenericRecord>()))
+            using (var serdeProvider = new AvroSerdeProvider(new AvroSerdeProviderConfig { SchemaRegistryUrl = schemaRegistryUrl }))
```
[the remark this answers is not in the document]
mhowlett: I thought about this a lot, convinced this is best factoring. The requirements for Avro serdes requirements really make things difficult. Worth noting that there is one remaining related issue here - the serialize / deserialize functions should really never block - they should return `Task`s. However i'm concerned about the perf impact on the non-avro serdes (todo).

src/ConfigGen/Program.cs
```diff
@@ -0,0 +1,717 @@
...
+            { "partitioner", new List<string> { "random", "consistent", "consistent_random", "murmur2", "murmur2_random" } } 
+        };
+
+        /// <summary>
+        ///     A function that filters out properties from the librdkafka list that should
+        ///     not be privided in the 
+        /// </summary>
+        internal static List<PropertySpecification> RemoveLegacyOrNotRelevant(List<PropertySpecification> props) 
+            => props.Where(p => {
+                if (p.Name == "sasl.mechanisms") { return false; } // handled as a special case.
```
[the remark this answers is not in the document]
mhowlett: ok, i should have used a collection.... needless to say, this list wasn't quite so long when I started writing the function. I don't think this is a big deal and at this point going to leave as is.

examples/AvroGeneric/Program.cs
```diff
@@ -80,8 +77,12 @@ static void Main(string[] args)
             CancellationTokenSource cts = new CancellationTokenSource();
             var consumeTask = Task.Run(() =>
             {
-                using (var consumer = new Consumer<string, GenericRecord>(consumerConfig, new AvroDeserializer<string>(), new AvroDeserializer<GenericRecord>()))
+                using (var serdeProvider = new AvroSerdeProvider(avroConfig))
+                using (var consumer = new Consumer<string, GenericRecord>(consumerConfig, serdeProvider.GetDeserializerGenerator<string>(), serdeProvider.GetDeserializerGenerator<GenericRecord>()))
                 {
+                    consumer.OnError += (_, e)
```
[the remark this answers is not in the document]
mhowlett: it will be lost. the plan ultimately is to do some caching on startup and trigger events when the handler is added. this is not ideal, but having iterated over pretty much every other API option, i'm convinced it's the best all things considered. it's the only tolerable API choice imo.

src/ConfigGen/Program.cs
```diff
@@ -0,0 +1,717 @@
...
+                if (p.Name == "sasl.mechanism") { return false; } // handled as a special case.
+                if (p.Name == "consume.callback.max.messages") { return false; }
+                if (p.Name == "offset.store.path") { return false; }
+                if (p.Name == "offset.store.sync.interval.ms") { return false; }
+                if (p.Name == "enabled_events") { return false; }
```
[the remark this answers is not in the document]
mhowlett: good point, fixed.

src/ConfigGen/Program.cs
```diff
@@ -0,0 +1,717 @@
...
+        public SaslMechanismType? SaslMechanism
+        {
+            get
+            {
+                var r = Get(""sasl.mechanism"");
+                if (r == null) { return null; }
+                if (r == ""GSSAPI"") { return  SaslMechanismType.Gssapi; }
+                if (r == ""PLAIN"") { return SaslMechanismType.Plain; }
+                if (r == ""SCRAM-SHA-256"") { return SaslMechanismType.ScramSha256; }
+                if (r == ""SCRAM-SHA-512"") { return SaslMechanismType.ScramSha512; }
```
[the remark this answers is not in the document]
mhowlett: librdkafka CONFIGURATION docs also need updating.

src/ConfigGen/Program.cs
```diff
@@ -0,0 +1,717 @@
...
+
+";
+    }
+
+    
+    class PropertySpecification
+    {
+        public bool IsGlobal { get; set; }
+        public string Name { get; set; }
+        public string CPorA { get; set; }
```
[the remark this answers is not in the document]
mhowlett: Consumer, Producer, or All. Added a comment. this is an internal class, so not especially important.

src/Confluent.Kafka.Avro/AvroSerdeProviderConfig.cs
```diff
@@ -0,0 +1,80 @@
...
+            ///     array returned by the Serialize method to estimate an appropriate value.
+            ///     Note: each call to serialize creates a new buffer.
+            ///
+            ///     default: 1024
+            /// </summary>
+            public const string AvroSerializerBufferBytes = "avro.serializer.buffer.bytes";
+
+            /// <summary>
+            ///     Specifies whether or not the Avro serializer should attempt to auto-register
+            ///     unrecognized schemas with Confluent Schema Registry.
```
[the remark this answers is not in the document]
mhowlett: you have a good point. i think the word unrecognized adds value though since it doesn't add to confusion, and potentially prevents some.

src/Confluent.Kafka.Avro/AvroSerdeProviderConfig.cs
```diff
@@ -0,0 +1,80 @@
...
+        public new static class PropertyNames
+        {
+            /// <summary>
+            ///     Specifies the initial size (in bytes) of the buffer used for Avro message
+            ///     serialization. Use a value high enough to avoid resizing the buffer, but
+            ///     small enough to avoid excessive memory use. Inspect the size of the byte
+            ///     array returned by the Serialize method to estimate an appropriate value.
+            ///     Note: each call to serialize creates a new buffer.
+            ///
+            ///     default: 1024
```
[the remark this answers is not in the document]
mhowlett: yep. now tracking this in the 1.0 todo list.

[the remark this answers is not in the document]
mhowlett: also related to this are changes under consideration for the serializer api in which the .net client may do it's own memory management rather than relying on the GC.

src/Confluent.Kafka/Config.cs
mhowlett: tracking in #614

src/Confluent.Kafka/Config.cs
mhowlett: tracking in #614

src/Confluent.Kafka/Config.cs
mhowlett: A dictionary - by definition - does not retain ordering. A configuration interface that relies on ordering is unusual and necessarily more difficult to use / less terse. we should discuss this.

src/Confluent.Kafka/Config.cs
mhowlett: As part of a longer name for sure. Maybe not for an acronym in isolation. Added to #614 as a todo to check

src/Confluent.Kafka/Deserializers.cs
mhowlett: Agree this is not good, but I think it's the best of the standard exception types and probably don't want to invent another one.

https://stackoverflow.com/questions/774104/what-exceptions-should-be-thrown-for-invalid-or-unexpected-parameters-in-net

src/Confluent.Kafka/Deserializers.cs
mhowlett: most likely. it's the same as previous serializers. I'll add it as a todo to #614 to double check.

src/Confluent.Kafka/Producer.cs
mhowlett: nope

src/Confluent.Kafka/Serializers.cs
mhowlett: would be less efficient (though less error prone).

src/Confluent.Kafka/Serializers.cs
mhowlett: added to #614. I probably did this as a perf optimization, but it's probably not significant. reflects previous code.

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
mhowlett: good idea

test/Confluent.Kafka.Avro.UnitTests/Configuration.cs
mhowlett: but non-standard and the new way of doing things is to flip where the conversion happens to strongly typed config classes on top.

src/Confluent.SchemaRegistry/Rest/RestService.cs
mhowlett: `Produce`

examples/AvroGeneric/Program.cs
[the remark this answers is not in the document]
edenhill: Okay, loosing events is not okay though, so this needs to be solved prior to release.
You could queue events until a delegate is set, but if the application never sets one they'll just queue up. Using a timeout, say 60s, is not foolproof either because the application might set up the Kafka client, do something else for a prolonged period of time, and then try to use it.

API delicacy is all good, but not at the expense of correctness.

src/Confluent.Kafka/Config.cs
edenhill: I agree, but `plugin.library.paths` is an exception that we'll need to support.
The Python and Go clients extract and apply plugin.library.paths before any other configuration is set, which is probably a good workaround. (cc @rnpridgeon )
https://github.com/confluentinc/confluent-kafka-python/blob/master/confluent_kafka/src/confluent_kafka.c#L1582

PR #623, opened 2018-09-12
strongly typed config

mhowlett:
some notes:

- @edenhill - please check the config class carefully. ideally, we want to minimize the strongly typed properties / enum values etc as much as possible I think - exclude legacy / those not relevant to the .net api. 
- error callback delegate type tbd.
- the serdes now don't implement `IDisposable`.  before 1.0, i expect to remove `ISerializer` and `IDeserializer` and replace them with delegates (refer to #618), but this is not impactful to the public api, and it's a fair bit of effort so it can wait 'till after beta.

note: this pr is blocking 1.0-beta since it's a pretty impactful change and we want to include it before encouraging people to use the new api.


qed-:
It would be great if this would fit in with the options pattern in .net core:
https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-2.1

I think it would need public getters on the properties for this to work.

mhowlett:
thanks @qed- - we do indeed need to provide getters otherwise the properties won't bind. i had assumed this wasn't the case (hadn't tested it yet).

i added a small commit to this PR which provides getters that will throw an exception if the property hasn't been set. this is also evidently not good enough.

since we don't want to / can't make the `Config` class `IDisposable`, we'll need to grab the defaults for all properties from librdkafka in the constructor in order to implement getters. this creates a dependency on loading librdkafka.so/.dll (happens automatically) before the `Config` class can be used. I don't see that this could cause any issues, but maybe i'm missing some edge case.

i think we should track defaults / user set values separately and when enumerated over, the client config classes should expose only the user set options.

https://github.com/edenhill/librdkafka/pull/2001 has annoying implications to the config classes. the plan there is to have one or more defaults magically change if `enable.idempotence` is set to true. e.g. `acks` would change from 1 to `all`. in order to provide a completely consistent implementation here, I think we'll need to send the user specified configs over to librdkafka, and re-retrieve all the values on EVERY property set, or duplicate the logic in librdkafka in .net.

This also means the config classes are not POCOs, which is what the options pattern is intended to work with. I did some basic testing that indicates this doesn't seem to be a problem, but I haven't checked everything, and even if it works maybe it's not future proof. It's probably going to be ok, but I don't have full confidence.

qed-:
@mhowlett I wrote something similar that works around the defaults problem by using nullable types for all the properties and returning null if the setting isn't in the dictionary. This works fine with the .Net core binding.

mhowlett:
@qed- - i did consider that earlier but discounted it out as undesirable (non-conventional). You are right though that the semantics are appropriate to the task at hand. I'll reconsider it.

mhowlett:
There is some displeasure internally around the mapping the period separated names to C# conventions in config files (e.g. `retry.backoff.ms` to `RetryBackoffMs`). @qed- do you have any opinion / input on this? I made the config classes `DataContract`s, though I'm not sure this will be really useful anywhere in practice - i was hoping to find something similar that was compatible with the option pattern config, but didn't see anything.

qed-:
I also chose C# conventions. The config will typically be expressed as a JSON fragment where upper camel case looks more natural. You can also deal with the config programatically via first class properties rather than generic get/set methods.
Maybe support both, a strongly typed .Net style POCO with some helper method to initialise from a dictionary of strings.

mhowlett:
@qed- yep, I have both `Get` and `Set` methods in the current implementation, allow initialization from an `IEnumerable<KeyValuePair<string, string>>` and the config classes themselves implement `IEnumerable<KeyValuePair<string, string>>`. So it's a hybrid, which pretty much strictly adds functionality over just using a `Dictionary<string,string>`.

People don't have to use the option pattern, they just can. some people might actually see this as a downside though - it makes it easy to use property names that differ from the standard librdkafka ones, therefore promoting this. they would argue that having consistent property names is more important than being idiomatic + providing convenience to .net developers. This is not my view though. Importantly, there is a clear mapping between the two property name conventions (enforced programmatically).

examples/AvroBlogExamples/Program.cs
```diff
@@ -35,13 +35,10 @@ class Program
...
-            var config = new Dictionary<string, object>
-            {
-                { "bootstrap.servers", bootstrapServers },
-                { "schema.registry.url", schemaRegistryUrl }
-            };
+            var producerConfig = new ProducerConfig { BootstrapServers = bootstrapServers };
 
-            using (var producer = new Producer<Null, GenericRecord>(config, null, new AvroSerializer<GenericRecord>()))
+            using (var serdeProvider = new AvroSerdeProvider(new AvroSerdeProviderConfig { SchemaRegistryUrl = schemaRegistryUrl }))
+            using (var producer = new Producer<Null, GenericRecord>(producerConfig, null, serdeProvider.CreateValueSerializer<GenericRecord>()))
```
mhowlett: it's not great that the type system isn't helping ensure `CreateValueSerializer` isn't being used to create a key serializer. also, there's no run time check of this currently. doing something better here is a todo before 1.0. the 'provider' idea is much better than having the serdes IDisposable though, and broadly speaking the way forward.

README.md
```diff
@@ -93,7 +93,7 @@ class Program
         var config = new Dictionary<string, object> { { "bootstrap.servers", "localhost:9092" } };
 
         // A Producer for sending messages with null keys and UTF-8 encoded values.
-        using (var p = new Producer<Null, string>(config, null, new StringSerializer(Encoding.UTF8)))
+        using (var p = new Producer<Null, string>(config))
```
mhowlett: I need to update the readme examples. just noticed this.

PR #632, opened 2018-09-26
example code cleanup

mhowlett:
quick cleanup. most importantly, the namespace `Confluent.Kafka.Serialization` actually no longer exists, so the examples on the README.md won't compile as is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,11 +81,8 @@ optimizing communication with the Kafka brokers for you, batching requests as ap
 
 ```csharp
 using System;
-using System.Collections.Generic;
-using System.Text;
 using System.Threading.Tasks;
 using Confluent.Kafka;
-using Confluent.Kafka.Serialization;
 
 class Program
 {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -183,9 +178,11 @@ class Program
                 catch (ConsumeException e)
                 {
                     Console.WriteLine($"Error occured: {e.Error.Reason}");
+                    break;
```
edenhill: I thought we dont want to break out on errors.

mhowlett: forgot about that. the reason I keep trying to put this in is the `Close` at the end is pointless as there is no exit condition on the loop. but we want this in the example so people know to use it. we also don't want to exit the loop as you say (now you point this out, again). i'll take out the break, and have the pointless `Close` rather than make the example longer.

mhowlett:
ok, I've added an exit condition to the consume loop in the case of fatal errors. here's why I think that is a good idea:

1. The consumer currently never emits a fatal error. However it may in the future, and having people consider this possibility in their code now is good future proofing.
2. It documents how people are supposed to respond to error events - a common source of confusion. this knowledge transfers between client types.
3. It keeps the examples terse and focused on documenting kafka client specific things c.f. having Ctrl-C handling garbage which is verbose in c# and off topic.

README.md
edenhill: Ctrl-C is a good way to show-case termination.

mhowlett:
Ctrl-C is not my preferred option here as noted.

PR #640, opened 2018-10-01
ProduceAsync not disposing CancellationTokenRegister object

DHirani:
Fixed a memory leak bug where ProduceAsync was NOT disposing CancellationTokenRegister object when calling Register.

### DIFF
--- a/src/Confluent.Kafka/Producer.cs
+++ b/src/Confluent.Kafka/Producer.cs
@@ -564,19 +564,20 @@ namespace Confluent.Kafka
                     producer.enableDeliveryReportKey ? message.Key : default(TKey),
                     producer.enableDeliveryReportValue ? message.Value : default(TValue));
 
-                cancellationToken.Register(() => handler.TrySetException(new TaskCanceledException()));
+                using (cancellationToken.Register(() => handler.TrySetException(new TaskCanceledException())))
+                {
+                    var keyBytes = keySerializer(topic, message.Key);
+                    var valBytes = valueSerializer(topic, message.Value);
 
-                var keyBytes = keySerializer(topic, message.Key);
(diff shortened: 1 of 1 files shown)


DHirani:
Once merged can we urgently have a new BETA build as this bug is really serious as my process hit 4GB of memory

mhowlett:
thanks for that @DHirani and sorry for the inconvenience!

AppVeyor automatically builds the nuget packages for all pull requests and makes them available at the following package source: https://ci.appveyor.com/nuget/confluent-kafka-dotnet . i'd recommend using that until we do a beta2. The version corresponding to this pr is `1.0.0-beta-ci-663`. You can also download the artifacts here: https://ci.appveyor.com/project/ConfluentClientEngineering/confluent-kafka-dotnet/build/1.0.663/artifacts

DHirani:
Do you have an ETA when beta 2 will be available as our build server can only access packages from Nuget?

mhowlett:
can you sign the contributor agreement so I can merge? 

Ok, I will try to do it sooner rather than later. Additionally, i'll reference a librdkafka with idempotent producer + sparse connection capability. I really wanted to finalize the serializer types as well (these are going to change), but this'll probably take a while and this bug is urgent enough.

DHirani:
Signed

mhowlett:
no breaking changes from 1.0-beta-ci-663 to 1.0-beta2.

there will be breaking changes before 1.0. the biggest will be changes to serialization (not yet finalized). however they shouldn't be too difficult to resolve.

mhowlett:
i'm not sure this is exactly what we want since the using block doesn't extend over the lifetime of the produce request, but it'll certainly stop the memory leak, so merging (todo: revisit this).

PR #645, opened 2018-10-15
updated strong config

mhowlett:
`Config.cs` generated from librdkafka v1.0.0-PRE1.

### DIFF
--- a/src/Confluent.Kafka/Config.cs
+++ b/src/Confluent.Kafka/Config.cs
@@ -382,7 +382,7 @@ namespace Confluent.Kafka
         public int? MessageCopyMaxBytes { get { return GetInt("message.copy.max.bytes"); } set { this.SetObject("message.copy.max.bytes", value); } }
 
         /// <summary>
-        ///     Maximum Kafka protocol response message size. This serves as a safety precaution to avoid memory exhaustion in case of protocol hickups. This value is automatically adjusted upwards to be at least `fetch.max.bytes` + 512 to allow for protocol overhead.
+        ///     Maximum Kafka protocol response message size. This serves as a safety precaution to avoid memory exhaustion in case of protocol hickups. This value must be at least `fetch.max.bytes`  + 512 to allow for protocol overhead; the value is adjusted automatically unless the configuration property is explicitly set.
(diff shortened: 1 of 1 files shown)


src/Confluent.Kafka/Config.cs
```diff
@@ -432,7 +432,7 @@ protected void SetObject(string name, object val)
         public int? SocketTimeoutMs { get { return GetInt("socket.timeout.ms"); } set { this.SetObject("socket.timeout.ms", value); } }
 
         /// <summary>
-        ///     Maximum time a broker socket operation may block. A lower value improves responsiveness at the expense of slightly higher CPU usage. **Deprecated**
+        ///     **DEPRECATED** No longer used.
```
edenhill: Deprecated should probably be filtered out completely.

src/Confluent.Kafka/Config.cs
```diff
@@ -472,19 +472,29 @@ protected void SetObject(string name, object val)
...
+        /// <summary>
+        ///     The maximum time to wait before reconnecting to a broker after the connection has been closed.
+        /// </summary>
+        public int? ReconnectBackoffMaxMs { get { return GetInt("reconnect.backoff.max.ms"); } set { this.SetObject("reconnect.backoff.max.ms", value); } }
 
         /// <summary>
-        ///     Logging level (syslog(3) levels)
+        ///     librdkafka statistics emit interval. The application also needs to register a stats callback using `rd_kafka_conf_set_stats_cb()`. The granularity is 1000ms. A value of 0 disables statistics.
         /// </summary>
-        public int? Log_Level { get { return GetInt("log_level"); } set { this.SetObject("log_level", value); } }
```
whschultz: Should this have been removed?  It still exists in the [librdkafka documentation](https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md).

mhowlett: yeah, I noticed that. I assumed it was for good reason (this is a useless property in practice, and therefore confusing - use `debug` instead), but don't quickly see why. will investigate.

PR #656, opened 2018-10-24
revamped serialization

mhowlett:
key points:

- `Producer` and `Consumer` no longer have type parameters.
- `ProduceAsync`, `BeginProduce`, and `Consume` have both typed and untyped variants.
- Serdes were not separated from the `Producer`/`Consumer` entirely because:
  - `Producer`: There is often a requirement that delivery reports contain the corresponding object that was produced. If the serialized message is provided to the producer, the original unserialized values are not available to include in the delivery report.
  - `Consumer`: If the `Consume` method is deserializer aware, a copy to `byte[]` can be avoided. This results in a significant performance improvement.
- Serializers and deserializers are registered with the `Producer` and `Consumer`. invocation is automatic, based on type. This results in a very terse, easy to use API.
- The serde delegates that the `Producer` and `Consumer` work with are not sufficient to provide functionality required by avro serdes (they lack a topic parameter and can not be awaited). Avro serde functionality is a completely separate layer, implemented via inheritance.
- The avro serdes now never block when awaited (it's async/await all the way down). Associated deadlock issues should be solved. This was not possible with any of the previous serialization implementations.
- performance: to localhost: 460k 100 byte msgs/s using `BeginProduce`, and 380k 100 byte msgs/s using `Consume<Ignore,Ignore>`. This is on par with the previous implementation.

PR #662, opened 2018-10-29
Add Close() to IConsumer interface

jvaesteves:
When I tried to use IConsumer, from the version 1.0.0-beta2 package, through DI in a project, I realized that the Close() method, meant to actually finalize a consumer, was not available in the interface, just in the concrete class.


jvaesteves:
[clabot:check]

mhowlett:
thanks for the PR @jvaesteves! - i'm closing since I've already slipped this into #656, and it'll be in -beta3.

PR #684, opened 2018-11-21
convert static properties

### LINKED ISSUE #682: [Suggestion] Use "static readonly" for const like properties
Description
===========
This is more of a coding pattern / perf suggestion.
In classes like `Offset` there are static properties that create new instances. eg
`public static Offset Beginning { get { return new Offset(RD_KAFKA_OFFSET_BEGINNING); } }`

This pattern means that a new object instance is created everytime the property is used. As `Offset` is a readonly type is would be "better" if the same instance was returned every time. ie

`public static readonly Offset Beginning = new Offset(RD_KAFKA_OFFSET_BEGINNING);`
Or, if you really want it to look like a property (but prefer the above)
`public static Offset Beginning { get; } = new Offset(RD_KAFKA_OFFSET_BEGINNING); `

This makes the property essentially free rather than allocating and GC'ing every time.

How to reproduce
================

`var x = Offset.Beginning`
Creates a new instance of `Offset`

Checklist
=========

 - [ ] Confluent.Kafka nuget version:
 - [ ] Apache Kafka version: 
 - [ ] Client configuration:
 - [ ] Operating system:
 - [ ] Provide logs (with "debug" : "..." as necessary in configuration)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

AndyPook:
resolves #682

TL;DR:
Convert
`public static Offset Beginning { get { return new Offset(RD_KAFKA_OFFSET_BEGINNING); } }`
to
`public static readonly Offset Beginning = new Offset(RD_KAFKA_OFFSET_BEGINNING);`

### DIFF
--- a/src/Confluent.Kafka/Offset.cs
+++ b/src/Confluent.Kafka/Offset.cs
@@ -18,7 +18,6 @@
 
 using System;
 
-
 namespace Confluent.Kafka
 {
     /// <summary>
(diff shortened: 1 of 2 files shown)


AndyPook:
[clabot:check]

mhowlett:
LGTM, will merge after #656 is in.

mhowlett:
Finally getting to this -

random thoughts:
1. `Offset` is a struct, so GC isn't a concern.
2. The runtime may very well optimize the overhead of the property variant away.
3. The standard library uses properties almost exclusively.
4. But it doesn't for `Int64.MaxValue`, and I think this is pretty analogous to that.

so LGTM! Thanks!

Merging to master directly as there are problems going to 1.0.x, so this won't make it in the next release, but the one after.

PR #706, opened 2018-12-17
Make parsing of enum values in config case insensitive

wheerd:
There is a bug where you cannot read an enum config value after setting it:
```c#
var config = new ConsumerConfig
{
    AutoOffsetReset = AutoOffsetResetType.Earliest
};

Assert.AreEqual(config.AutoOffsetReset, AutoOffsetResetType.Earliest);
```
Because the value is converted to lower case it cannot be parsed afterwards.

### DIFF
--- a/src/Confluent.Kafka/Config.cs
+++ b/src/Confluent.Kafka/Config.cs
@@ -290,7 +290,7 @@ namespace Confluent.Kafka
         {
             var result = Get(key);
             if (result == null) { return null; }
-            return Enum.Parse(type, result);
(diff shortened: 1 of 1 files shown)


wheerd:
[clabot:check]

mhowlett:
thanks for contributing @wheerd ! but you were beaten to it and this fix has been merged up to 1.0.x (which i'll release soon and merge up to master). not surprised you didn't see as we've been working a few branches deep for a while.. thanks again!

wheerd:
That is awesome to hear, thank you 👍

PR #710, opened 2018-12-21
Demo should show consumed message Value

miticv:


miticv:
[clabot:check]

mhowlett:
thanks for the suggestion! unfortunately avro classes don't implement `ToString`, so this will just print the class name. I will change the word choice here though to be less confusing.

PR #722, opened 2019-01-04
various enum related changes

mhowlett:
- Audited all the enum types in Confluent.Kafka for compliance with .net naming standards and made appropriate changes.
- Changed the `acks` configuration parameter to an enum (of type int32). Note that it's valid in C# to set an enum type to a value that isn't one of it's discrete names (e.g. Acks = 2), so i think this has precisely the semantics we want, also when considering possible future requirements.

@edenhill @rnpridgeon

### DIFF
--- a/src/ConfigGen/Program.cs
+++ b/src/ConfigGen/Program.cs
@@ -149,14 +149,14 @@ namespace ConfigGen
     public enum Acks : int
     {
         /// <summary>
-        ///     Zero
+        ///     None
         /// </summary>
-        Zero = 0,
+        None = 0,
 
         /// <summary>
-        ///     One
+        ///     Leader
         /// </summary>
-        One = 1,
+        Leader = 1,
 
         /// <summary>
         ///     All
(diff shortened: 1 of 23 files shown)


src/Confluent.Kafka/Config.cs
```diff
@@ -131,6 +139,10 @@ protected object GetEnum(Type type, string key)
         {
             var result = Get(key);
             if (result == null) { return null; }
+            if (EnumNameToConfigValueSubstitutes.Values.Count(v => v == result) > 0)
```
mhowlett: it's getting a little yucky that this is here, but imo it doesn't matter and is appropriate.

src/ConfigGen/Program.cs
```diff
@@ -108,35 +112,115 @@ public enum SaslMechanismType
...
+    {
+        /// <summary>
+        ///     Zero
+        /// </summary>
+        Zero = 0,
+
+        /// <summary>
+        ///     One
+        /// </summary>
+        One = 1,
```
edenhill: Leader is probably better.

src/ConfigGen/Program.cs
```diff
@@ -108,35 +112,115 @@ public enum SaslMechanismType
...
+        /// </summary>
+        internal static string AcksEnumString =>
+@"
+    /// <summary>
+    ///     Acks enum values
+    /// </summary>
+    public enum Acks : int
+    {
+        /// <summary>
+        ///     Zero
```
edenhill: None

mhowlett:
i like your thinking, updated.

Issue #679, opened 2018-11-14
SchemaRegistryBasicAuthCredentialsSource is missing from SchemaRegistryConfig in 1.0.0-beta2


ghost:
Description
===========
SchemaRegistryBasicAuthCredentialsSource is missing from SchemaRegistryConfig in 1.0.0-beta2

How to reproduce
================

Checklist
=========
Please provide the following information:

 - [ ] Confluent.Kafka nuget version:
 - [ ] Apache Kafka version: 
 - [ ] Client configuration:
 - [ ] Operating system:
 - [ ] Provide logs (with "debug" : "..." as necessary in configuration)
 - [ ] Provide broker log excerpts
 - [ ] Critical issue

mhowlett:
thanks for reporting @prem-aiven-io ! just opened PR #728 to fix this.

anchitj:
Closing as the fix is merged

PR #728, opened 2019-01-04
added SchemaRegistryConfigBasicAuthCredentialsSource

mhowlett:
addresses #679. this was accidentally omitted from 1.0-beta2, probably because at one point i was considering not including it.

### DIFF
--- a/src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
+++ b/src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
@@ -81,29 +81,22 @@ namespace Confluent.SchemaRegistry
             // Convert.ToString returns "" in the null case here.
             var basicAuthSource = Convert.ToString(config.FirstOrDefault(prop => prop.Key.ToLower() == SchemaRegistryConfig.PropertyNames.SchemaRegistryBasicAuthCredentialsSource).Value);
             var basicAuthInfo = Convert.ToString(config.FirstOrDefault(prop => prop.Key.ToLower() == SchemaRegistryConfig.PropertyNames.SchemaRegistryBasicAuthUserInfo).Value);
-            if (basicAuthInfo != "" && basicAuthSource == "")
-            {
-                // default to USER_INFO if no source specified, since the strongly typed config
-                // class doesn't even expose the config source property.
-                basicAuthSource = "USER_INFO";
-            }
 
             string username = null;
             string password = null;
 
             if (basicAuthSource == "USER_INFO")
             {
-                if (basicAuthInfo == "")
+                if (basicAuthInfo != "")
                 {
-                    throw new ArgumentException($"CachedSchemaRegistryClient: {SchemaRegistryConfig.PropertyNames.SchemaRegistryBasicAuthCredentialsSource} set to 'USER_INFO', but no value specified for {SchemaRegistryConfig.PropertyNames.SchemaRegistryBasicAuthUserInfo}.");
+                    var userPass = (basicAuthInfo).Split(':');
+                    if (userPass.Length != 2)
+                    {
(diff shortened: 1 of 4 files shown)


src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
```diff
@@ -53,6 +53,8 @@ public static class PropertyNames
 
             /// <summary>
             ///     Specifies the configuration property(ies) that provide the basic authentication credentials.
+            ///     USER_INFO: credentials are specified via the `schema.registry.basic.auth.user.info` config property in the form username:password.
+            ///     SASL_INHERIT: credentials are specified via the `sasl.username` and `sasl.password` configuration properties.
```
edenhill: Shouldn't this be an enum thing?

mhowlett: it shouldn't be here - this string values would be best to match the JAVA config values. But it should be in the static config class, following the pattern of other configs, and the property should translate between the two names.

good catch.

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -81,29 +81,22 @@ public CachedSchemaRegistryClient(IEnumerable<KeyValuePair<string, string>> conf
...
-                if (basicAuthInfo == "")
-                {
-                    throw new ArgumentException($"CachedSchemaRegistryClient: {SchemaRegistryConfig.PropertyNames.SchemaRegistryBasicAuthCredentialsSource} set to 'USER_INFO', but no value specified for {SchemaRegistryConfig.PropertyNames.SchemaRegistryBasicAuthUserInfo}.");
-                }
-                var userPass = (basicAuthInfo).Split(':');
-                if (userPass.Length != 2)
+                if (basicAuthInfo != "")
                 {
-                    throw new ArgumentException($"CachedSchemaRegistryClient: Configuration property {SchemaRegistryConfig.PropertyNames.SchemaRegistryBasicAuthUserInfo} must be of the form 'username:password'.");
+                    var userPass = (basicAuthInfo).Split(':');
```
edenhill: We should support escaping the :

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -126,7 +119,7 @@ public CachedSchemaRegistryClient(IEnumerable<KeyValuePair<string, string>> conf
             }
             else
             {
-                // no basic auth info.
+                throw new ArgumentException($"CachedSchemaRegistryClient: Invalid value '{basicAuthSource}' specified for property '{SchemaRegistryConfig.PropertyNames.SchemaRegistryBasicAuthCredentialsSource}'");
```
edenhill: nit: hint of the correct values.

src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
```diff
@@ -14,12 +14,30 @@
...
+        ///     Credentials are specified via the `schema.registry.basic.auth.user.info` config property in the form username:password.
+        ///     If `schema.registry.basic.auth.user.info` is not set, authentication is disabled.
+        /// </summary>
+        UserInfo,
+
+        /// <summary>
+        ///     Credentials are specified via the `sasl.username` and `sasl.password` configuration properties.
+        /// </summary>
+        SaslInherit
+    }
```
edenhill: what about retrieving it from the URL?

mhowlett: left as a todo. I had problems getting this to work and just omitted it in the end.

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
milindl: For the sake of documenting this - the RFC for basic auth disallows any colon in the username, and rather everything after the first colon should be considered to be a part of the password.

See #1767

PR #736, opened 2019-01-11
Moved ConsumeResult Commit methods to correct place

mhowlett:
also changed the return types to `void`, as returning something effectively provided by the user was pretty pointless and for consistency with the base method used to implement.

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -266,6 +266,38 @@ namespace Confluent.Kafka
             => (keyDeserializer != null && valueDeserializer != null)
                 ? Consume<TKey, TValue>(timeout.TotalMillisecondsAsInt(), keyDeserializer, valueDeserializer) // fast path for simple case
                 : Consume(timeout.TotalMillisecondsAsInt());
+
+
+        /// <summary>
+        ///     Commits an offset based on the topic/partition/offset of a ConsumeResult.
+        /// </summary>
+        /// <param name="result">
+        ///     The ConsumeResult instance used to determine the committed offset.
+        /// </param>
+        /// <param name="cancellationToken">
+        ///     A cancellation token that can be used to cancel this operation
+        ///     (currently ignored).
+        /// </param>
+        /// <exception cref="Confluent.Kafka.KafkaException">
+        ///     Thrown if the request failed.
+        /// </exception>
+        /// <exception cref="Confluent.Kafka.TopicPartitionOffsetException">
+        ///     Thrown if the result is in error.
+        /// </exception>
+        /// <remarks>
+        ///     A consumer which has position N has consumed messages with offsets up to N-1 
+        ///     and will next receive the message with offset N. Hence, this method commits an 
+        ///     offset of <paramref name="result" />.Offset + 1.
+        /// </remarks>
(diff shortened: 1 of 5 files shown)


src/Confluent.Kafka/Consumer.cs
```diff
@@ -266,6 +266,38 @@ IAsyncDeserializer<TValue> taskValueDeserializer
...
+        ///     Thrown if the request failed.
+        /// </exception>
+        /// <exception cref="Confluent.Kafka.TopicPartitionOffsetException">
+        ///     Thrown if the result is in error.
+        /// </exception>
+        /// <remarks>
+        ///     A consumer which has position N has consumed messages with offsets up to N-1 
+        ///     and will next receive the message with offset N. Hence, this method commits an 
+        ///     offset of <paramref name="result" />.Offset + 1.
+        /// </remarks>
```
edenhill: There is no indication, apart from the unused cancellation token,  if this is a synchronous/blocking call or not

src/Confluent.Kafka/Consumer.cs
```diff
@@ -266,6 +266,38 @@ IAsyncDeserializer<TValue> taskValueDeserializer
...
+        public void Commit(ConsumeResult<TKey, TValue> result, CancellationToken cancellationToken = default(CancellationToken))
+        {
+            if (result.Message == null)
+            {
+                throw new InvalidOperationException("Attempt was made to commit offset corresponding to an empty consume result");
+            }
+
+            Commit(new [] { new TopicPartitionOffset(result.TopicPartition, result.Offset + 1) });
```
edenhill: Is an exception raised on commit error?

src/Confluent.Kafka/Consumer.cs
mhowlett: This is idiomatic, pre `async`/`await`. e.g. `System.IO.File.ReadAllLines`. Ideally I'd implement this only as an async method, with cancellation token, and this would be clear. I really don't like that there's no distinction as well. It will become clear though when Async variants are implemented, because the blocking functions will all have Async equivalents. The API will be analogous to `System.IO.File` (and others).

Noteworthy: in the next unopened PR, which brings consistency to all the cancellation tokens / timeout parameters across the library, and i think the best thing to do right now is remove all the unused cancellation tokens (and consider adding them, with defaults, a non-breaking change).

src/Confluent.Kafka/Consumer.cs
mhowlett: yes

mhowlett: for your reference: https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs#L688

all the different error cases throw an appropriate exception. whilst we're on the topic though, we need to introduce more exception types. These will derive from `KafkaException` and `TopicPartitionOffsetException`. Since doing that is not a breaking change, I think an overhaul of exceptions can come after v1.0.

src/Confluent.Kafka/Consumer.cs
edenhill: I don't think it hurts with a small doc string for the blocking behaviour.

mhowlett:
yep, good idea - even prefix the summary with [blocking call] or something. have made a note of that, and will do this consistently in the PR that fixes all this up more broadly.

PR #737, opened 2019-01-11
fixed failover connection issue

### LINKED ISSUE #711: Schema Registry URL redundancy does not work as expected
There is an issue with referencing two urls in “schema.registry.url”.

Here are the cases we have tested:

- Case 1: {"schema.registry.url", "good_url,good_url"}

o The producer client works correctly and send a message to Kafka

- Case 2: {"schema.registry.url", "good_url,bad_url"}

o The producer client works correctly and send a message to Kafka

- Case 3: {"schema.registry.url", "bad_url,good_url"}

o The producer client crash with the following error

“The request message was already sent. Cannot send the same request message multiple times.”

mhowlett:
resolves #711 

verified manually - we should add tests for this case, but that would complicate the dependencies required (an additional SR instance) making it less trivial to start up. we should do this, but after or as part of the task to automate execution of the integration tests. i'll add an issue to track this.

The problem was `HttpClient` doesn't allow the same request instance to be sent more than once. I don't like this behavior TBH. One of a growing number of things I don't like about `HttpClient`.

### DIFF
--- a/src/Confluent.SchemaRegistry/Rest/RestService.cs
+++ b/src/Confluent.SchemaRegistry/Rest/RestService.cs
@@ -71,7 +71,7 @@ namespace Confluent.SchemaRegistry
 
         #region Base Requests
 
-        private async Task<HttpResponseMessage> ExecuteOnOneInstanceAsync(HttpRequestMessage request)
+        private async Task<HttpResponseMessage> ExecuteOnOneInstanceAsync(Func<HttpRequestMessage> createRequest)
         {
             // There may be many base urls - roll until one is found that works.
             //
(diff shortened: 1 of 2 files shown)


edenhill:
I do feel a test is warranted. Can't you just point the first address to "http://localhost/65432" or some other high port that is unlikely to be bound (and even if it is, unlikely to be a running SR), or point it at "http://google.tazmania"

rnpridgeon:
Adding an integration test for all of the scenarios described in #711 still seems beneficial to me. If you use docker-compose adding an additional SR instance is fairly trivial.

edenhill:
And testing "bad_sr,bad_sr2" would be possible without any cluster.

mhowlett:
you're completely right, i can just point to a non-existent place to test. I did actually realize this having written the above but was in a hurry. will add, ok. and yes, setting up docker-compose is easy and the extent of the tests is getting to the point where this is beneficial, but something to do in a different pr.

PR #743, opened 2019-01-16
Fix when not installed librdkafka.redist cannot load librdkafka.dll

pengweiqhca:
Fix when not installed librdkafka.redist cannot load librdkafka.dll in WebApplication project
tim 20190116132823


mhowlett:
thanks @pengweiqhca  - what is the reason to limit this to `NETFRAMEWORK`? do you know of a scenario where this is not desirable?

pengweiqhca:
Now I only found `NETFRAMEWORK` need to copy librdkafka.dll to the bin directory, `NETCORE` doesn't need to copy. But also can remove the limit.

mhowlett:
thanks @pengweiqhca! i've had chance to look into this now, and it'll be in the next release.

PR #751, opened 2019-01-24
Message<byte[],byt[]> -> Message

### LINKED ISSUE #747: ConsumeResult should not derive from ConsumeResult<T,V>. Also DRs
`ConsumeResult` derives from `ConsumeResult<byte[],byte[]>`. This is a problem because the `Message` property has type `Message<byte[],byte[]>` not `Message`, which is not correct and as a consequence makes it inconvenient to use in `Consume` then `Produce` use cases.

The same is true for `DeliveryReport` and `DeliveryResult`.

side note: at this point, it's fairly clear there is little value in providing the non-typed `Producer` and `Consumer` variants (this has not always been clear + there have been some strong advocates for). It's substantial work to remove them though because many of the tests make use of them, so they can remain.

mhowlett:
resolves #747 

The non-generic `ConsumeResult`, `DeliveryReport` and `DeliveryResult` classes can't derive from the generic `<byte[],byte[]>` equivalents because the `Message` property should have type `Message` not `Message<byte[],byte[]>`. I could have made some base classes to cut down on the code duplication in this PR, but these would increase the complexity of the class hierarchy from a user perspective and so this is not desirable. Also, the amount of duplication is minimal and it's nature won't cause a meaningful maintenance overhead.

The code actually was like this previously, but a contributor made the 'improvement' and i wasn't quick enough at the time to realize the consequences.

As noted in #747, I'm now also quite sure the non-generic producer / consumer are a waste of energy, but they are a little bit useful, harmless, somewhat analogous to the generic/non generic collection classes, and would take quite a number of loc changes to extract from the integration tests, so I think they should stay.

note: unit + integration tests pass.

### DIFF
--- a/src/Confluent.Kafka/ConsumeResult.cs
+++ b/src/Confluent.Kafka/ConsumeResult.cs
@@ -22,7 +22,95 @@ namespace Confluent.Kafka
     /// <summary>
     ///     Represents a message consumed from a Kafka cluster.
     /// </summary>
-    public class ConsumeResult : ConsumeResult<byte[], byte[]> {}
+    public class ConsumeResult
+    {
+        /// <summary>
+        ///     The topic associated with the message.
+        /// </summary>
+        public string Topic { get; set; }
+
+        /// <summary>
+        ///     The partition associated with the message.
+        /// </summary>
+        public Partition Partition { get; set; }
+
+        /// <summary>
+        ///     The partition offset associated with the message.
+        /// </summary>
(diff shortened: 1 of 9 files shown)


mhowlett:
appveyor made a lier of me. test failure due to last minute clean up of the test that could not possibly have gone wrong.

mhowlett:
we've decided to get rid of the `byte[]`, `byte[]` client variants.

PR #760, opened 2019-02-04
remove non-serializing `Producer` and `Consumer` classes

mhowlett:
a couple of simple little changes:
- Removed the non-serializing `Producer` and `Consumer` classes (and related).
- Merged `ProducerBase` and `ConsumerBase` into `Producer` and `Consumer`.


src/Confluent.Kafka/IProducer.cs
```diff
@@ -103,5 +67,23 @@ public interface IProducer<TKey, TValue> : IProducerBase, IClient
...
+        /// <summary>
+        ///     Refer to <see cref="Confluent.Kafka.Producer{TKey,TValue}.Flush(TimeSpan)" />
+        /// </summary>
+        int Flush(TimeSpan timeout);
+
+
+        /// <summary>
+        ///     Refer to <see cref="Confluent.Kafka.Producer{TKey,TValue}.Flush(CancellationToken)" />
+        /// </summary>
+        void Flush(CancellationToken cancellationToken = default(CancellationToken));
```
edenhill: I thought we were not to include the cancellationToken methods until we could properly implement cancellable librdkafka calls, yesno?

src/Confluent.Kafka/Producer.cs
```diff
@@ -98,20 +522,135 @@ public class Producer<TKey, TValue> : ProducerBase, IProducer<TKey, TValue>
...
+            {
+                Librdkafka.conf_set_dr_msg_cb(configPtr, DeliveryReportCallback);
+            }
+
+            // Explicitly keep references to delegates so they are not reclaimed by the GC.
+            errorCallbackDelegate = ErrorCallback;
+            logCallbackDelegate = LogCallback;
+            statisticsCallbackDelegate = StatisticsCallback;
+
+            // TODO: provide some mechanism whereby calls to the error and log callbacks are cached until
```
edenhill: This comment is no longer valid thanks to the builder pattern, right?

mhowlett: yep, well spotted.

PR #763, opened 2019-02-06
SetRebalanceHandler implementation

mhowlett:
1. Replaces `SetPartitionsAssignedHandler` and `SetPartitionsRevokedHandler` with `SetRebalanceHandler`.
2. Enforces that the handler may not be set after `Subscribe` (todo: after unsubscribe and subsequent revoke, this could be re-enabled, but left as todo).
3. Rename the `Assignment` property to `AssignedPartitions`.

Advantages:
1. Allows for consistent and precise use of terminology. The word `Assignment` refers to a consumer group partition assignment, the words `Assign` and `Assigned` to what the consumer is actually assigned to consume from. 
2. `SetRebalanceHandler` is less awkward than `SetPartitionsAssignedHandler` and `SetPartitionsRevokedHandler`.

Disadvantages:
1. Often (most but not all cases) leads to a bit more complex user code with a few more LOC.
2. API is less helpful in preventing unintended usage - user could accidentally assign to a set of partitions on revocation notification instead of un-assign from them.

In making this change, the dominating factor in my mind is advantage number 1. The assignment process is a difficult thing for new people to understand, and we need to be able to explain it clearly.

I feel a bit silly for mentioning this, but I didn't think of the word `revocation` until this PR was nearly finished. We could alternatively just change the names of `SetPartitionsAssignedHandler` and `SetPartitionsRevokedHandler` to `SetPartitionAssignmentHandler` and `SetPartitionRevocationHandler`. In terms of terminology, this is still more awkward than what this PR to me, but it may be good enough. OTOH, this API is only for advanced use cases, or just logging, where this variant is actually terser assuming both logging of assignment and revocation is desired, so not hugely important and I still prefer the variant in this PR.

opinion?


edenhill:
> 2\. Enforces that the handler may not be set after `Subscribe` (todo: after unsubscribe and subsequent revoke, this could be re-enabled, but left as todo).

There is no 1:1 mapping between subscribe/unsubscribe and subsequent rebalances.
The first rebalance following subscribe/unsubscribe could've been enqueued prior to those calls, triggered by some other event (group rebalance).

edenhill:
> We could alternatively just change the names of `SetPartitionsAssignedHandler` and `SetPartitionsRevokedHandler` to `SetPartitionAssignmentHandler` and `SetPartitionRevocationHandler`

I think we should avoid making up new names for existing things. Java and the other clients already have onPartitionsRevoked, on PartitionsAssigned, let's stick to that form of revoked and assigned.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -133,30 +133,30 @@ private void LogCallback(IntPtr rk, SyslogLevel level, string fac, string buf)
 
             if (err == ErrorCode.Local_AssignPartitions)
             {
-                if (partitionsAssignedHandler != null)
+                if (rebalanceHandler != null)
                 {
                     assignCallCount = 0;
-                    partitionsAssignedHandler(partitionList);
+                    rebalanceHandler(new RebalanceEvent(partitionAssignment, true));
                     if (assignCallCount == 1) { return; }
```
edenhill: add a comment saying why we do this

src/Confluent.Kafka/Consumer.cs
```diff
@@ -133,30 +133,30 @@ private void LogCallback(IntPtr rk, SyslogLevel level, string fac, string buf)
 
             if (err == ErrorCode.Local_AssignPartitions)
             {
-                if (partitionsAssignedHandler != null)
+                if (rebalanceHandler != null)
                 {
                     assignCallCount = 0;
```
edenhill: needs locking

src/Confluent.Kafka/Consumer.cs
```diff
@@ -62,6 +62,7 @@ internal class Config
 
         private int cancellationDelayMaxMs;
 
+        private bool subscribeHasBeenCalled = false;
```
edenhill: needs locking

src/Confluent.Kafka/Consumer.cs
```diff
@@ -243,7 +243,10 @@ public List<string> Subscription
         ///     in the group.
         /// </remarks>
         public void Subscribe(IEnumerable<string> topics)
-            => kafkaHandle.Subscribe(topics);
+        {
+            this.subscribeHasBeenCalled = true;
```
edenhill: needs locking

src/Confluent.Kafka/Consumer.cs
```diff
@@ -269,10 +272,38 @@ public void Unsubscribe()
 
 
         /// <summary>
-        ///     Update the assignment set to a single <paramref name="partition" />.
+        ///     Update the set of assigned partitions (the set of partitions the
```
edenhill: update suggests incremental, use Sets or Replaces

src/Confluent.Kafka/Consumer.cs
```diff
@@ -269,10 +272,38 @@ public void Unsubscribe()
...
+                Assign(rebalanceEvent.Partitions);
+            }
+            else
+            {
+                Unassign();
+            }
+        }
+
+        /// <summary>
+        ///     Update the set of assigned partitions (the set of partitions the
```
edenhill: not sure this variant is that useful in practice

src/Confluent.Kafka/Consumer.cs
```diff
@@ -284,10 +315,12 @@ public void Assign(TopicPartition partition)
 
 
         /// <summary>
-        ///     Update the assignment set to a single <paramref name="partition" />.
+        ///     Update the set of assigned partitions (the set of partitions the
```
edenhill: dito , both for wording and neccesity

src/Confluent.Kafka/Consumer.cs
```diff
@@ -695,50 +733,42 @@ protected virtual void Dispose(bool disposing)
...
+        ///     assignment dictated by the consumer group. This default behavior will not occur if
+        ///     you call Assign yourself in the handler. The set of partitions you assign to is not
+        ///     required to match the assignment provided by the consumer group, but typically will.
+        /// 
+        ///     Consumption will resume from the last committed offset for each partition (unless an
+        ///     offset is manually specified in a call to Assign), or if there is no committed
+        ///     offset, in accordance with the `auto.offset.reset` configuration property.
         /// </summary>
         /// <remarks>
         ///     Executes as a side-effect of the Consumer.Consume call (on the same thread).
```
edenhill: may

src/Confluent.Kafka/Consumer.cs
```diff
@@ -695,50 +733,42 @@ protected virtual void Dispose(bool disposing)
...
+        ///     If you do not call an Assign or Unassign method in this handler, or do not specify
+        ///     a handler, partitions will be assigned (or unassigned) automatically, matching the
+        ///     assignment dictated by the consumer group. This default behavior will not occur if
+        ///     you call Assign yourself in the handler. The set of partitions you assign to is not
+        ///     required to match the assignment provided by the consumer group, but typically will.
+        /// 
+        ///     Consumption will resume from the last committed offset for each partition (unless an
+        ///     offset is manually specified in a call to Assign), or if there is no committed
+        ///     offset, in accordance with the `auto.offset.reset` configuration property.
         /// </summary>
```
edenhill: mention it must only be called once

src/Confluent.Kafka/Consumer.cs
```diff
@@ -695,50 +733,42 @@ protected virtual void Dispose(bool disposing)
...
             if (value == null)
             {
-                partitionsAssignedHandler = null;
-                return;
+                throw new ArgumentNullException("Rebalance handler cannot be null.");
             }
-            partitionsAssignedHandler = partitions => value(this, partitions);
-        }
 
+            if (this.rebalanceHandler != null)
```
edenhill: locking

src/Confluent.Kafka/Consumer.cs
```diff
@@ -695,50 +733,42 @@ protected virtual void Dispose(bool disposing)
...
-        ///     unassigned. This default behavior will not occur if you call Unassign (or Assign)
-        ///     yourself.
-        /// </summary>
-        /// <remarks>
-        ///     Executes as a side-effect of the Consumer.Consume call (on the same thread).
-        /// </remarks>
-        public void SetPartitionsRevokedHandler(Action<IConsumer<TKey, TValue>, List<TopicPartition>> value)
-        {
-            if (value == null)
+            if (this.subscribeHasBeenCalled)
```
edenhill: needs locking

src/Confluent.Kafka/IConsumer.cs
```diff
@@ -70,9 +64,9 @@ public interface IConsumer<TKey, TValue> : IClient
 
 
         /// <summary>
-        ///     Refer to <see cref="Confluent.Kafka.Consumer{TKey,TValue}.Assignment" />
+        ///     Refer to <see cref="Confluent.Kafka.Consumer{TKey,TValue}.AssignedPartitions" />
         /// </summary>
-        List<TopicPartition> Assignment { get; }
+        List<TopicPartition> AssignedPartitions { get; }
```
edenhill: I dont see why the .net client should deviate from the other clients with its own nomenclature.
consistency and familiarity is important

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_RebalanceHandlerAfterSubscribe.cs
```diff
@@ -0,0 +1,56 @@
...
+            var consumerConfig = new ConsumerConfig
+            {
+                GroupId = Guid.NewGuid().ToString(),
+                BootstrapServers = bootstrapServers,
+            };
+
+            using (var consumer = new ConsumerBuilder<byte[], byte[]>(consumerConfig).Build())
+            {
+                consumer.Subscribe(singlePartitionTopic);
+                Assert.Throws<InvalidOperationException>(() => consumer.SetRebalanceHandler((_, e) => {}));
```
edenhill: add a test for multiple set calls too

mhowlett:
> I think we should avoid making up new names for existing things. Java and the other clients already have onPartitionsRevoked, on PartitionsAssigned, let's stick to that form of revoked and assigned.

but this is not an existing thing - with the java client these events correspond to the partitions actually being assigned (as in the consumer will consume from them). with this client, they do not. so my view is very much the opposite - if we use the same names, we will imply they are the same, and this will cause more confusion.

mhowlett:
> There is no 1:1 mapping between subscribe/unsubscribe and subsequent rebalances.
The first rebalance following subscribe/unsubscribe could've been enqueued prior to those calls, triggered by some other event (group rebalance).

good point

examples/Consumer/Program.cs
```diff
@@ -63,25 +63,22 @@ public static void Run_Consume(string brokerList, List<string> topics, Cancellat
...
-                {
-                    Console.WriteLine($"Assigned partitions: [{string.Join(", ", tps)}], member id: {consumer.MemberId}");
-                });
-
-                // Called when the consumer's current assignment set has been revoked.
-                consumer.SetPartitionsRevokedHandler((_, tps) =>
+                // Set the handler to call when the consumer group that this consumer is a
+                // part of rebalances. Except for the initial partition assignment and final 
+                // assignment revocation, a group rebalance will cause this handler to be
+                // called twice. The first time to advise that the current assignment has been
```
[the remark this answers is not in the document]
mhowlett: 👍 great, that's a much better / the correct way of putting it / thinking about it.

examples/Consumer/Program.cs
```diff
@@ -63,25 +63,22 @@ public static void Run_Consume(string brokerList, List<string> topics, Cancellat
...
+                // revoked and the second time to advise of the new assignment. You can
+                // use this callback to perform actions such as retrieving offsets from an
+                // external source and manually setting start offsets using the Assign method.
+                // You can even call Assign with a different set of partitions than those in
+                // the consumer group assignment set. If you do not call Assign, the consumer
```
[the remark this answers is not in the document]
mhowlett: yeah, I'll try to reword this.

src/Confluent.Kafka/IConsumer.cs
mhowlett: Leaving this property as `Assignment` would negate the whole point of this PR - to allow what's going on to be explained without ambiguity. A property called `Assignment` would imply the set of partitions in the consumer group assignment, not the set of partitions assigned to be read from.

mhowlett: On second thoughts, I think  you're right - use of the word `Assignment` here is probably ok. The super important thing to get right is the same word shouldn't be overloaded to mean two different things in the api, and the rebalance handler name change fixes that (by avoiding the need to refer to the consumer group assignment set in the api).

It would be ideal to have different terms for the set of partitions in a consumer group assignment and the set of partitions assigned to - this would make documentation easier / clearer. The only viable thing I could think of previously was to consider assign/assigned and assignment to be two different things. But this is a clutching at straws really and isn't much better than really wordy explanations. I'll change back to `Assignment`, and resign myself to really wordy explanations.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -133,30 +133,30 @@ private void LogCallback(IntPtr rk, SyslogLevel level, string fac, string buf)
...
             if (err == ErrorCode.Local_AssignPartitions)
             {
-                if (partitionsAssignedHandler != null)
+                if (rebalanceHandler != null)
                 {
                     assignCallCount = 0;
-                    partitionsAssignedHandler(partitionList);
+                    rebalanceHandler(new RebalanceEvent(partitionAssignment, true));
                     if (assignCallCount == 1) { return; }
                     if (assignCallCount > 1)
```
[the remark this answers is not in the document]
mhowlett: fail fast - my thinking was it almost certainly indicates incorrect usage (and if not, the restriction could be easily worked around by the programmer).

examples/Consumer/Program.cs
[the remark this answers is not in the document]
mhowlett: I'm really struggling to write this well and propose just removing the comment and adding it back after 1.0 along with other documentation improvements.

examples/Consumer/Program.cs
```diff
@@ -63,25 +63,22 @@ public static void Run_Consume(string brokerList, List<string> topics, Cancellat
...
+                // will be automatically assigned to (will consume from) the partitions of the
+                // consumer group assignment set and consumption will start from last committed
+                // offsets or in accordance with the auto.offset.reset configuration parameter
+                // for partitions where there is no committed offset.
+                consumer.SetRebalanceHandler((_, e) =>
                 {
-                    Console.WriteLine($"Revoked partitions: [{string.Join(", ", tps)}]");
+                    Console.WriteLine($"{(e.IsAssignment ? "Assigned" : "Revoked")} partitions: [{string.Join(", ", e.Partitions)}]");
```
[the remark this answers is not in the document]
mhowlett: ok, i'm indifferent, but in light of the suggestion of taking a short cut with the docs for now, sounds good.

src/Confluent.Kafka/Consumer.cs
mhowlett: i omitted this on purpose and was half expecting a comment :-) - for this to ever come into play, the user would need to have code that is so screwed up that any problem they had would be un-solvable by adding a lock here. but ok, sure.

mhowlett: btw: i'm proactively trying to be less rude in PR comments when there's absolutely no reason to be. failing, apparently.

src/Confluent.Kafka/Consumer.cs
mhowlett: yeah, good point. same sentiment as before (wouldn't like to be reading who-ever's code made this necessary!), but it's a bigger deal here.

src/Confluent.Kafka/Consumer.cs
mhowlett: there's a comment above the if block that covers this (which probably didn't make it into the bounds of the pr diff).

src/Confluent.Kafka/Consumer.cs
mhowlett: me either - I'll remove it.

src/Confluent.Kafka/Consumer.cs
mhowlett: updated the wording. I do think this overload is useful enough to keep (in particular, it's very useful for the single partition topic case). It's used 31 times in the integration tests, and I use it myself in real code in practice.

src/Confluent.Kafka/Consumer.cs
mhowlett: yeah. 

As an aside, I'm still questioning the decision to have some of the set methods on the consumer - finding the tradeoffs tricky to assess. If this changes though, it'll be a separate PR.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_RebalanceHandlerAfterSubscribe.cs
mhowlett: good idea

src/Confluent.Kafka/Consumer.cs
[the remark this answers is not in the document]
edenhill: But it is not an error, so why fail.
This just limits the usability for no win.
And if this is to be enforced it should be enforced in librdkafka, thus for all clients.

src/Confluent.Kafka/Consumer.cs
edenhill: So you're thinking of moving the rebalancehandler to the builder?

mhowlett: I am, but I want to work through it more and defer that to another PR. it also affects the offsets committed handler.

src/Confluent.Kafka/IConsumer.cs
edenhill: Sure, there is the assignment from the group leader, and there is the assigned partitions to consume.
In practice they're almost always the same, and the time window where they are different things is about the first couple of lines in the rebalance handler, so I don't think we need to overly educate our users with the difference.

mhowlett: I disagree, but at least we have an api now that we can both tolerate.

src/Confluent.Kafka/Consumer.cs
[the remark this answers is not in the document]
mhowlett: ok, changed.

mhowlett:
as discussed offline, added a commit to move the `SetRebalanceHandler` and `SetOffsetsCommittedHandler` methods to the consumer builder.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -272,33 +269,7 @@ public void Unsubscribe()
...
-        ///     consume from. If the consumer is already assigned to a set of partitions,
-        ///     the previous set will be replaced.
-        /// </summary>
-        /// <param name="rebalanceEvent">
-        ///     The rebalance event to use to determine which partitions to assign
-        ///     to. If this is an assignment event, the consumer will be assigned to
-        ///     all partitions in the assignment. If this is a revocation event, the
-        ///     consumer will be unassigned from all partitions.
-        /// </param>
-        public void Assign(RebalanceEvent rebalanceEvent)
```
edenhill: Wasn't this variant useful? I thought so :)

mhowlett: yeah, just actioning some good feedback

PR #773, opened 2019-02-11
audit of exception use across library

mhowlett:
ok, ready for review.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ class Program
                 var dr = await p.ProduceAsync("test-topic", new Message<Null, string> { Value="test" });
                 Console.WriteLine($"Delivered '{dr.Value}' to '{dr.TopicPartitionOffset}'");
             }
-            catch (KafkaException e)
+            catch (ProduceException<Null, string> e)
             {
                 Console.WriteLine($"Delivery failed: {e.Error.Reason}");
             }
@@ -177,25 +177,26 @@ class Program
                 cts.Cancel();
             };
 
-            while (!cts.IsCancellationRequested)
+            try
             {
-                try
+                while (true)
                 {
-                    var cr = c.Consume(cts.Token);
-                    Console.WriteLine($"Consumed message '{cr.Value}' at: '{cr.TopicPartitionOffset}'.");
-                }
-                catch (ConsumeException e)
-                {
-                    Console.WriteLine($"Error occured: {e.Error.Reason}");
-                }
-                catch (OperationCanceledException)
-                {
-                    break;
+                    try
+                    {
+                        var cr = c.Consume(cts.Token);
+                        Console.WriteLine($"Consumed message '{cr.Value}' at: '{cr.TopicPartitionOffset}'.");
+                    }
+                    catch (ConsumeException e)
+                    {
+                        Console.WriteLine($"Error occured: {e.Error.Reason}");
+                    }
                 }
             }
-            
-            // Ensure the consumer leaves the group cleanly and final offsets are committed.
-            c.Close();
+            catch (OperationCanceledException)
+            {
+                // Ensure the consumer leaves the group cleanly and final offsets are committed.
+                c.Close();
+            }
         }
     }
 }
@@ -227,11 +228,11 @@ For more information about working with Avro in .NET, refer to the the blog post
 
 Errors delivered to a client's error handler should be considered informational except when the `IsFatal` flag
 is set to `true`, indicating that the client is in an un-recoverable state. Currently, this can only happen on
-the producer, and only when `enable.itempotence` has been set to `true`. In all other scenarios, clients are
+the producer, and only when `enable.idempotence` has been set to `true`. In all other scenarios, clients are
 able to recover from all errors automatically.
 
 Although calling most methods on the clients will result in a fatal error if the client is in an un-recoverable
-state, you should generally only need to explicitly check for fatal errors in your `OnError` handler, and handle
+state, you should generally only need to explicitly check for fatal errors in your error handler, and handle
 this scenario there.
 
 #### Producer
@@ -246,10 +247,8 @@ will be thrown.
 
 #### Consumer
 
-If you are using the deserializing version of the `Consumer`, any error encountered during deserialization (which
-happens during your call to `Consume`) will throw a `DeserializationException`. All other `Consume` errors will
-result in a `ConsumeException` with further information about the error and context available via the `Error` and
-`ConsumeResult` fields.
+All `Consume` errors will result in a `ConsumeException` with further information about the error and context
+available via the `Error` and `ConsumeResult` fields.
 
 
 ### Confluent Cloud
--- a/README.md
+++ b/README.md
@@ -228,8 +228,8 @@ For more information about working with Avro in .NET, refer to the the blog post
 
 Errors delivered to a client's error handler should be considered informational except when the `IsFatal` flag
 is set to `true`, indicating that the client is in an un-recoverable state. Currently, this can only happen on
-the producer, and only when `enable.idempotence` has been set to `true`. In all other scenarios, clients are
-able to recover from all errors automatically.
+the producer, and only when `enable.idempotence` has been set to `true`. In all other scenarios, clients will
+attempt to recover from all errors automatically.
 
 Although calling most methods on the clients will result in a fatal error if the client is in an un-recoverable
 state, you should generally only need to explicitly check for fatal errors in your error handler, and handle
--- a/examples/AvroGeneric/Program.cs
+++ b/examples/AvroGeneric/Program.cs
@@ -68,23 +68,29 @@ namespace Confluent.Kafka.Examples.AvroGeneric
                 {
                     consumer.Subscribe(topicName);
 
-                    while (!cts.Token.IsCancellationRequested)
+                    try
                     {
-                        try
+                        while (true)
                         {
-                            var consumeResult = consumer.Consume(cts.Token);
+                            try
+                            {
+                                var consumeResult = consumer.Consume(cts.Token);
 
-                            Console.WriteLine($"Key: {consumeResult.Message.Key}\nValue: {consumeResult.Value}");
-                        }
-                        catch (ConsumeException e)
-                        {
-                            Console.WriteLine("Consume error: " + e.Error.Reason);
+                                Console.WriteLine($"Key: {consumeResult.Message.Key}\nValue: {consumeResult.Value}");
+                            }
+                            catch (ConsumeException e)
+                            {
+                                Console.WriteLine($"Consume error: {e.Error.Reason}");
+                            }
                         }
                     }
-
-                    consumer.Close();
+                    catch (OperationCanceledException)
+                    {
+                        // commit final offsets and leave the group.
+                        consumer.Close();
+                    }
                 }
-            }, cts.Token);
+            });
 
             using (var schemaRegistry = new CachedSchemaRegistryClient(new SchemaRegistryConfig { SchemaRegistryUrl = schemaRegistryUrl }))
             using (var producer =
--- a/examples/Consumer/Program.cs
+++ b/examples/Consumer/Program.cs
@@ -90,7 +90,7 @@ namespace Confluent.Kafka.Examples.ConsumerExample
                             if (consumeResult.IsPartitionEOF)
                             {
                                 Console.WriteLine(
-                                    $"Reached end of topic {consumeResult.Topic}, partition {consumeResult.Partition}.");
+                                    $"Reached end of topic {consumeResult.Topic}, partition {consumeResult.Partition}, offset {consumeResult.Offset}.");
 
                                 continue;
                             }
--- a/examples/Consumer/Program.cs
+++ b/examples/Consumer/Program.cs
@@ -79,40 +79,53 @@ namespace Confluent.Kafka.Examples.ConsumerExample
             {
                 consumer.Subscribe(topics);
 
-                while (!cancellationToken.IsCancellationRequested)
+                try
                 {
-                    try
+                    while (true)
                     {
-                        var consumeResult = consumer.Consume(cancellationToken);
-
-                        if (consumeResult.IsPartitionEOF)
+                        try
                         {
-                            Console.WriteLine(
-                                $"Reached end of topic {consumeResult.Topic}, partition {consumeResult.Partition}.");
+                            var consumeResult = consumer.Consume(cancellationToken);
 
-                            continue;
+                            if (consumeResult.IsPartitionEOF)
+                            {
+                                Console.WriteLine(
+                                    $"Reached end of topic {consumeResult.Topic}, partition {consumeResult.Partition}.");
+
+                                continue;
+                            }
+
+                            Console.WriteLine($"Received message at {consumeResult.TopicPartitionOffset}: {consumeResult.Value}");
+
+                            if (consumeResult.Offset % commitPeriod == 0)
+                            {
+                                // The Commit method sends a "commit offsets" request to the Kafka
+                                // cluster and synchronously waits for the response. This is very
+                                // slow compared to the rate at which the consumer is capable of
+                                // consuming messages. A high performance application will typically
+                                // commit offsets relatively infrequently and be designed handle
+                                // duplicate messages in the event of failure.
+                                try
+                                {
+                                    consumer.Commit(consumeResult);
+                                }
+                                catch (KafkaException e)
+                                {
+                                    Console.WriteLine($"Commit error: {e.Error.Reason}");
+                                }
+                            }
                         }
-
-                        Console.WriteLine($"Received message at {consumeResult.TopicPartitionOffset}: {consumeResult.Value}");
-
-                        if (consumeResult.Offset % commitPeriod == 0)
+                        catch (ConsumeException e)
                         {
-                            // The Commit method sends a "commit offsets" request to the Kafka
-                            // cluster and synchronously waits for the response. This is very
-                            // slow compared to the rate at which the consumer is capable of
-                            // consuming messages. A high performance application will typically
-                            // commit offsets relatively infrequently and be designed handle
-                            // duplicate messages in the event of failure.
-                            consumer.Commit(consumeResult);
+                            Console.WriteLine($"Consume error: {e.Error.Reason}");
                         }
-                    }
-                    catch (ConsumeException e)
-                    {
-                        Console.WriteLine($"Consume error: {e.Error}");
                     }
                 }
-
-                consumer.Close();
+                catch (OperationCanceledException)
+                {
+                    Console.WriteLine("Closing consumer.");
+                    consumer.Close();
+                }
             }
         }
 
@@ -143,28 +156,34 @@ namespace Confluent.Kafka.Examples.ConsumerExample
             {
                 consumer.Assign(topics.Select(topic => new TopicPartitionOffset(topic, 0, Offset.Beginning)).ToList());
 
-                while (!cancellationToken.IsCancellationRequested)
+                try
                 {
-                    try
+                    while (true)
                     {
-                        var consumeResult = consumer.Consume(cancellationToken);
-                        // Note: End of partition notification has not been enabled, so
-                        // it is guaranteed that the ConsumeResult instance corresponds
-                        // to a Message, and not a PartitionEOF event.
-                        Console.WriteLine($"Received message at {consumeResult.TopicPartitionOffset}: ${consumeResult.Value}");
-                    }
-                    catch (ConsumeException e)
-                    {
-                        Console.WriteLine($"Consume error: {e.Error}");
+                        try
+                        {
+                            var consumeResult = consumer.Consume(cancellationToken);
+                            // Note: End of partition notification has not been enabled, so
+                            // it is guaranteed that the ConsumeResult instance corresponds
+                            // to a Message, and not a PartitionEOF event.
+                            Console.WriteLine($"Received message at {consumeResult.TopicPartitionOffset}: ${consumeResult.Value}");
+                        }
+                        catch (ConsumeException e)
+                        {
+                            Console.WriteLine($"Consume error: {e.Error.Reason}");
+                        }
                     }
                 }
-
-                consumer.Close();
+                catch (OperationCanceledException)
+                {
+                    Console.WriteLine("Closing consumer.");
+                    consumer.Close();
+                }
             }
         }
 
         private static void PrintUsage()
-            => Console.WriteLine("Usage: .. <poll|consume|manual> <broker,broker,..> <topic> [topic..]");
+            => Console.WriteLine("Usage: .. <subscribe|manual> <broker,broker,..> <topic> [topic..]");
 
         public static void Main(string[] args)
         {
@@ -188,7 +207,7 @@ namespace Confluent.Kafka.Examples.ConsumerExample
 
             switch (mode)
             {
-                case "consume":
+                case "subscribe":
                     Run_Consume(brokerList, topics, cts.Token);
                     break;
                 case "manual":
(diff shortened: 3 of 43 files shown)


README.md
```diff
@@ -177,25 +177,26 @@ class Program
                 cts.Cancel();
             };
 
-            while (!cts.IsCancellationRequested)
```
mhowlett: this is pointless, consume will throw.

README.md
```diff
@@ -177,25 +177,26 @@ class Program
...
+                while (true)
                 {
-                    Console.WriteLine($"Error occured: {e.Error.Reason}");
-                }
-                catch (OperationCanceledException)
-                {
-                    break;
+                    try
+                    {
+                        var cr = c.Consume(cts.Token);
```
mhowlett: note: throwing an `OperationCanceledException` when an operation is canceled is idiomatic.

examples/AvroGeneric/Program.cs
```diff
@@ -68,23 +68,29 @@ static async Task Main(string[] args)
...
                     }
-
-                    consumer.Close();
+                    catch (OperationCanceledException)
+                    {
+                        // commit final offsets and leave the group.
+                        consumer.Close();
+                    }
                 }
-            }, cts.Token);
```
mhowlett: it doesn't hurt having this here, but it won't do anything in practice in this case so it's better to remove it or people might think it's important. what this does is tell the task scheduler to check if the token is cancelled before running the task. in this case, it never will be.

src/Confluent.Kafka/AdminClientBuilder.cs
```diff
@@ -77,7 +77,7 @@ public AdminClientBuilder SetStatisticsHandler(Action<AdminClient, string> stati
         {
             if (this.StatisticsHandler != null)
             {
-                throw new ArgumentException("Statistics handler may not be specified more than once.");
+                throw new InvalidOperationException("Statistics handler may not be specified more than once.");
```
mhowlett: this is not an argument exception because the argument is valid. `InvalidOperationException` much more appropriate here (and is a very commonly used exception).

src/Confluent.Kafka/Config.cs
```diff
@@ -225,7 +225,7 @@ protected void SetObject(string name, object val)
                 }
                 if (cancellationDelayMaxMs < 1 || cancellationDelayMaxMs > 10000)
                 {
-                    throw new ArgumentException(
+                    throw new ArgumentOutOfRangeException(
```
mhowlett: derives from argument exception, just getting more specific.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -806,15 +807,16 @@ internal Consumer(ConsumerBuilder<TKey, TValue> builder)
             }
             else
             {
-                throw new ArgumentException("Invalid key deserializer configuration.");
+                // enforced by the builder class.
+                Trace.Fail("Both async and sync key deserializers were set.");
```
mhowlett: `Trace.Fail` is semantically more appropriate for program logic checks (i.e. only occur if we made a mistake). If this ever happened, what would be best from a user experience? that's open for debate i think, i'm not 100% sure. Worth noting the behavior in the case of a `Trace.Fail` is customizable ( refer to https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.trace.fail?view=netframework-4.7.2 ), and it's built for this.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -951,8 +953,8 @@ internal Consumer(ConsumerBuilder<TKey, TValue> builder)
                             },
                             IsPartitionEOF = false
                         },
-                        new Error(ErrorCode.Local_KeyDeserialization, ex.ToString())
-                    );
+                        new Error(ErrorCode.Local_KeyDeserialization),
+                        exception);
```
mhowlett: The inner exception structure here allows the deserializers to throw whatever they like, whilst still providing predictability to users of the `Consume` method (they only need to test for `ConsumeException`).

It's debatable whether deserialization errors should be exposed on something that derives from `KafkaException` which is 1:1 with a librdkafka error otherwise. The main advantage is it means that users of `Consume` can just check for one very specific exception type (`ConsumeException`), otherwise it'd be `Exception` if they just want one catch block. The other advantage is the functionality of the `ConsumeException` is a perfect fit for what we need here (the `ConsumeRecord` field).

mhowlett: Something I'm not 100% sure about: In the future, we could add very specific exception types that are 1:1 with librdkafka error codes. These would derive from `ConsumeException`, `ProduceException` or `KafkaException`. There is a problem here though: if a given error code can be thrown during one or more of consume, produce or outside the context of either - the existence of the produce and consume exception types in the hierarchy mean that we'd need more than on exception type per librdkafka error code. we don't want that. I haven't thought this through at all, but I'm pretty sure this is the case.

so i think we face a choice of having the features the produce and consume exception types provide (basically the consume result / delivery report info) or fine grained exception types. i think the choice is clear - we want the consume and produce exception types, and as I said, i think we're locking in the decision here that we won't ever provide finer grained exception types (at least comprehensively).

practical consequence is that to handle different librdkafka errors differently, people would need to do that with control logic inside a `catch (KafkaException)` block, not using multiple catch statements.

src/Confluent.Kafka/ProduceException.cs
```diff
@@ -31,13 +31,32 @@ public class ProduceException<TKey, TValue> : KafkaException
         /// <param name="error"> 
         ///     The error associated with the delivery report.
         /// </param>
-        /// <param name="deliveryReport">
+        /// <param name="deliveryResult">
```
mhowlett: note: a `DeliveryResult` is a `DeliveryReport` without the error, useful in several contexts.

src/Confluent.Kafka/Producer.cs
```diff
@@ -608,7 +609,7 @@ internal Producer(ProducerBuilder<TKey, TValue> builder)
             var configHandle = SafeConfigHandle.Create();
 
             modifiedConfig.ToList().ForEach((kvp) => {
-                if (kvp.Value == null) throw new ArgumentException($"'{kvp.Key}' configuration parameter must not be null.");
+                if (kvp.Value == null) throw new ArgumentNullException($"'{kvp.Key}' configuration parameter must not be null.");
```
mhowlett: could argue this shouldn't be an argument exception, but i think most appropriate.

src/Confluent.Kafka/Producer.cs
```diff
@@ -665,53 +666,96 @@ internal Producer(ProducerBuilder<TKey, TValue> builder)
             Message<TKey, TValue> message,
             CancellationToken cancellationToken = default(CancellationToken))
         {
-            var keyBytes = (keySerializer != null)
```
mhowlett: there are a bunch of changes to this method which mean that any error encountered due to a call to produce will result in a `ProduceException<K,V>` (note: which has a `DeliveryResult<K,V>` property). Formerly, a `ProduceException` would only occur on receiving a delivery report with an error (in the `Task` based `ProduceAsync` case only). Now, this will also be generated when a produce call returns an error immediately, or due to a serialization problem (refer to discussion in consume case).

This is more convenient. Imagine we have a code block with a bunch of `async` produce calls and some other logic that could also throw. We could wrap it in a `try`/`catch` block, and catch a `ProduceException<K,V>`. We know that if a produce call failed for any reason, we'll find out about it here. We'll also have the context (i.e. topic, key, value etc) regardless of whether the failure was immediate or not. Before this change, we'd need to catch the `KafkaException` type as well + whatever we expect the serializers could throw (+ we'd need to know what that is!) to get every case + in the immediate cases the exception wouldn't include context, which is useful.

Note: we could add a non-generic typed variant of `ProduceException` to allow exceptions from different typed producers to be caught at once. can be added without breaking the API, so left as a todo.

src/Confluent.Kafka/Producer.cs
```diff
@@ -665,53 +666,96 @@ internal Producer(ProducerBuilder<TKey, TValue> builder)
...
+                        enableDeliveryReportKey ? message.Key : default(TKey),
+                        enableDeliveryReportValue ? message.Value : default(TValue));
 
-                ProduceImpl(
-                    topicPartition.Topic,
-                    valBytes, 0, valBytes == null ? 0 : valBytes.Length,
-                    keyBytes, 0, keyBytes == null ? 0 : keyBytes.Length,
-                    message.Timestamp, topicPartition.Partition, message.Headers,
-                    handler);
+                    cancellationToken.Register(() => handler.TrySetException(new TaskCanceledException()));
```
mhowlett: don't worry about this. it doesn't work and it's going away in the future clean-up-the-cancellationtoken-stuff PR.

src/Confluent.Kafka/Producer.cs
```diff
@@ -787,35 +831,75 @@ internal Producer(ProducerBuilder<TKey, TValue> builder)
...
-                topicPartition.Topic,
-                valBytes, 0, valBytes == null ? 0 : valBytes.Length, 
-                keyBytes, 0, keyBytes == null ? 0 : keyBytes.Length, 
-                message.Timestamp, topicPartition.Partition, 
-                message.Headers, 
-                new TypedDeliveryHandlerShim_Action<TKey, TValue>(
+                throw new InvalidOperationException("A delivery handler was specified, but delivery reports are disabled.");
+            }
+
+            byte[] keyBytes;
```
mhowlett: The `BeginProduce` method has also been revamped in this PR to provide uniformity across all types of errors. In this case though, exceptions are never thrown, all errors are reported via the callback. It's not completely without doubt that this is correct (this style of method is not idiomatic c#, so there's not much precedent), but I believe it'll be the most convenient in practice for analogous reasons to the async case.

test/Confluent.Kafka.VerifiableClient/Program.cs
```diff
@@ -207,14 +207,7 @@ public void HandleDelivery(DeliveryReport<byte[], byte[]> record)
...
         {
-            try
-            {
-                Handle.BeginProduce(topic, new Message<byte[], byte[]> { Value = Encoding.UTF8.GetBytes(value) }, record => HandleDelivery(record));
-            }
-            catch (KafkaException e)
-            {
-                Fatal($"Produce({topic}, {value}) failed: {e}");
-            }
+            Handle.BeginProduce(topic, new Message<byte[], byte[]> { Value = Encoding.UTF8.GetBytes(value) }, record => HandleDelivery(record));
```
mhowlett: after this PR, never throws

src/Confluent.Kafka/Admin/AlterConfigsException.cs
```diff
@@ -35,11 +35,11 @@ public class AlterConfigsException : Exception
         ///     results will be in error.
         /// </param>
         public AlterConfigsException(List<AlterConfigsExceptionResult> results)
-            : base(
+            : base(new Error(ErrorCode.Unknown,
```
mhowlett: Currently, 1. a top level error will raise a `KafkaException` and 2. if one of the results is in error this will raise an `AlterConfigsException` which does not derive from `KafkaException`. It would be good if it did since there will be scenarios where handling at the `KafkaException` level would be desired.

However, there's a need for an error code for the scenario 'one of the constituents is in error'. `Unknown` is the best of the available i could see, but it's not really good enough.

If we can't get that in librdkafka pre 1.0, it would be best to revert this now (adding derivation from `KafkaException` later is a non-breaking change).

same comment on other admin client APIs.

src/Confluent.Kafka/Admin/AlterConfigsException.cs
```diff
@@ -34,12 +34,12 @@ public class AlterConfigsException : Exception
         ///     (whether or not they were in error). At least one of these
         ///     results will be in error.
         /// </param>
-        public AlterConfigsException(List<AlterConfigsExceptionResult> results)
-            : base(
+        public AlterConfigsException(List<AlterConfigsReport> results)
```
mhowlett: `AlterConfigsReport` is `AlterConfigsResult`, consistent with `DeliveryReport` vs `DelveryResult`. nice :-). `..ExceptionResult` was an annoyingly horrible name.

src/Confluent.Kafka/Exceptions/TopicPartitionException.cs
```diff
@@ -24,7 +24,7 @@ namespace Confluent.Kafka
     /// <summary>
     ///     Represents an error that occured during a Consumer.Position request.
     /// </summary>
-    public class TopicPartitionException : Exception
+    public class TopicPartitionException : KafkaException
```
mhowlett: two things:

1. like the admin client exceptions, it's best if this derives from `KafkaException` but we need a new `ErrorCode` for it, unknown isn't good enough.

2. left as todo: we should have exception types that drive from this e.g. `ConsumerPauseException`.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -977,7 +978,7 @@ internal static IntPtr GetCTopicPartitionList(IEnumerable<TopicPartitionOffset>
             IntPtr list = Librdkafka.topic_partition_list_new((IntPtr)offsets.Count());
             if (list == IntPtr.Zero)
             {
-                throw new OutOfMemoryException("Failed to create topic partition list");
+                throw new Exception("Failed to create topic partition list");
```
mhowlett: the manual says that only the runtime should raise OOM exceptions.

src/ConfigGen/Program.cs
```diff
@@ -558,10 +558,10 @@ static List<PropertySpecification> extractAll(string configDoc)
 
                 var desc = columns[5].Trim();
                 bool isAlias = desc.StartsWith("Alias");
-                if (isAlias != !desc.Contains("<br>*Type")) { throw new Exception("Inconsistent indication of alias parameter"); }
                 if (isAlias)
                 {
-                    prop.AliasFor = desc.Substring(desc.IndexOf('`')+1, desc.LastIndexOf('`') - desc.IndexOf('`') - 1);
+                    var firstIdx = desc.IndexOf('`')+1;
```
mhowlett: also updated ConfigGen to be compatible with the RC7 config.

README.md
```diff
@@ -100,7 +100,7 @@ class Program
                 var dr = await p.ProduceAsync("test-topic", new Message<Null, string> { Value="test" });
                 Console.WriteLine($"Delivered '{dr.Value}' to '{dr.TopicPartitionOffset}'");
             }
-            catch (KafkaException e)
+            catch (ProduceException<Null, string> e)
```
mhowlett: It's a pity this is more verbose. I'm tempted to just leave this as `KafkaException` since that's all that's needed by the log message, but it's instructional to use `ProduceException` since something this specific is always expected.

src/Confluent.Kafka/Producer.cs
mhowlett: exception to the above: programmer error type exceptions (eg argument exception) are better if not coerced into a ProduceException.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -806,16 +807,17 @@ internal Consumer(ConsumerBuilder<TKey, TValue> builder)
             }
             else
             {
-                throw new ArgumentException("Invalid key deserializer configuration.");
+                // enforced by the builder class.
+                throw new Exception("FATAL: Both async and sync key deserializers were set.");
```
mhowlett: I was initially using `Trace.Fail` for programmer errors (i.e. things that would only occur if we made a mistake), but that was causing problems on .netframework45.

src/Confluent.Kafka/Consumer.cs
mhowlett: note: the netframework case was causing test failures, so i reverted to Exception for this type of error.

src/Confluent.Kafka/Admin/CreatePartitionsException.cs
```diff
@@ -34,12 +34,12 @@ public class CreatePartitionsException : Exception
         ///     (whether or not they were in error). At least one of these
         ///     results will be in error.
         /// </param>
-        public CreatePartitionsException(List<CreatePartitionsExceptionResult> results)
-            : base(
+        public CreatePartitionsException(List<CreatePartitionsReport> results)
+            : base(new Error(ErrorCode.Unknown,
```
mhowlett: same comments as `AlterConfigs` case re: deriving from `KafkaException` and the `CreatePartitionsReport` class name change.

README.md
edenhill: I'm not thrilled with having the fully typed producer here.
Is there any way an untyped ProduceException can be used instead?

README.md
```diff
@@ -227,11 +228,11 @@ For more information about working with Avro in .NET, refer to the the blog post
 
 Errors delivered to a client's error handler should be considered informational except when the `IsFatal` flag
 is set to `true`, indicating that the client is in an un-recoverable state. Currently, this can only happen on
-the producer, and only when `enable.itempotence` has been set to `true`. In all other scenarios, clients are
+the producer, and only when `enable.idempotence` has been set to `true`. In all other scenarios, clients are
```
edenhill: "clients are able" -> "clients will attempt"

examples/Consumer/Program.cs
```diff
@@ -79,40 +79,53 @@ public static void Run_Consume(string brokerList, List<string> topics, Cancellat
...
-                            Console.WriteLine(
-                                $"Reached end of topic {consumeResult.Topic}, partition {consumeResult.Partition}.");
-
-                            continue;
+                            var consumeResult = consumer.Consume(cancellationToken);
+
+                            if (consumeResult.IsPartitionEOF)
+                            {
+                                Console.WriteLine(
+                                    $"Reached end of topic {consumeResult.Topic}, partition {consumeResult.Partition}.");
```
edenhill: print the offset too.
there is a consumeResult.TopicPartitionOffset for EOFs, right?

examples/Consumer/Program.cs
```diff
@@ -79,40 +79,53 @@ public static void Run_Consume(string brokerList, List<string> topics, Cancellat
...
+                            }
+
+                            Console.WriteLine($"Received message at {consumeResult.TopicPartitionOffset}: {consumeResult.Value}");
+
+                            if (consumeResult.Offset % commitPeriod == 0)
+                            {
+                                // The Commit method sends a "commit offsets" request to the Kafka
+                                // cluster and synchronously waits for the response. This is very
+                                // slow compared to the rate at which the consumer is capable of
+                                // consuming messages. A high performance application will typically
```
edenhill: Worth pointing out: or preferably automatically by enable.auto.commit

src/Confluent.Kafka/Admin/AlterConfigsException.cs
edenhill: From a user's perspective, do I really care if the entire request failed or if one or more (possibly all) resources failed?
It is sort of pointless in first checking request-level errors and then entity-level errors, when the entity-level errors will reflect the request-level error.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -447,11 +439,39 @@ public List<TopicPartitionOffset> Commit(CancellationToken cancellationToken = d
...
+        ///     The ConsumeResult instance used to determine the committed offset.
+        /// </param>
+        /// <exception cref="Confluent.Kafka.KafkaException">
+        ///     Thrown if the request failed.
+        /// </exception>
+        /// <exception cref="Confluent.Kafka.TopicPartitionOffsetException">
+        ///     Thrown if the result is in error.
+        /// </exception>
+        /// <remarks>
+        ///     A consumer which has position N has consumed messages with offsets up to N-1
```
edenhill: A consumer at position N ..

src/Confluent.Kafka/Consumer.cs
edenhill: Why use the base Exception here and not InvalidOperation?

src/Confluent.Kafka/Consumer.cs
```diff
@@ -829,7 +836,8 @@ internal Consumer(ConsumerBuilder<TKey, TValue> builder)
             }
             else
             {
-                throw new ArgumentException("Invalid value deserializer configuration.");
+                // enforced by the builder class.
+                throw new Exception("FATAL: Both async and sync value deserializers were set.");
```
edenhill: dito

src/Confluent.Kafka/Consumer.cs
edenhill: I too believe detailed librdkafka 1:1 exceptions are the most user friendly.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -340,7 +340,7 @@ private IntPtr marshalHeaders(IEnumerable<Header> headers)
             {
                 if (valOffset != 0 || valLength != 0)
                 {
-                    throw new ArgumentException("valOffset and valLength parameters must be 0 when producing null values.");
+                    throw new InvalidOperationException("valOffset and valLength parameters must be 0 when producing null values.");
```
edenhill: this sounds like an argumentException to me

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -353,7 +353,7 @@ private IntPtr marshalHeaders(IEnumerable<Header> headers)
             {
                 if (keyOffset != 0 || keyLength != 0)
                 {
-                    throw new ArgumentException("keyOffset and keyLength parameters must be 0 when producing null key values.");
+                    throw new InvalidOperationException("keyOffset and keyLength parameters must be 0 when producing null key values.");
```
edenhill: dito

src/Confluent.Kafka/Producer.cs
```diff
@@ -488,7 +488,7 @@ public Handle Handle
             }
             else
             {
-                throw new ArgumentException("Invalid key serializer configuration.");
+                throw new Exception("FATAL: Both async and sync key serializers were set.");
```
edenhill: Sounds like an Argument or InvalidOperation exception

src/Confluent.Kafka/Producer.cs
edenhill: > Note: we could add a non-generic typed variant of `ProduceException` to allow exceptions from different typed producers to be caught at once. can be added without breaking the API, so left as a todo.

That answers my previous concern :+1:

src/Confluent.Kafka/Producer.cs
edenhill: I don't think QueueFull should be raised through the delivery report since it voids any way for the application to maintain produce ordering.

src/Confluent.Kafka/Admin/AlterConfigsException.cs
edenhill: Why is the error code set to Unknown?

mhowlett: most of the time you may not care, which is why you want to be able to just catch `KafkaException` (and have it work for all cases) - this is why the `Unknown` error code (which should be something else not unknown), but sometimes you do care if some of the results are in error and others aren't in which case you want to additionally catch the more specific exception.

src/Confluent.Kafka/Producer.cs
mhowlett: exceptionally good point

src/Confluent.Kafka/Consumer.cs
mhowlett: as discussed above, the use of `ConsumeException` and `ProduceException` may make this problematic.

src/Confluent.Kafka/Producer.cs
mhowlett: well you can already do this by catching `KafkaException` - only need `ProduceException<K,V>` if you need access to the info it provides. `ProduceException` is kind of halfway in between (doesn't provide K,V).

README.md
mhowlett: there's always the base class `KafkaException`.

src/Confluent.Kafka/Consumer.cs
mhowlett: that's better yes, changing.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
mhowlett: i was on the fence. will revert.

src/Confluent.Kafka/Producer.cs
mhowlett: I've changed this to throw `ProduceException<K,V>`'s which are `KafkaException`s with extra information. Also, did the same with k/v exceptions, since now they aren't the odd ones out.

PR #774, opened 2019-02-11
fixed error with schema registry config in .net 461

mkellogg:
The assertion made in the comment:

"// Convert.ToString returns "" in the null case here." 

Is incorrect for framework 4.6.1. Added a null coalescing operator to resolve the problem.

This appears to be the only place in the project that uses Convert.ToString.


mkellogg:
Not sure if I can do anything about those failed checks (at least the appveryor one, the travis one seems to be in much worse shape beyond whats in this PR).

It's trying to restore the 4.6.1 versions of xunit and failing. The unit test I created is more a of a regression, if you care. It is just making sure that a bare bones instance of CachedSchemaRegistryClient can be created while running under 4.6.1.

mhowlett:
thanks @mkellogg ! - change lgtm though it seems really weird that there is an incompatibility here - shouldn't be.

if you can move that test to the existing test project and remove the net461 specific project, i'll merge. yes, we should expand automated test coverage to the netframework case, but this can happen more broadly and in a different PR for that purpose.

mkellogg:
A test like this already exists in the non net461 test project I created. 

Tbh, not sure if there's a framework bug here, and if there is, not sure which side is right :)

I'll just remove the test project completely and resubmit and you'll just have to trust me that it's fixed.

mhowlett:
thanks! fyi: the project is still referenced in the sln file - also need to remove that.

mkellogg:
> thanks! fyi: the project is still referenced in the sln file - also need to remove that.

Done. Sorry about that.

mhowlett:
lgtm!

nvladimiroff:
fyi I think this is a bug that affects all clients that don't use auth, not just those on 4.6.1. The problem here looks to be that
```
var basicAuthSource = Convert.ToString(config.FirstOrDefault(prop => prop.Key.ToLower() == SchemaRegistryConfig.PropertyNames.SchemaRegistryBasicAuthCredentialsSource).Value);
```
uses `Convert.ToString(string)`, not `Convert.ToString(object)`, and the string overload will [return null if given a null](https://stackoverflow.com/questions/13514337/convert-tostring-behaves-differently-for-null-object-and-null-string?lq=1).

A workaround we found that worked for us is to directly assign the property:
```
schemaRegistryConfig.Set("schema.registry.basic.auth.credentials.source", "");
```
to get around the property setter on the config object (`SchemaRegistryBasicAuthCredentialsSource`) being an enum.

mkellogg:
I never experienced the problem in my .net core 2.0 test app, but other teams running my wrapper library in 4.6.1 did. Regardless, the fix in the PR should take care of everyone.

PR #779, opened 2019-02-13
partital error exception types

mhowlett:
it's useful for these exception types to have a base class `KafkaException` since the user may not care about the details of the problem, and this allows these exceptions to be caught in the same handler as other librdkafka generated exceptions.

the `Local_` prefix is a bit misleading, but i think this is ok.

this pr is reverting the revert in #773 (but uses an appropriate error code).

### DIFF
--- a/src/Confluent.Kafka/Admin/AlterConfigsException.cs
+++ b/src/Confluent.Kafka/Admin/AlterConfigsException.cs
@@ -24,7 +24,7 @@ namespace Confluent.Kafka.Admin
     /// <summary>
     ///     Represents an error that occured during an alter configs request.
     /// </summary>
-    public class AlterConfigsException : Exception
+    public class AlterConfigsException : KafkaException
     {
         /// <summary>
         ///     Initializes a new instance of AlterConfigsException.
@@ -35,11 +35,11 @@ namespace Confluent.Kafka.Admin
         ///     results will be in error.
         /// </param>
         public AlterConfigsException(List<AlterConfigsReport> results)
-            : base(
+            : base(new Error(ErrorCode.Local_Partial,
                 "An error occurred altering the following resources: [" +
                 String.Join(", ", results.Where(r => r.Error.IsError).Select(r => r.ConfigResource)) +
                 "]: [" + String.Join(", ", results.Where(r => r.Error.IsError).Select(r => r.Error)) + 
-                "].")
+                "]."))
         {
             Results = results;
         }
--- a/src/Confluent.Kafka/Admin/CreatePartitionsException.cs
+++ b/src/Confluent.Kafka/Admin/CreatePartitionsException.cs
@@ -24,7 +24,7 @@ namespace Confluent.Kafka.Admin
     /// <summary>
     ///     Represents an error that occured during a create partitions request.
     /// </summary>
-    public class CreatePartitionsException : Exception
+    public class CreatePartitionsException : KafkaException
     {
         /// <summary>
         ///     Initialize a new instance of CreatePartitionsException.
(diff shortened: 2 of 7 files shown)


src/Confluent.Kafka/Admin/AlterConfigsException.cs
```diff
@@ -24,7 +24,7 @@ namespace Confluent.Kafka.Admin
     /// <summary>
     ///     Represents an error that occured during an alter configs request.
     /// </summary>
-    public class AlterConfigsException : Exception
+    public class AlterConfigsException : KafkaException
```
edenhill: I'm still not convinced that an aggregating exception is the best choice for the Admin APIs.
If the entire request fails a top-level aggregate exception makes sense, but if the request succeeds and 1 or more entities fail the user will want to check each entity's status explicitly, otherwise she'll miss the successful operations (i.e., a successful topic creation).

And from a user perspective I don't really want two levels of error handling, first check the request-level and then per-entity. Only the latter should be needed.

Python example:
```
    new_topics = [NewTopic(topic, num_partitions=3, replication_factor=1) for topic in topics]
    # Call create_topics to asynchronously create topics, a dict
    # of <topic,future> is returned.
    fs = a.create_topics(new_topics)

    # Wait for operation to finish.
    # Timeouts are preferably controlled by passing request_timeout=15.0
    # to the create_topics() call.
    # All futures will finish at the same time. (WHICH MIGHT CHANGE IN FUTURE VERSIONS, stupid comment)
    for topic, f in fs.items():
        try:
            f.result()  # The result itself is None
            print("Topic {} created".format(topic))
        except Exception as e:
            print("Failed to create topic {}: {}".format(topic, e))
```
https://github.com/confluentinc/confluent-kafka-python/blob/master/examples/adminapi.py#L43

mhowlett: the way i have it is better for a number of reasons:

1. If individual `Task`s (which can have the exception set on them individually) are returned in a collection, this cannot be `await`ed. That's a very big downside.
2. It's structurally less complex.
3. The way you have it is equally verbose as the way I have it - i see no advantage there.
4. The we you have it forces error handling inline. Maybe you want this in most practical scenarios, but it's not idiomatic C# to force this (and I don't think you do always want this - i'd write a quick little app where I really don't care if the app crashes if the result fails. the way i have it is much better for that sort of thing).
5. If we have a situation in the future where partial requests can complete at different times, the API can be extended to that then, similar to how you suggest (see comments in code). But generally, people would not use that in preference to the current implementation if they don't need to, because of all the downsides.

edenhill: Can you show a .NET example doing the same thing as the python code above?

mhowlett:
```
try
{
    adminClient.CreateTopicsAsync(new TopicSpecification[] { new TopicSpecification { Name = "topic", NumPartitions = 2, ReplicationFactor = 1 } });
}
catch (CreateTopicsException e)
{
    e.Results.ForEach(r => Console.WriteLine(r.Error.IsError ? "had error" : r.Error.Reason));
}
```

mhowlett: notes: the above is substantially less verbose than doing something analogous to what you've done in python in C#. what you've done is more powerful - but i'm arguing it's not necessary (and can be added later without destroying the api if really needed, and most people are going to chose to use the simpler API even if the more powerful one is added). i expect this argument would also apply in python - the above approach would be simpler to write because it's simpler.

edenhill: A problem is that it just returns one single Task to be awaited, while future version of the Admin API might reach out to different brokers and thus finish at different times.

mhowlett: >5. If we have a situation in the future where partial requests can complete at different times, the API can be extended to that then, similar to how you suggest (see comments in code). But generally, people would not use that in preference to the current implementation if they don't need to, because of all the downsides.

mhowlett: there's no point having the complex option when it's not yet required, may not ever be needed, and even when it becomes useful, people are probably often going to prefer the simpler one, so if we do get to the point where it's needed we'll want both anyway.

edenhill: Okay, sounds good

mhowlett:
:-)

mhowlett:
thanks for the discussion

PR #788, opened 2019-02-19
Mono support

4garbage:
Support Mono on linux/macos.. maybe usefull if netcore is unacceptable (wcf)
native kafka libraries (librdkafka) should be placed in bin folder manually according platform version


mhowlett:
thanks for the contribution @4garbage - inclusion of this is on the radar for after 1.0 - we also want to do some other changes related to library loading and this can be included/tested at the same time. Note that we won't back-port to 0.11.x.

mhowlett:
continuing in #1443

PR #791, opened 2019-02-20
move get/query watermark offset methods to consumer

mhowlett:
Follows the java client and puts the query/get watermark functionality on the consumer where it is commonly needed rather than requiring a wrapper admin client. this change takes the query wm offsets capability away from being available on a librdkafka producer instance (via being wrapped by an admin client), but I don't think that's a common requirement. Since I can't think when this would be useful and the java producer doesn't provide similar functionality, I took the view that `QueryWatermarkOffsets` could be added back to `AdminClient` or to the `Producer` later if the need becomes clear (non-breaking change). For `GetWatermarkOffsets`, it was an especially questionable decision to effectively allow this to be called on a producer instance, since it favors runtime over compile enforcement.

It would be good to have `endOffsets` and `beginningOffsets` method functionality as per the java consumer, but that's not easy right yet?

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -604,6 +604,48 @@ namespace Confluent.Kafka
 
 
         /// <summary>
+        ///     Get the last cached low (oldest available/beginning) and high (newest/end)
+        ///     offsets for a topic/partition.
+        /// </summary>
+        /// <remarks>
+        ///     The low offset is updated periodically (if statistics.interval.ms 
+        ///     is set) while the high offset is updated on each fetched message set from
+        ///     the broker. If there is no cached offset (either low or high, or both) then
+        ///     Offset.Invalid will be returned for the respective offset.
+        /// </remarks>
+        /// <param name="topicPartition">
+        ///     The topic/partition of interest.
+        /// </param>
+        /// <returns>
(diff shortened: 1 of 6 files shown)


mhowlett:
additional note: the `ListGroups` method overloads remain on `AdminClient`. As the group protocol is general, and these methods expose binary data only, this makes sense. However, in the future, we'll probably want to parse the binary payload for the CG specific case, in which case that would make sense on the consumer. We may also not bother providing an api for the general case. Regardless, these methods are marked as unstable, so it's not important to make any decisions now.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -600,7 +600,44 @@ public List<TopicPartitionOffset> Position(IEnumerable<TopicPartition> partition
...
+        /// <returns>
+        ///     The requested WatermarkOffsets (see that class for additional documentation).
+        /// </returns>
+        public WatermarkOffsets GetWatermarkOffsets(TopicPartition topicPartition)
+            => kafkaHandle.GetWatermarkOffsets(topicPartition.Topic, topicPartition.Partition);
+
+
+        /// <summary>
+        ///     Query the Kafka cluster for low (oldest available/beginning) and high (newest/end)
+        ///     offsets for the specified topic/partition (blocking).
```
edenhill: The blocking thing warrants its own line of explanation.

PR #792, opened 2019-02-21
rebalance alternative II  [for discussion only, not for merge]

mhowlett:
if the consumer group assign method is restricted to only being able to accept the set of partitions in the current active consumer group assignment set (if there is one), similar to how the java consumer works, then there is no nomenclature problem.

The argument for this alternative is:

1. the requirement for consumer group assignment / assigned partitions to be different is very infrequently needed.
2. the primary use-case this enables can be achieved with another consumer instance. As far as I can tell, I don't think this has any real practical downside.

i.e. this is a case of simple things being simple (api is straightforward for 99% of users, nothing to get confused over) with complex things being possible.

note: I believe this PR is very close to being good to merge, but as I expect it will be rejected, I didn't do too many iterations over it. If we are going with this option, i'll probably need a 'minor tweak' commit.


mhowlett:
a modification to this PR which would allow for 'control partitions' or 'global table partitions' (streams use case) without requiring an additional consumer instance would be add another parameter to the subscribe method: 

`Consumer.Subscribe(List<string> topics, IEnumerable<TopicPartition> auxilaryPartitions = null)`

this could be done now or as a non-breaking change at a later time.

mhowlett:
with the above modification, I'm preferring this over #789. Credit to @norwood for opening up this line of thinking by suggesting another alternative, which would be to have an additional method, something like "SubscribeGlobal" (but not that name), for the fixed partitions (I like the additional parameter better though).

mhowlett:
This is what usage would look like:

```
using (var consumer = new ConsumerBuilder<Ignore, string>(config)
    // The SetPartitionsAssignedHandler method could be overloaded - two variants:
    // the simple one that has 2 arguments, and the one that takes 3 arguments.
    // This seems more convenient than using the 2 argument variant in both cases.
    .SetPartitionsAssignedHandler((c, ps, aps) =>
    {
        // (optionally) get offsets from external store.
        var externalOffsets = new Dictionary<TopicPartition, Offset>();

        ps.Select(p => new TopicPartitionOffset(p, externalOffsets[p]))
            .Concat(aps.Select(p => new TopicPartitionOffset(p, 0)));

        // blows up if ps TopicPartition collection doesn't match expected.
        c.Assign(ps);
    })
    .Build())
{
    // note: second argument is optional.
    consumer.Subscribe(
        topics,
        new List<TopicPartition> { new TopicPartition("control-topic", 0) });
    ...
}
```

mhowlett:
having written that out, I'm less convinced the structure is ideal.

also, using `Assign` in the `SetAssignedPartitionsHandler` means the naming is muddled up still (in java this seek would be used for this).

so this option is feeling like a dead end to me.

PR #801, opened 2019-02-27
 remove ./bootstrap-topics.sh, use AdminClient to create test topics

mhowlett:
This PR was initially opened to allow integration tests to run on Windows.
It's now been extended to include lots of additional improvements to the integration tests.


test/Confluent.Kafka.IntegrationTests/TemporaryTopic.cs
```diff
@@ -23,19 +23,24 @@ namespace Confluent.Kafka
 {
     public class TemporaryTopic : IDisposable
     {
-        IAdminClient adminClient;
+        private string bootstrapServers;
```
mhowlett: changes to this class are probably not necessary. I was thinking it would be better (more focussed test) if the adminclient was'd open over the duration of the lifetime of the temporary topic. but the improvement is marginal really. anyway, since i made the change, I left it in.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Exiting.cs
```diff
@@ -82,8 +82,8 @@ public static void Consumer_Exiting(string bootstrapServers, string singlePartit
                     switch (i)
                     {
                         case 0:
-                            LogToFile("  -- Unsubscribe [BROKEN!]");
-                            // consumer.Unsubscribe();
+                            LogToFile("  -- Unsubscribe");
```
mhowlett: now works!

test/Confluent.Kafka.IntegrationTests/Tests/Producer_ProduceAsync_Await.cs
```diff
@@ -44,7 +44,7 @@ public static void Producer_ProduceAsync_Await_Serializing(string bootstrapServe
                         singlePartitionTopic,
                         new Message<Null, string> { Value = "test string" });
                     Assert.Equal(0, producer.Flush(TimeSpan.FromSeconds(10)));
-                    Assert.True(dr.Offset > 0);
+                    Assert.True(dr.Offset != Offset.Invalid);
```
mhowlett: bonus! discovered an error in this test. the offset could be 0.

test/Confluent.Kafka.IntegrationTests/Tests/Tests.cs
```diff
@@ -14,21 +14,75 @@
...
+using Xunit;
 
 namespace Confluent.Kafka.IntegrationTests
 {
-    public static partial class Tests
+    public class GlobalFixture : IDisposable
+    {
+        private string bootstrapServers;
+
+        public GlobalFixture()
```
mhowlett: this `GlobalFixture` nonsense is required to implement lifecycle methods before and after a test set in Xunit (as opposed to before and after every test, which is easy to do).

test/Confluent.Kafka.IntegrationTests/Tests/AdminClient_AlterConfigs.cs
```diff
@@ -28,13 +28,13 @@
 
 namespace Confluent.Kafka.IntegrationTests
 {
-    public static partial class Tests
+    public partial class Tests
```
mhowlett: the move away from `static` classes allows for per-test lifecycle methods.

test/Confluent.Kafka.IntegrationTests/TemporaryTopic.cs
```diff
@@ -23,19 +23,24 @@ namespace Confluent.Kafka
...
+        
+        public string Name { get; set; }
 
         public TemporaryTopic(string bootstrapServers, int numPartitions)
```
edenhill: I think it might be better name prefix all these topics with a known name, librdkafka has rdktest_.... or something, which allows a tool to scan and delete these topics at any point.
The problem with the current approach is that temporary topics will not be cleaned up if the tests crash.

The way the librdkafka integration tests does this is use the prefix, and then after all tests have finished it scans through and deletes the topics (optional).

test/Confluent.Kafka.IntegrationTests/Tests/AddBroker.cs
```diff
@@ -17,96 +17,55 @@
...
+        ///     https://github.com/edenhill/librdkafka/wiki/Broker-version-compatibility
+        ///     A metadata request is used to check if there is a broker connection.
+        /// </remarks>
         [Theory, MemberData(nameof(KafkaParameters))]
-        public static void AddBrokers(string bootstrapServers, string singlePartitionTopic, string partitionedTopic)
+        public void AddBrokers(string bootstrapServers)
         {
-            // This test assumes broker v0.10.0 or higher:
-            // https://github.com/edenhill/librdkafka/wiki/Broker-version-compatibility
+            var producerConfig = new ProducerConfig { BootstrapServers = "unknown" };
```
edenhill: Albeit unlikely, unknown might resolve to something for other developer's environments, or simply stall while trying to resolve.
Suggest you use "localhost:65533" or something that is unlikely to be bound, if you just want a broker that will not respond.

test/Confluent.Kafka.IntegrationTests/Tests/AdminClient_AlterConfigs.cs
edenhill: I strongly suggest you add a mode for running multiple tests in parallel, which is not only a great way to speed up test runs, but also verifies that there are no concurrency issues between client instances.
The librdkafka integration tests will run (by default) 5 simultaneous tests.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_OffsetsForTimes.cs
```diff
@@ -56,7 +54,7 @@ public static void Consumer_OffsetsForTimes(string bootstrapServers, string sing
                 // If empty request, expect empty result.
                 var result = consumer.OffsetsForTimes(new TopicPartitionTimestamp[0], timeout).ToList();
                 Assert.Empty(result);
-
+
```
edenhill: This wont work

test/Confluent.Kafka.IntegrationTests/Tests/Tests.cs
edenhill: Suggest you add that as a comment, because there are no clues in the code.

test/Confluent.Kafka.IntegrationTests/Tests/Tests.cs
```diff
@@ -14,21 +14,73 @@
...
+        {
+            var assemblyPath = typeof(Tests).GetTypeInfo().Assembly.Location;
+            var assemblyDirectory = Path.GetDirectoryName(assemblyPath);
+            var jsonPath = Path.Combine(assemblyDirectory, "testconf.json");
+            var json = JObject.Parse(File.ReadAllText(jsonPath));
+            bootstrapServers = json["bootstrapServers"].ToString();
+
+            SinglePartitionTopic = Guid.NewGuid().ToString();
```
edenhill: Should have a fixed prefix

test/Confluent.Kafka.IntegrationTests/Tests/Tests.cs
```diff
@@ -14,21 +14,73 @@
...
+    {
+        private string bootstrapServers;
+
+        public GlobalFixture()
+        {
+            var assemblyPath = typeof(Tests).GetTypeInfo().Assembly.Location;
+            var assemblyDirectory = Path.GetDirectoryName(assemblyPath);
+            var jsonPath = Path.Combine(assemblyDirectory, "testconf.json");
+            var json = JObject.Parse(File.ReadAllText(jsonPath));
+            bootstrapServers = json["bootstrapServers"].ToString();
```
edenhill: Is there an automatic Json-to-type reader in .NET that you could use to deserialize JSON into a Configuration object right away?

test/Confluent.Kafka.IntegrationTests/testconf.json
```diff
@@ -1,5 +1,3 @@
 {
-  "bootstrapServers": "localhost:9092",
-  "singlePartitionTopic": "test-topic-1",
-  "partitionedTopic": "test-topic-2"
+  "bootstrapServers": "localhost:9092"
```
edenhill: I think we should avoid having default brokers for anything, if this happens to be run on the wrong machine it could mess with a production cluster.

test/Confluent.Kafka.IntegrationTests/TemporaryTopic.cs
mhowlett: yep, I can do that, makes it generally more robust. that said, my integration test clusters are always temporary, and the docker scripts i'm setting up to auto-run integration tests atm will also bring up new clusters for a given test run, so it doesn't really matter.

test/Confluent.Kafka.IntegrationTests/Tests/AddBroker.cs
mhowlett: good point

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_OffsetsForTimes.cs
mhowlett: yes it will, i explicitly followed the java api here:

https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs#L519

test/Confluent.Kafka.IntegrationTests/Tests/Tests.cs
mhowlett: don't know - this is only one line, so i don't thinks warrants thinking about.

test/Confluent.Kafka.IntegrationTests/testconf.json
mhowlett: technically, `testconf.json` shouldn't be checked in yes, I just do it as a convenience. this isn't a regression and it's all about to be replaced with some dockerizing i'm doing at the moment so I don't think needs changing here and now.

test/Confluent.Kafka.IntegrationTests/Tests/Tests.cs
mhowlett: ok

mhowlett:
re: parallel running of tests - good point. to allow that, we'd need to change all the tests to use the `TemporaryTopic` (obviously, what I guess you're getting at), and I haven't done that. Future enhancement. xUnit can easily run test sets in parallel.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_OffsetsForTimes.cs
mhowlett: it is arguable which is better - on the one hand, returning an empty list for this case means code that can take an arbitrarily long list of partitions doesn't need to handle a special case. on the other hand, and empty list may indicate an error. however, other code is going to pick up on that in practice. my view is the former is better, and that java works this way enforces that viewpoint.

mhowlett: (but my opinion isn't strong).

test/Confluent.Kafka.IntegrationTests/TemporaryTopic.cs
```diff
@@ -30,7 +30,7 @@ public class TemporaryTopic : IDisposable
...
         {
             this.bootstrapServers = bootstrapServers;
-            this.Name = Guid.NewGuid().ToString();
+            this.Name = "dotnet_test_" + Guid.NewGuid().ToString();
```
edenhill: nit: I'd define this prefix once.

PR #817, opened 2019-03-10
rebalance API mk IV [incomplete, seeking high level feedback on approach]

mhowlett:
currently my favourite option by far.

for discussion, refer to:
https://github.com/confluentinc/confluent-kafka-dotnet/pull/789#discussion_r264025539
https://github.com/confluentinc/confluent-kafka-dotnet/pull/789#discussion_r264047950

this PR is not complete. The implementation is pretty much in place but need to update docs, tests and examples.

opening early for feedback on the general approach.


mhowlett:
to extend the api to support subscribing to a fixed set of partitions in addition to those specified by the group coordinator, you'd just add an overload of the `Subscribe` method and specify them there:

`Consumer<K,V>.Subscribe(topic, IEnumerable<TopicPartition> additional)`

i was also thinking convenience overloads of the `SetPartitionsAssignedHandler` and `SetPartitionsRevokedHandler` could also be added that separate the types of assigned partitions, but i think that'd be a net negative (ordinary users would see them and wonder if they needed them).

mhowlett:
to extend the api to support modifying the set of partitions specified by the group coordinator, you'd just add an overload of the Subscribe method which takes a function that transforms the coordinator assignment to the actual partitions to assign (typically adding a specific partition):

`Consumer<K,V>.Subscribe(IEnumerable<string> topics, Func<IEnumerable<TopicPartition>,IEnumerable<TopicPartition>> transformPartitions)`

this is simple enough to be in scope for this PR.

examples/Consumer/Program.cs
```diff
@@ -61,19 +61,18 @@ public static void Run_Consume(string brokerList, List<string> topics, Cancellat
...
-                    {
-                        Console.WriteLine($"Revoked partitions: [{string.Join(", ", e.Partitions)}]");
-                        // consumer.Unassign()
-                    }
+                    Console.WriteLine($"Partition assignment: [{string.Join(", ", partitions)}]");
+                    // By default, consumption will resume from the last committed offset for each
+                    // partition, or if there is no committed offset, in accordance with the 
+                    // `auto.offset.reset` configuration property. You can override this behavior
+                    // by manually seeking to the desired start offsets in this handler:
+                    // c.Seek(...)
```
edenhill: I think faking/delaying the Seek() here will be problematic:  any other calls after the Seek() will not function as expected, no debugging will be emitted, and there are probably other corner cases as well.

Not sure we should really trick the user, I believe that will come back and bite us.

edenhill: An alternative would be for the handler to return the partitions it wants to have assigned, which allows the handler to add/remove and set offsets.
But then we're basically back to assign().

mhowlett: I think returning the assignment as you suggest is quite a step forward from allowing the `Assign` call in the partitions assigned handler - I hadn't thought of this. I need to let the idea sink in a bit more, but at the moment I think I'm agreeing with you that this is the best option - a great one even.

mhowlett: downgrading 'great' to 'maybe the best compromise'. still thinking about it.

mhowlett: your proposal doesn't solve my main complaint (there still isn't clear separation of ideas), it just smudges over this a bit better. But you have given me another idea - and I can't see any compromise with it at all. How about we have a `SetGroupAssignmentHandler` in addition to `SetPartitionsAssignedHandler`. The group assignment handler is a filter which takes an arbitrary collection of `TopicPartition`s and returns an arbitrary collection of `TopicPartitionOffset`s. The `Consumer.Assign` happens between the two handlers - by the time execution gets to the partitions assigned handler, the partitions have already been assigned (just like the name suggests). 

Tell me why this isn't perfect!

```
.SetGroupAssignmentHandler(partitions => partitions.Select(tp => new TopicPartitionOffset(tp, Offset.Invalid)))
.SetPartitionsAssignedHandler((c, partitions) =>
{
    Console.WriteLine($"Partition assignment: [{string.Join(", ", partitions)}]");
})
.SetPartitionsRevokedHandler((c, partitions) =>
{
    Console.WriteLine($"Revoked partitions: [{string.Join(", ", partitions)}]");
})
```

mhowlett: (answering my own question): oh, you'd need to do a similar thing revoking... and we arrive back at what you suggested but with two extra handlers that are pretty superfluous and the awkward naming.

```
SetGroupAssignmentHandler
SetPartitionsAssignedHandler
SetGroupAssignmentRevokedHandler
SetPartitionsRevokedHandler
```

mhowlett: What about your idea exactly, but with names that are very differentiated (you didn't specify what the names should be, but I admit I was assuming you meant `SetPartitionsAssignedHandler`, `SetPartitionsRevokedHandler`). Your filter idea solves the unconventional `Assign` semantics problem and the names below solve the naming ambiguity problem:

```
using (var consumer = new ConsumerBuilder<Ignore, string>(config)
    .SetErrorHandler((_, e) => Console.WriteLine($"Error: {e.Reason}"))
    .SetGroupAssignmentReceivedHandler((c, partitions) =>
    {
        Console.WriteLine($"Partition assignment: [{string.Join(", ", partitions)}]");
        // optional:
        // return partitions.Select(tp => new TopicPartitionOffset(tp, Offset.Invalid));
    })
    .SetGroupAssignmentRevokedHandler((c, partitions) =>
    {
        Console.WriteLine($"Revoked partitions: [{string.Join(", ", partitions)}]");
        // optional:
        // return partitions.Select(tp => new TopicPartitionOffset(tp, Offset.Invalid));
    })
    .Build())
```

I think you won't like the word received because it's not used in other APIs. It acts to distance the name from something which occurs due to an action on the client though, so I quite like it.

something else: currently the assignment revoked handler only provides `TopicPartition`s. Maybe this should really be `TopicPartitionOffset`s.

mhowlett: Other potential names for the two handlers:

```
SetPartitionsAssignedHandler
SetPartitionsRevokedHandler
```
(but names don't differentiate from local assign)

```
SetGroupPartitionsAssignedHandler
SetGroupPartitionsRevokedHandler
```
(better, but the names in previous comment more clearly describe purpose).

Note: I (strongly) favor this new 'filter' approach over my suggestion in this PR since I don't think there is any compromise.

mhowlett: *assignment defn*:  "the allocation of a job or task to someone."
  - i.e. something you've been given to do, but haven't necessarily done yet. This word fits perfectly with the partition allocation given to us by the consumer group.

*assign defn*: "appoint (someone) to a particular job, task, or organization."
*assigned*:  past tense of assign.
 - i.e. assign is more ordering to do (fits perfectly with our use in the method `Assign`). Assigned is the past tense of that, so it's a perfect match for the property currently named `Assignment` (which I maintain is misleading due to the above definitions).

Also, I do like the word received because it further reinforces that the partition allocation came from somewhere else (which is does).

In light of that, I have a strong preference for:

```
SetGroupAssignmentReceivedHandler
SetGroupAssignmentRevokedHandler
```

And also changing the `Assignment` property to: `AssignedPartitions`

I don't think the latter is splitting hairs, I think it's strictly more correct.

mhowlett: in other words:

- assignment: something you've been given.
- assign: something you do.
- assigned: something you did.

PR #826, opened 2019-03-15
Rebalance API mk X

mhowlett:
Primarily changes the rebalance API to something which should be uncontroversial at this point.

Testing uncovered a problem (left as a todo to fix) - exceptions thrown in a librdkafka callback don't propagate through librdafka as i'd expect - they can't be caught (yet cause the program to terminate). This could be solved via additional bookkeeping but we can fix this in a future PR.

Also, a number of other smaller things:
1. `Consumer.Position` now operates only on a single `TopicPartition`, not a collection. This (A) actually does what the name says (B) matches the java API and (C) is no more difficult to use on a collection (because of LINQ) than the method that takes a collection whilst at the same time being easier to use for a single partition.
2. Similar changes to `StoreOffsets` (now `StoreOffset`). Similar reasons, except additionally, there is already an overload of `StoreOffset` which takes a `ConsumeResult`. Having both messys up the top level intellisense.
3. Change `Offset.Invalid` to `Offset.Unset`. There are two reasons. (A) The word Invalid is less appropriate in most of the places this constant is used. (B) The word Invalid is also a bit alarming - it risks users incorrectly inferring there is an issue when in fact there is not.
4. Fixed a mistake in the `AssignPastEnd` which has by fluke been passing 'till now due to semi-deterministic ordering of test execution.


test/Confluent.SchemaRegistry.Serdes.IntegrationTests/Tests/ConsumePartitionEOF.cs
```diff
@@ -103,13 +98,8 @@ public static void ConsumePartitionEOF(string bootstrapServers, string schemaReg
...
                         .SetValueDeserializer(new AvroDeserializer<User>(schemaRegistry))
-                        .SetRebalanceHandler((c, e) =>
-                        {
-                            if (e.IsAssignment)
-                            {
-                                c.Assign(e.Partitions.Select(tp => new TopicPartitionOffset(tp, Offset.Beginning)));
-                            }
-                        })
+                        .SetPartitionsAssignedHandler((c, partitions) =>
+                            c.Assign(partitions.Select(tp => new TopicPartitionOffset(tp, Offset.Beginning))))
```
mhowlett: whoops, this shouldn't be `Assign` here - will fix.

edenhill:
> Testing uncovered a problem (left as a todo to fix) - exceptions thrown in a librdkafka callback don't propagate through librdafka as i'd expect - they can't be caught (yet cause the program to terminate). 
> This could be solved via additional bookkeeping but we can fix this in a future PR.

If this the case where rd_kafka_poll() is called, triggers a callback, the managed code throws an exception, the callback and poll returns, but the exception is not properly propagated?

What we do in Python is have a bit of wrapping the callback  trampoline that checks if the managed code raised an exception, and if so save it on a callstate struct that is later from the code that called rd_kafka_poll() and then re-raises the exception, sort of.

src/Confluent.Kafka/ConsumerBuilder.cs
```diff
@@ -243,32 +251,113 @@ public ConsumerBuilder(IEnumerable<KeyValuePair<string, string>> config)
...
-        ///     a handler, partitions will be assigned (or unassigned) automatically, matching the
-        ///     assignment dictated by the consumer group. This default behavior will not occur if
-        ///     you call Assign yourself in the handler. The set of partitions you assign to is not
-        ///     required to match the assignment provided by the consumer group, but typically will.
-        ///     Note: you may call Assign or Unassign at most once in your rebalance handler.
-        /// 
-        ///     Consumption will resume from the last committed offset for each partition (unless an
-        ///     offset is manually specified in a call to Assign), or if there is no committed
-        ///     offset, in accordance with the `auto.offset.reset` configuration property.
+        ///     This handler is called when a new consumer group partition assignment has been received
```
edenhill: has been received by this consumer from the group leader.

src/Confluent.Kafka/ConsumerBuilder.cs
```diff
@@ -243,32 +251,113 @@ public ConsumerBuilder(IEnumerable<KeyValuePair<string, string>> config)
...
+        ///     by this consumer. Note: corresponding to every call to this handler there will be a
```
edenhill: Note on new line

src/Confluent.Kafka/ConsumerBuilder.cs
```diff
@@ -243,32 +251,113 @@ public ConsumerBuilder(IEnumerable<KeyValuePair<string, string>> config)
...
+        ///     SetPartitionsRevokedHandler).
+        ///
+        ///     The actual partitions to consume from and start offsets are specfied by the return value
+        ///     of the handler. This set of partitions is not required to match the assignment provided
+        ///     by the consumer group, but typically will. Partition offsets may be a specific offset, or
+        ///     special value (Beginning, End or Unset). If Unset, consumption will resume from the
+        ///     last committed offset for each partition, or if there is no committed offset, in accordance
+        ///     with the `auto.offset.reset` configuration property.
+        /// </summary>
+        /// <remarks>
```
edenhill: add a warning that assign() must not be called

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Assignment.cs
```diff
@@ -40,30 +41,23 @@ public void Consumer_Assignment(string bootstrapServers)
             {
                 GroupId = Guid.NewGuid().ToString(),
                 BootstrapServers = bootstrapServers,
-                SessionTimeoutMs = 6000
+                SessionTimeoutMs = 6000,
```
edenhill: With librdkafka v1.0.0 the default session timeout is 10s, so this is probably not needed anymore.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -129,40 +131,65 @@ private void LogCallback(IntPtr rk, SyslogLevel level, string fac, string buf)
...
-            }
-            else
-            {
-                throw new KafkaException(kafkaHandle.CreatePossiblyFatalError(err, null));
+                var assignTo = partitionsRevokedHandler(assignmentWithPositions);
+                lock (assignCallCountLockObj)
+                {
+                    if (assignCallCount > 0)
+                    {
+                        throw new InvalidOperationException("Assign/Unassign must not be called in the partitions revoked handler.");
```
edenhill: What about reversing the logic here and throwing the exception from assign instead?
And the current locking does not really work for the case where assign() is called from another thread while the rebalance handler is executing.
You could make assignCallCount a TLS variable instead to see if it was called from the handler or not.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Subscription.cs
```diff
@@ -45,21 +46,15 @@ public void Consumer_Subscription(string bootstrapServers)
 
             using (var consumer =
                 new ConsumerBuilder<byte[], byte[]>(consumerConfig)
-                    .SetRebalanceHandler((c, e) =>
+                    .SetPartitionsAssignedHandler((c, partitions) =>
```
edenhill: Would like to see a test where assign() is called from the handler, to see that the exception is properly thrown.

test/Confluent.Kafka.IntegrationTests/testconf.json
```diff
@@ -1,3 +1,3 @@
 {
-  "bootstrapServers": "localhost:9092"
+  "bootstrapServers": "127.0.0.1:9092"
```
edenhill: osx ipv6, huh?

test/Confluent.Kafka.VerifiableClient/Program.cs
```diff
@@ -308,18 +308,12 @@ public VerifiableConsumer(VerifiableConsumerConfig clientConfig)
...
-                    {
-                        HandleAssign(e.Partitions);
-                    }
-                    else
-                    {
-                        HandleRevoke(e.Partitions);
-                    }
-                })
-                .SetOffsetsCommittedHandler((_, offsets) => SendOffsetsCommitted(offsets))
+                .SetPartitionsAssignedHandler(
```
edenhill: neat

src/Confluent.Kafka/Consumer.cs
```diff
@@ -129,40 +131,65 @@ private void LogCallback(IntPtr rk, SyslogLevel level, string fac, string buf)
...
+                    }
+                }
+
+                // This distinction is important because calling Assign whilst the consumer is being
+                // closed (which will generally trigger this callback) is disallowed.
```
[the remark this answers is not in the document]
mhowlett: yes - this was not working using assign to an empty set.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Subscription.cs
mhowlett: I wrote these, but as noted the exception can't be caught and i'm deferring writing the exception wrapping / pass-through code until a later PR.

examples/Consumer/Program.cs
```diff
@@ -61,19 +61,17 @@ public static void Run_Consume(string brokerList, List<string> topics, Cancellat
                 // Note: All handlers are called on the main .Consume thread.
                 .SetErrorHandler((_, e) => Console.WriteLine($"Error: {e.Reason}"))
                 .SetStatisticsHandler((_, json) => Console.WriteLine($"Statistics: {json}"))
-                .SetRebalanceHandler((_, e) =>
+                .SetPartitionsAssignedHandler((c, partitions) =>
```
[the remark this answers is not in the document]
mhowlett: yes. the terminology can absolutely be improved on for clarity, which is the root cause of why it took so long for me to arrive at a clear opinion on this API, but these names sound good and are justifiable.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -129,40 +131,65 @@ private void LogCallback(IntPtr rk, SyslogLevel level, string fac, string buf)
...
-                        if (assignCallCount > 0) { return; }
+                        assignmentWithPositions.Add(new TopicPartitionOffset(tp, Offset.Unset));
                     }
                 }
-                Unassign();
-            }
-            else
-            {
-                throw new KafkaException(kafkaHandle.CreatePossiblyFatalError(err, null));
+                var assignTo = partitionsRevokedHandler(assignmentWithPositions);
```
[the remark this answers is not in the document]
mhowlett: yes: https://github.com/confluentinc/confluent-kafka-dotnet/pull/826/files#diff-cd35c16c2d0b4f2131b3dc8b4372e1aeR358

src/Confluent.Kafka/Consumer.cs
```diff
@@ -129,40 +131,65 @@ private void LogCallback(IntPtr rk, SyslogLevel level, string fac, string buf)
...
+            if (err == ErrorCode.Local_RevokePartitions)
             {
-                if (rebalanceHandler != null)
+                if (partitionsRevokedHandler == null)
+                {
+                    Unassign();
+                    return;
+                }
+
+                lock (assignCallCountLockObj) { assignCallCount = 0; }
```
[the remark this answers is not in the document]
mhowlett: good point, whoops.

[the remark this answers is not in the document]
mhowlett: i've just moved the lock further down. i don't think we can make this foolproof, it's just a guard against application programming errors, which should be effective in 99.9% of cases.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -379,32 +406,35 @@ public void Unassign()
         ///     Thrown if result is in error.
         /// </exception>
         public void StoreOffset(ConsumeResult<TKey, TValue> result)
-            => StoreOffsets(new[] { new TopicPartitionOffset(result.TopicPartition, result.Offset + 1) });
+            => StoreOffset(new TopicPartitionOffset(result.TopicPartition, result.Offset + 1));
```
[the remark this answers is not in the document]
mhowlett: `ConsumeResult` doesn't have an error property (errors are by exceptions now).

src/Confluent.Kafka/Consumer.cs
mhowlett: yep, all good points. it's like this because I just morphed earlier code. in the interests of velocity, i want to leave this as is (and will open an issue to track) - it's just a check for application logic errors and it'll work pretty much always.

PR #831, opened 2019-03-19
Fix Async hangs for MVC WebApi, Add WebApi and AWS Lambda examples.

BackTrak:
Hello, 

Thanks,
Nick.


BackTrak:
Sending an updated request with a better sample.

mhowlett:
Hi @BackTrak - thanks for picking up on the `GetResult` in `ProduceAsync` - that's a problem! I was thinking we weren't doing this. I'm about to open a PR that fixes this (a bit different to what you've done but along the same lines), and it'll be included in 1.0-RC1, which should be out today.

The primary downside of marking `ProduceAsync` as `async` is exceptions are wrapped in `AggregateException` in the case the method isn't awaited, but this is more a flaw / annoyance (done for backward compatibility reasons) in .NET.

Happy to review PRs for examples, but keep all PRs / examples as small and focused as possible.

PR #832, opened 2019-03-19

BackTrak:
Hello,

Thanks,
Nick.


BackTrak:
Failing travis build, will downgrade samples to .net core 2.1

PR #833, opened 2019-03-19

BackTrak:
Hello,

Thanks,
Nick.


Confluent.Kafka.sln
```diff
@@ -5,49 +5,53 @@ VisualStudioVersion = 15.0.26124.0
 MinimumVisualStudioVersion = 15.0.26124.0
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{1EFCD839-0726-4BCE-B745-1E829991B1BC}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Confluent.Kafka", "src\Confluent.Kafka\Confluent.Kafka.csproj", "{09C3255B-1972-4EB8-91D0-FB9F5CD82BCB}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Confluent.Kafka", "src\Confluent.Kafka\Confluent.Kafka.csproj", "{09C3255B-1972-4EB8-91D0-FB9F5CD82BCB}"
```
mhowlett: avoid white-space changes in PRs

examples/AvroSpecificLambda/Entities/Producer.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Web;
+
+namespace AvroSpecificLambda.Entities
+{
+    public class Producer
```
mhowlett: Producer isn't a good name for this

examples/AvroSpecificWebApi/Controllers/ProducerController.cs
```diff
@@ -0,0 +1,77 @@
...
+            };
+
+            // Note: The User class in this project was generated using the Confluent fork of the avrogen.exe tool 
+            // (avaliable from: https://github.com/confluentinc/avro/tree/confluent-fork) which includes modifications
+            // that prevent namespace clashes with user namespaces that include the identifier 'Avro'. AvroSerializer
+            // and AvroDeserializer are also compatible with classes generated by the official avrogen.exe tool 
+            // (available from: https://github.com/apache/avro), with the above limitation.
+
+            using (var schemaRegistry = new CachedSchemaRegistryClient(schemaRegistryConfig))
+            using (var producer =
```
mhowlett: creating a new producer for each request will be un-workably slow - there is a lot of overhead in establishing connections / querying API versions / getting metadata on startup. you should arrange things so there is one producer instance per application process, eg using application event hooks or a singleton class.

note that the kafka client doesn't work like a typical DB driver which typically manages a pool of open connections, giving you one of those when you request a new connection. the kafka client maintains a single open connection to each broker and makes maximal use of these connections (batching requests together etc).

BackTrak: Done! I will cancel this PR, and submit a new one with all changed. Thank you!

examples/AvroSpecificLambda/Entities/Producer.cs
BackTrak: Changed to ProducerSettings.

Confluent.Kafka.sln
BackTrak: I will revert, and hand edit to ensure VS doesn't modify existing project guids.

PR #835, opened 2019-03-19

BackTrak:
Hello,

Thanks,
Nick.


mhowlett:
thanks for the contribution - we do want to add WebAPI and AWS Lambda examples, we're just trying to get 1.0 out the door first. We typically prefer much smaller PRs so as to make reviewing more manageable. For that reason, I anticipate opening different PRs for this (will leave this PR open until we do so though). If you sign the contributor agreement, we can make use of material in this PR.

BackTrak:
[clabot:check]

BackTrak:
Thanks @mhowlett , sorry for the delay, I lost track of this thread.

AndyPook:
Nice. There's an awful lot in here.
Most of it is the 2 new examples, yes?
The core change has now been superseded by changes in origin.

Maybe there's a case for having a "Contrib" repo for this type of thing?
Maybe just a space in the top README listing links to examples like this?

BackTrak:
@AndyPook that is correct, the core change has already been added to the 1.0 release. The main contribution are the two examples which show how to correctly use the library. If the user doesn't specifically make their web api / lambda async, then the confluent library silently hangs (or it did the last time I checked, I have not tried with the current 1.0 release).  I can cancel this PR and resubmit with just the samples if that makes it easier?

BackTrak:
@AndyPook I can confirm that in the current version of the library, if you do not mark your controller methods with async, the call to ProduceAsync will silently hang indefinitely. The goal of my examples is to show people how to get this to work for WebApi and AWS Lambdas.

AndyPook:
@BackTrak If you're running this in "full-fat" net472 rather than Core then I think you're seeing the is the old async deadlock (https://blogs.msdn.microsoft.com/alazarev/2017/05/20/fun-with-configureawait-and-deadlocks-in-asp-net/) 
this is because it uses a single threaded SynchronizationContext and so the await continuation gets blocked. Core does not experience this as there is no sync context and the continuation is dispatched on the thread pool.
The "fix" is usually to include `.ConfigureAwait(false)` between the `ProduceAsync` and the `ContinueWith`.

Good luck 😄

BackTrak:
Thanks @AndyPook , I have seen that article as well. I can confirm that this lock up happens in both the Net472 and Core versions of the example when async is removed from the web methods. The library issue (there was one non-async call inside the SchemaRegistryClient that was locking up for Avro calls) was fixed in the 1.0.0 release. I agree this is a AspNet issue not directly related to the Confluent library. However, as there is no non-async methods on the Confluent library, this issue will stump many people trying to use it from inside WebAPI or Lamdba which is why I made the examples. My samples work fine, I wanted to share them with your users as well as I spent many hours working through the Confluent library to determine where the initial hang was coming from.

PR #839, opened 2019-03-21
improvements to produce/consume exceptions

mhowlett:
two things related to exceptions in this PR:

1. `ProduceException` and `ConsumeException` are used to expose all errors (except application logic errors) with the produce or consume methods - users need only check for this exception type, which additionally provides complete context (topic, partition etc), making handling errors at a higher level in the call stack very convenient. Documented this.

2. Removed `SerializationException` and `DeserializationException`. These really provide no value above the ErrorCode serialization codes. In fact, standard error types are better and using the custom serde types is worse because it implies they provide useful information, but they don't.


src/Confluent.Kafka/Consumer.cs
```diff
@@ -1100,19 +1100,35 @@ internal Consumer(ConsumerBuilder<TKey, TValue> builder)
...
-            TKey key = keyDeserializer != null
-                ? keyDeserializer.Deserialize(rawResult.Key, rawResult.Key == null, new SerializationContext(MessageComponentType.Key, rawResult.Topic))
-                : asyncKeyDeserializer.DeserializeAsync(new ReadOnlyMemory<byte>(rawResult.Key), rawResult.Key == null, new SerializationContext(MessageComponentType.Key, rawResult.Topic))
-                    .ConfigureAwait(continueOnCapturedContext: false)
-                    .GetAwaiter()
-                    .GetResult();
+            TKey key;
+            try
+            {
+                key = keyDeserializer != null
```
edenhill: These layered ternary if statements do not improve readability

src/Confluent.Kafka/Consumer.cs
```diff
@@ -1140,10 +1156,14 @@ internal Consumer(ConsumerBuilder<TKey, TValue> builder)
         ///     The consume result.
         /// </returns>
         /// <remarks>
-        ///     OnPartitionsAssigned/Revoked and OnOffsetsCommitted events may
+        ///     The partitions assigned/revoked and offsets commited handlers may
```
edenhill: committtttted

src/Confluent.Kafka/Consumer.cs
```diff
@@ -1140,10 +1156,14 @@ internal Consumer(ConsumerBuilder<TKey, TValue> builder)
...
         ///     be invoked as a side-effect of calling this method (on the same
         ///     thread).
         /// </remarks>
+        /// <exception cref="ConsumeException">
+        ///     Thrown if a consume request is unsuccessful for any reason. Inspect
```
edenhill: This makes it sound like there is a consume request sent to the broker, which only covers a subset of possible error cases.
Might want to rephrase as "Thrown when consumption fails..", "Thrown on consumer-level errors..", ..

src/Confluent.Kafka/Deserializers.cs
```diff
@@ -69,7 +58,7 @@ public Null Deserialize(ReadOnlySpan<byte> data, bool isNull, SerializationConte
             {
                 if (!isNull)
                 {
-                    throw new DeserializationException("Deserializer<Null> may only be used to deserialize data that is null.");
+                    throw new ArgumentException("Deserializer<Null> may only be used to deserialize data that is null.");
```
edenhill: Since the deserializer is not being passed data from this applicatoin I would argue that ArgumentException is incorrect here, TypeException is better.

src/Confluent.Kafka/Deserializers.cs
```diff
@@ -130,12 +119,12 @@ public int Deserialize(ReadOnlySpan<byte> data, bool isNull, SerializationContex
             {
                 if (isNull)
                 {
-                    throw new DeserializationException($"Null data encountered deserializing an Int32 value");
+                    throw new ArgumentNullException($"Null data encountered deserializing an Int32 value");
```
edenhill: drop "an"

src/Confluent.Kafka/Deserializers.cs
```diff
@@ -158,12 +147,12 @@ public float Deserialize(ReadOnlySpan<byte> data, bool isNull, SerializationCont
             {
                 if (isNull)
                 {
-                    throw new DeserializationException($"Null data encountered deserializing an float value.");
+                    throw new ArgumentNullException($"Null data encountered deserializing an float value.");
```
edenhill: drop an

src/Confluent.Kafka/Deserializers.cs
```diff
@@ -209,12 +191,12 @@ public double Deserialize(ReadOnlySpan<byte> data, bool isNull, SerializationCon
             {
                 if (isNull)
                 {
-                    throw new DeserializationException($"Null data encountered deserializing an double value.");
+                    throw new ArgumentNullException($"Null data encountered deserializing an double value.");
```
edenhill: drop an

src/Confluent.Kafka/Producer.cs
```diff
@@ -790,6 +804,22 @@ internal Producer(ProducerBuilder<TKey, TValue> builder)
         ///     A delegate that will be called with a delivery report corresponding
         ///     to the produce request (if enabled).
         /// </param>
+        /// <exception cref="ProduceException{TKey,TValue}">
+        ///     Thrown in response to any produce request error that is known
```
edenhill: "produce request" is somewhat problematic here since these errors are raised prior to any produce request.

Probably also want to say that these are local errors, not broker.

src/Confluent.Kafka/Producer.cs
```diff
@@ -790,6 +804,22 @@ internal Producer(ProducerBuilder<TKey, TValue> builder)
...
+        ///     produce requests us made available via the
+        ///     <paramref name="deliveryHandler" /> parameter (if specified). The
+        ///     Error property of the exception / delivery report provides more
+        ///     detailed information.
+        /// </exception>
+        /// <exception cref="ArgumentException">
+        ///     Thrown in response to invalid argument values.
+        /// </exception>
+        /// <exception cref="InvalidOperationException">
+        ///     Thrown in response to some error conditions that reflect an error in
```
edenhill: remove "some"

src/Confluent.Kafka/Consumer.cs
mhowlett: I find this very easy to read! It's more normal to write things like this in functional languages but C# is pretty functional.

src/Confluent.Kafka/Deserializers.cs
mhowlett: This is an ArgumentException WRT this function, so it's what this function should throw. It's the job of additional layers to wrap and re-throw as appropriate and include this as an InnerException for deeper information.

edenhill: Yeah, but in practice this will be seen as an (inner) ArgumentException from Consume(), which  would typically indicate an argument error to Consume().
I think the Serdes-exceptions were good, very clear what was wrong.

src/Confluent.Kafka/Producer.cs
mhowlett: yep good point. i didn't include the point about local errors since i think this is implied by 'known immediately' and i'm in a hurry :-)

src/Confluent.Kafka/Deserializers.cs
mhowlett: doesn't indicate that - the only time that would be the implication is if that exception is at the top level. Note that I think we may want to re-invent the Serialization and Deserialization exceptions in the future, but they should derive from ProduceException/ConsumeException. This is much better UX (as well as structurally better). Users who don't care can just do a blanket catch of ConsumeException, those who do can specialize. Specialization is possible now too of course, but the user needs to check the Error field (which is not idiomatic, but good enough for now and easily extended in this way). It's valuable for serde exceptions to derive from produce/consume exceptions because they carry with them context which may be useful in the handler, wherever that is.

mhowlett: We could also have Serialization/Deserialization exceptions as now that don't derive from produce/consume exceptions, but they would still be thrown by the producer/consumer (with the inner exception as in this PR). but as noted, i think the additional context is useful, so we'll want to re-implement these classes to derive from produce/consume exceptions.

PR #849, opened 2019-03-26
added idempotent producer example

mhowlett:
The method used to marshal a string is unnecessarily long-winded. It's like this because at one point I had a bug elsewhere causing the reason to not get through to the error handler and I was narrowing this down... I could change it back to using p/invoke magic and a StringBuilder, but all this is internal so I don't think it matters one bit, and the current way has the very marginal benefit of using UTF8 as the encoding, not whatever default .net uses (which is not that).


examples/IdempotentProducer/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+            {
+                BootstrapServers = bootstrapServers,
+                EnableIdempotence = true
+            };
+
+            using (var producer = new ProducerBuilder<Null, string>(config)
+                .SetErrorHandler((c, e) => 
+                {
+                    // Fatal errors will be reported first in this error handler, and are best 
+                    // handled here. Any use of a producer instance following a fatal error
```
edenhill: Not any use, only produce() will, and should, fail.
An application must be able to clean up and terminate cleanly.

examples/IdempotentProducer/Program.cs
```diff
@@ -0,0 +1,84 @@
...
+                })
+                .Build())
+            {
+                for (var msgCount = 0; ; msgCount += 1)
+                {
+                    var msg =  $"idempotent producer example message #{msgCount}";
+
+                    try
+                    {
+                        var dr = await producer.ProduceAsync(topicName, new Message<Null, string> { Value = msg });
```
edenhill: Im skeptical of show-casing sync producers, theres already a bunch of issues due to this where people complain about bad thruput.
We should show them the proper way, even if it is just a small example, because people build larger things on these small examples.

src/Confluent.Kafka/Internal/Util.cs
```diff
@@ -25,6 +25,13 @@ internal static class Util
     {
         internal static class Marshal
         {
+            public static byte[] StringToUtf8ByteArray(string s)
+            {
+                var bytes = new byte[Encoding.UTF8.GetByteCount(s) + 1];
```
edenhill: is the byte array zeroed out? (so we can be sure there is a trailing NUL)

mhowlett:
all good points. will redo and reopen.

PR #851, opened 2019-03-27
Fix mistaken assert in PartitionEOF.

Mousavi310:
It seems that we must assert cr2 instead of cr1. We already asserted cr1.


mhowlett:
you are indeed correct. can you sign the CLA and retarget to the 1.0.x branch? thanks!

Mousavi310:
Sure, I will sign it and resend my pull request.

Mousavi310:
[clabot:check]

mhowlett:
merge #852 instead

PR #858, opened 2019-04-01
fixed missing config parameters

mhowlett:


src/ConfigGen/Program.cs
```diff
@@ -40,6 +40,7 @@ internal static List<PropertySpecification> RemoveLegacyOrNotRelevant(List<Prope
                 if (p.Name == "request.required.acks") { return false; }
                 // legacy
                 if (p.Name == "consume.callback.max.messages") { return false; }
+                if (p.Name == "offset.store.method") { return false; }
```
mhowlett: the incorrect removeDuplicateTopicLevel implementation was by chance omitting this too. not removeDuplicateTopicLevel is fixed, this needs to be added here.

src/ConfigGen/Program.cs
```diff
@@ -621,6 +625,13 @@ static List<PropertySpecification> choosePreferredNames(List<PropertySpecificati
             }).ToList();
         }
 
+        static void PrintProps(IEnumerable<PropertySpecification> props)
```
mhowlett: these changes make debugging easier. since this code isn't very robust to changes in librdkafka config docs, I expect more debugging will be needed in the future, so leaving these changes in.

src/ConfigGen/Program.cs
```diff
@@ -230,6 +231,9 @@ public PropertySpecification(PropertySpecification other)
         public string Description { get; set; }
         public string Type { get; set; }
         public string AliasFor { get; set; }
+
+        public int CompareTo(object obj)
+            => Name.CompareTo(((PropertySpecification)obj).Name);
```
edenhill: Should this support aliases? I.e., `PropertySpecification("acks") == PropertySpecification("request.required.acks")`?

src/Confluent.Kafka/Config_gen.cs
```diff
@@ -1,4 +1,4 @@
-// *** Auto-generated from librdkafka branch v1.0.0-RC7 *** - do not modify manually.
+// *** Auto-generated from librdkafka branch v1.0.0 *** - do not modify manually.
```
edenhill: drop 'branch'

edenhill:
This is good to go if you add a test case for the enums that wouldve caught the reported issue.

src/ConfigGen/Program.cs
mhowlett: no, they're de-duped and that's done elsewhere. the semantics we want for this comparison is this.

PR #867, opened 2019-04-06
serde changes

mhowlett:
I've been building some lightweight abstractions to facilitate stateful stream processing in .NET. As part of that effort, i've been making a lot of composite serdes and this has highlighted some API usability issues. So we need an RC4.

1. `SerializationContext` was included in the non-async serde interface for consistency with the async serdes, but it'll never be used. It's been annoying to specify `SerializationContext.Empty` everywhere in these composite serdes (not to mention a bit less efficient), so I took this out. This really makes the non-async serde abstractions more generally useful. Note: `SerializationContext` is still an awesome abstraction and a great way to think about the async case.

2. In the async case, having serdes as classes (implementing interfaces) is a good idea because they typically need to own state, and using closures proved to be awkward in achieving this. The additional justification for interfaces is that they can be extended (are future proof), whereas delegates can't. That argument was enough for us to go with interfaces for the non-async case as well (particularly as the serializer interface isn't optimal and just something we're using until we implement something better). However, that reasoning was made before we introduced the builder classes. These provide an alternate (and better) point of extensibility - allow any number of different serde interfaces / delegates. The downside of interfaces is it means specifying the serdes is more verbose. I was arguing this is not a big deal (it's not something that needs to happen often), which is still true, but that said I've been specifying these a fair bit in the above mentioned project, and it's meaningfully nicer for them to be straight functions.

3. Our serializer API is not as efficient as it could be. This is a known issue - we intentionally cut scope to get 1.0 out sooner since working out what is best and implementing it is non-trivial. Why bring this up now? Creating composite serializers has really highlighted this inefficiency - there's no way around creating lots of temporary `byte[]` objects. I'm not proposing fixing this now, but i'm proposing changing the name of the delegate to `SimpleSerializer`, leaving the name `Serializer` open for the final delegate type. Note, the deserializer side is easier and already very efficient (so no name change there).

### DIFF
--- a/examples/Protobuf/Program.cs
+++ b/examples/Protobuf/Program.cs
@@ -29,31 +29,6 @@ using System.Threading.Tasks;
 /// </summary>
 namespace Confluent.Kafka.Examples.Protobuf
 {
-    /// <summary>
-    ///     protobuf serializer
-    /// </summary>
-    public class ProtobufSerializer<T> : ISimpleSerializer<T> where T : IMessage<T>, new()
-    {
-        public byte[] Serialize(T data)
-            => data.ToByteArray();
-    }
-
-    /// <summary>
-    ///     protobuf deserializer
-    /// </summary>
-    public class ProtobufDeserializer<T> : IDeserializer<T> where T : IMessage<T>, new()
-    {
-        private MessageParser<T> parser;
-
-        public ProtobufDeserializer()
-        {
-            parser = new MessageParser<T>(() => new T());
-        }
-
-        public T Deserialize(ReadOnlySpan<byte> data, bool isNull)
-            => parser.ParseFrom(data.ToArray());
-    }
-
     class Program
     {
         static async Task Main(string[] args)
@@ -65,12 +40,14 @@ namespace Confluent.Kafka.Examples.Protobuf
                 AutoOffsetReset = AutoOffsetReset.Latest
             };
 
+            var messageParser = new MessageParser<User>(() => new User());
+
             var consumeTask = Task.Run(() =>
             {
                 // consume a single message then exit.
                 using (var consumer =
                     new ConsumerBuilder<int, User>(consumerConfig)
-                        .SetValueDeserializer(new ProtobufDeserializer<User>())
+                        .SetValueDeserializer((data, isNull) => messageParser.ParseFrom(data.ToArray()))
                         .Build())
                 {
                     consumer.Subscribe("protobuf-test-topic");
@@ -86,7 +63,7 @@ namespace Confluent.Kafka.Examples.Protobuf
 
             using (var producer =
                 new ProducerBuilder<int, User>(producerConfig)
-                    .SetValueSerializer(new ProtobufSerializer<User>())
+                    .SetValueSerializer(data => data.ToByteArray())
                     .Build())
             {
                 await producer.ProduceAsync("protobuf-test-topic", new Message<int, User> { Key = 0, Value = new User { FavoriteColor = "green" } });
--- a/examples/Protobuf/Program.cs
+++ b/examples/Protobuf/Program.cs
@@ -32,9 +32,9 @@ namespace Confluent.Kafka.Examples.Protobuf
     /// <summary>
     ///     protobuf serializer
     /// </summary>
-    public class ProtobufSerializer<T> : ISerializer<T> where T : IMessage<T>, new()
+    public class ProtobufSerializer<T> : ISimpleSerializer<T> where T : IMessage<T>, new()
     {
-        public byte[] Serialize(T data, SerializationContext context)
+        public byte[] Serialize(T data)
             => data.ToByteArray();
     }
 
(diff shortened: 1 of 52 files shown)


examples/Protobuf/Program.cs
```diff
@@ -65,12 +40,14 @@ static async Task Main(string[] args)
...
 
+            var messageParser = new MessageParser<User>(() => new User());
+
             var consumeTask = Task.Run(() =>
             {
                 // consume a single message then exit.
                 using (var consumer =
                     new ConsumerBuilder<int, User>(consumerConfig)
-                        .SetValueDeserializer(new ProtobufDeserializer<User>())
+                        .SetValueDeserializer((data, isNull) => messageParser.ParseFrom(data.ToArray()))
```
edenhill: What if the deserializer needs something else from the message, i.e., the value schema is in a message header, how would that look?

src/Confluent.Kafka/Producer.cs
```diff
@@ -39,20 +39,20 @@ internal class Config
             public Action<string> statisticsHandler;
         }
 
-        private ISerializer<TKey> keySerializer;
-        private ISerializer<TValue> valueSerializer;
+        private SimpleSerializer<TKey> keySerializer;
```
edenhill: I dont like the word Simple here, it doesn't imply what is simple. Is it the type, or the serialization that is simple?

How about Base, Scalar, Primitive, Builtin, .. Serializer?

src/Confluent.Kafka/SimpleSerializers.cs
```diff
@@ -0,0 +1,146 @@
+// Copyright 2018 Confluent Inc.
```
edenhill: 2019, or thereabout.

examples/Protobuf/Program.cs
mhowlett: the use-case you mention is something it'd be great to cater for, though note that it isn't actually covered by the `SerializationContext` struct as it currently stands. there are two ways the API could be extended to support this. 1. extend the `SerializationContext` struct. 2. Add an additional deserializer type.

the motivation for this PR is that the simple case is going to be very commonly needed (the most commonly needed), and common things should be easy, so a separate type is justified. The use case you bring up can be just as easily catered for by adding a the more comprehensive delegate type later (+ likely extending `SerializationContext`). Alternatively, we could require people use `IAsyncDeserializer` for this use case (not as good - i don't see a good reason for limiting the number of deserializer types, if they serve a purpose).

src/Confluent.Kafka/Producer.cs
mhowlett: I can leave the name as `Serializer`, it's not a huge deal. In the future, we could have `NoAllocSerializer` or `ZeroAllocSerializer` or `PipelineSerializer` or `SpanSerializer` or however we decide to implement it. Also, it's likely we may never implement this (certainly not soon - there are bigger priorities).

The primary message i was trying to get across here is that the serializer returns `byte[]` - is not as optimal as some more complex API that would allow allocations to be avoided in the serializer. It's also simple in the sense that it doesn't take a `SerializationContext`, but the delegate type to replace it also won't take a `SerializationContext`.

perhaps the word `basic` is an alternative? your suggestions seem to be getting more at the lack of `SerializationContext`.

The word `Simple` worked for me as it's a bit apologetic - when people take issue with it, it suggests we've though of that and something better is coming.

mhowlett: I'll revert to `Serializer`.

src/Confluent.Kafka/SimpleSerializers.cs
mhowlett: really 2018 - 2019 here.

AndyPook:
Changing to delegates from interfaces is a huge breaking change really messes with what I'm working on. Things lik:e being able to resolve a set of serdes from a DI container; Having types and pattern matching; Resolving types from config with `Type.GetType(name)`  etc etc

Having `SerializationContext` is generally useful in both sync and async. No need to specify `SerializationContext.Empty` everywhere: just `default` will work fine; Or make the arg optional `... , SerializationContext ctx = default)`; Or add an extension method to do it for you.

The `ReadOnlyMemory` vs `ReadOnlySpan` thing is a pita. It would be nice to make the arg signatures identical. No idea how to do that though :(

+1,000,000 for leaving them as interfaces

mhowlett:
thanks for the feedback. keep it coming!

I expect to add `SerializationContext` on the sync interface as a solution to the async problems (based on idea from you). Given your use cases for interfaces, i'll reinstate that as well. May or may not also have a `Deserializer` delegate type, which is also useful.

AndyPook:
cool...

I guess it'd be simple enough to create some helper/adapters to "convert" from one to the other.

Q: Is it possible t describe the situation where delegates are working better for you?

mhowlett:
i'm working on a mini stream processing framework. In that, i'm creating composite serdes in a fair few places and the `SerializationContext` was getting in the way. Also, the code could be shortened a fair bit by specifying these inline with delegates in places.

AndyPook:
Sounds like another builder/fluent interface would work well. Wrapping
whatever delegate signatures you need with classes impl the interfaces

```
class DelegateSerializer<T>:ISerializer<T>
{
  private ro delegate serDelegate;
  ctor(delegate)
  {
    serDelegate = delegate;
  }
  public void Serialize(...) => serDelegate(...);
}
```

PR #877, opened 2019-04-10
Extensible avro serializer/deserializer 

imhotepper:
Updated the SerializeAsync and DeserializeAsync for the Avro serializer/deserializer so that they are extensible. As a use case, now it is possible to add encryption/decryption to the messages sent to and from the topics.


imhotepper:
[clabot:check]

mhowlett:
can you use composition for this instead?

imhotepper:
indeed I can use composition. thanks for the feed back. I will close the pull request.

mhowlett:
cool. i think composition here is better, so i think better to have an API which pushes people towards that.

PR #878, opened 2019-04-11
use IConsumer/IProducer

AndyPook:
Use interfaces in Set*Handler delegates


mhowlett:
LGTM - thanks! great catch.

PR #880, opened 2019-04-11
addressing sync over async issues [WIP]

mhowlett:
changes:

- see changelog.

notes:
- sync serdes won't deadlock because our implementation has `ConfigureAwait(false)` all the way down, but it's not truly 'synchronous' because continue on captured context is false. most people won't care, but it's still good that the cause of this behaviour has now been pulled out of the main client classes.
- doesn't require re-writing schema registry integration to be sync (which is a step backwards imo).
- client classes don't incorporate workarounds.
- the API is the same as before, except now non-async methods can cause a runtime error if async serdes configured.
- it's on the user if they use sync over async inappropriately. there's nothing we can do to help out here -this is a tricky thing about working in .net.
- if we aren't able to ever implement `ConsumeAsync` for technical reasons, we probably want to get rid of `IAsyncDeserializer` now. If we do plan to implement this, I don't see much harm in leaving this code in. I've restricted the `ConsumerBuilder` so it can't be used (but its still used in the schema registry serdes which build on it and wrap it).

@AndyPook - this is based on your ideas. what's your opinion on this PR?


AndyPook:
Nice. I've a branch with almost the same kind of thing in it :)
Leaving the Consumer a sync only and making the sync-over-async aspect be the "problem" of the serdes seems a reasonable way to go.
Though I have `SyncOverAsyncSerializer<T>` types that look like the Wrappers that were here for a while. I do like this approach as it means there's only one version of the avro serde types.
It also means that the `SetKeyDeserializer` type methods on the builders can just take `IAsyncDeserializer<>`; decorate them with the `SyncOverAsync...` type; and assign to the KeyDeserializer prop. Neat.

Q: the builder/ctor will guarantee that the serdes are set. So is it necessary to do the check/throw in the Consume methods?

I see that this is a wip...
there's some tidyup needed to remove the IAsync... refs from the Consumer. There're also some comments like "// fast path for simple case" that aren't valid any more.

Anyway +1 for this approach.
Let me know if there's any way I can help

mhowlett:
so the thread pool starvation deadlock issue still exists even without using `Task.Run`. If `Producer.BeginProduce` method is being used on thread pool threads together with the 'sync' avro serializer in this PR and we utilize all available thread pool threads in doing those calls, then still, none of them will be able to unblock in the case where completion of the tasks depends on use of a thread pool thread (which is evidently usual).

So i think to make the sync avro case completely fool proof, we need to use an http library that isn't async under the covers (doesn't ever try to grab thread pool threads), but that means having a second implementation of the schema registry client or having async-over-sync in the async case, which also means majorly changing all of that code.

So the best alternative to me at the moment is seeming like reverting to an earlier commit in this PR (also a suggestion by @AndyPook ) of just having a wrapper helper method that converts async to sync serdes. We call it `SyncOverAsync` and have warnings on use. The problem is for people to use avro deserializers now, they must use this. Our simple examples will be littered with this `SyncOverAsync` bizzo.

Finally we could implement `ConusmeAsync`. We could do this now if it's blocking, but as I say elsewhere, I really don't want that as i think it goes against people's expectations (it's muddling together sync and async behavior) and is bad UX. Also it's only a fix in that it removes all the above issues from the library - there is still no way to use the avro serdes synchronously without caveats.

mhowlett:
liking your `SyncOverAsyncSerializer<T>` class idea. will update PR when it's not the middle of the night :-).

AndyPook:
The starvation aspect is quite confusing.
Do you have a little program that demonstrates this?

mhowlett:
Stephen Toub's blog post goes into it in depth: https://devblogs.microsoft.com/pfxteam/should-i-expose-synchronous-wrappers-for-asynchronous-methods/

I've not tried to make this happen in practice - about to do that now.

mhowlett:
just added a test commit. the deadlock happens for me in the integration test for SyncOverAsync as predicted. It seems due to threadpool exhaustion as it doesn't happen when the number of in-flight requests is taken down to just below the number of thread pool threads.

What I didn't expect and don't understand yet is why I don't see the same problem in the console app.

AndyPook:
yep, I understand the theory. I was wondering if you had some specific code
that you use to verify the problem within this library

mhowlett:
a minimal program demonstrating the problem is here (can tweak various things, and it all behaves according to the theory):
https://github.com/confluentinc/confluent-kafka-dotnet/pull/880/files#diff-bd3daa9f062c08537db9ed32e8b01165

since the avro serdes will suspend very infrequently in practice (though to be fair, in a typical scenario, parallel executing BeginProduce requests will all want to suspend together, if that is required), and the default threadpool size is 32k (on my machine at least), I think this is not normally going to be a problem in practice. An example scenario where this could be a problem is a high volume web server logging metrics to kafka - the web server is turned on and 32k simultaneous produce requests happen before the first has a chance to get a response from schema registry. If schema registry is down, that is more likely. 

i think an API that documents what's going on (has the `SyncOverAsync` terminology) and a plan to implement `ConsumeAsync` in the future is an appropriate design decision given where we're at (and the pressing need to release 1.0).

AndyPook:
Thanks for the explanation...

I may be missing some detail and nuance...

Do we get the same starvation using the `ProduceAsync` rather than the `BeginProduce` path?
It looks like BeginProduce is effectively ProduceSync. So it has the same sync over async issue as Consume does. So it's just providing both a sync and async api.

Ideas...
Implement BeginProduce (just call it "Produce") in the same pattern as Consumer.Consume ie only support ISerializer (in this path) and create a SyncOverAsyncSerializer to "hide" the aync shenanigans? Doesn't actually resolve the issue but simplifies Producer by isolating the code in the same way?

Another theory could be to just implement BeginProduce as explicitly sync over async with (crudely)
```
void BeginProduce(...) => ProduceAsync(...).ConfigureAwait(false).GetAwaiter().GetResult();
```
If this is acceptable then just remove it from Producer and do it as an extension method called "Produce".

Or go full APM
```
void BeginProduce(...) => _ = ProduceAsync(...).ContinueWith(t=>EndProduce(t.Result));
```
Although this would mean that BeginProduce would return immediately so it's a change in behavior.

Personally... I think I'd just remove BeginProduce altogether. Leave it as a client thing

mhowlett:
- I'm using `BeginProduce` instead of `Consume` for testing because it's easier. same problem, yes.
- `BeginProduce` is staying because it's much faster than `ProduceAsync` and ordering of delivery reports is guaranteed (callbacks are all on the same thread), which is sometimes important.
- `ProduceAsync` doesn't have these problems, no (unless it's called synchronously).
- yes, i have a `SyncOverAsyncSerializer` class, and async serializers are forbidden for use with `BeginProduce`.
- `BeginProduce` is perhaps not the most ideal name because it's a different pattern to APM.

mhowlett:
tidied this up and opened for review in #887

PR #881, opened 2019-04-12
rework ctors for Config classes

AndyPook:
* delegate to base ctors
 * simplify the creation of the kvp -> dictionary
 * add ability to update a dictionary "in place"
Last one is handy for config validation, setting defaults etc in dotnetCore IOptions stuff

### DIFF
--- a/src/ConfigGen/Program.cs
+++ b/src/ConfigGen/Program.cs
@@ -420,11 +420,19 @@ namespace Confluent.Kafka
 
         static string createClassConstructors(string name)
         {
-            var codeText = "\n";
-            codeText += $@"        /// <summary>
+            var codeText = $@"
+        /// <summary>
+        /// Creates a new {name} class by taking a copy of the provided <paramref name=""configSource""/>.
+        /// The <paramref name=""configSource""/> can be any of: IEnumerable; Dictonary; any of the Config types.
+        /// </summary>
+        /// <param name=""configSource""></param>
+        /// <returns></returns>
+        public static Config CopyFrom(IEnumerable<KeyValuePair<string, string>> configSource) => new Config(configSource.ToDictionary(a => a.Key, a => a.Value));
+
+        /// <summary>
         ///     Initialize a new empty <see cref=""{name}"" /> instance.
         /// </summary>
-        public {name}() {{ }}
+        public {name}() : base() {{ }}
 
         /// <summary>
         ///     Initialize a new <see cref=""{name}"" /> instance wrapping
@@ -439,13 +447,6 @@ namespace Confluent.Kafka
         ///     This will change the values ""in-place"" i.e. operations on this class WILL modify the provided collection
         /// </summary>
         public {name}(IDictionary<string, string> config) : base(config) {{ }}
-
-        /// <summary>
-        ///     Initialize a new <see cref=""{name}"" /> instance copying
-        ///     an existing key/value pair collection.
-        ///     This will make a copy of the provided values i.e. operations on this class WILL NOT modify the provided collection
-        /// </summary>
-        public {name}(IEnumerable<KeyValuePair<string, string>> config) : base(config) {{ }}
 ";
             return codeText;
         }
--- a/src/ConfigGen/Program.cs
+++ b/src/ConfigGen/Program.cs
@@ -427,22 +427,33 @@ namespace Confluent.Kafka
         public {name}() {{ }}
 
         /// <summary>
-        ///     Initialize a new <see cref=""{name}"" /> instance based on
+        ///     Initialize a new <see cref=""{name}"" /> instance wrapping
         ///     an existing <see cref=""ClientConfig"" /> instance.
+        /// </summary>
+        /// <summary>
+        ///     <para>
+        ///     This will change the values ""in-place"" i.e. the original will be modified
+        ///     </para>
         /// </summary>
         public {name}(ClientConfig config) : base(config) {{ }}
 
         /// <summary>
-        ///     Initialize a new <see cref=""{name}"" /> instance based on
+        ///     Initialize a new <see cref=""{name}"" /> instance wrapping
         ///     an existing key/value pair collection.
-        /// </summary>
-        public {name}(IEnumerable<KeyValuePair<string, string>> config) : base(config) {{ }}
-
-        /// <summary>
-        ///     Initialize a new <see cref=""{name}"" /> wrapping
-        ///     an existing key/value pair collection.
+        ///     <para>
+        ///     This will change the values ""in-place"" i.e. the original will be modified
+        ///     </para>
         /// </summary>
         public {name}(IDictionary<string, string> config) : base(config) {{ }}
+
+        /// <summary>
+        ///     Initialize a new <see cref=""{name}"" /> instance copying
+        ///     an existing key/value pair collection.
+        ///     <para>
+        ///     This will make a copy of the provided values i.e. the original will be NOT modified
+        ///     </para>
+        /// </summary>
+        public {name}(IEnumerable<KeyValuePair<string, string>> config) : base(config) {{ }}
 ";
             return codeText;
         }
--- a/src/ConfigGen/Program.cs
+++ b/src/ConfigGen/Program.cs
@@ -326,13 +326,6 @@ namespace Confluent.Kafka
         static string createConfigPropertyNames(IEnumerable<PropertySpecification> props)
         {
             var codeText = "\n";
-            codeText += $"    public partial class Config\n";
-            codeText += $"    {{\n";
-            codeText += $"        /// <summary>\n";
-            codeText += $"        ///     A reference to the configuration key names\n";
-            codeText += $"        /// </summary>\n";
-            codeText += $"        public static partial class PropertyNames\n";
-            codeText += $"        {{\n";
 
             var codeTextC = string.Empty;
             var codeTextP = string.Empty;
@@ -366,32 +359,38 @@ namespace Confluent.Kafka
             codeText += CreateClass("Producer", codeTextP) + "\n";
             codeText += CreateClass("Admin", codeTextA);
 
-            codeText += $"        }}\n";
-            codeText += $"    }}\n";
+            //codeText += $"        }}\n";
+            //codeText += $"    }}\n";
             return codeText;
 
             string CreateClass(string name, string code)
             {
                 return
-                    $"            /// <summary>\n" +
-                    $"            ///     A reference to the key names for {name} specific configuration properties.\n" +
-                    $"            /// </summary>\n" +
-                    $"            public static partial class {name}\n" +
-                    $"            {{\n" +
+                    $"    /// <summary>\n" +
+                    $"    ///     A reference to the key names for {name} specific configuration properties.\n" +
+                    $"    /// </summary>\n" +
+                    $"    public partial class {name}Config\n" +
+                    $"    {{\n" +
+                    $"        /// <summary>\n"+
+                    $"        ///     A reference to the configuration key names\n"+
+                    $"        /// </summary>\n"+
+                    $"        public static partial class PropertyNames\n"+
+                    $"        {{\n"+
                     code +
-                    $"            }}\n";
+                    $"        }}\n"+
+                    $"    }}\n";
             }
 
             string CreateProperty(PropertySpecification prop)
             {
                 return
-                    $"                /// <summary>\n" +
-                    $"                ///     {prop.Description}\n" +
-                    $"                ///\n" +
-                    $"                ///     default: {(prop.Default == "" ? "''" : prop.Default)}\n" +
-                    $"                ///     importance: {prop.Importance}\n" +
-                    $"                /// </summary>\n" +
-                    $"                public const string {ConfigNameToDotnetName(prop.Name)} = \"{prop.Name}\";\n\n";
+                    $"            /// <summary>\n" +
+                    $"            ///     {prop.Description}\n" +
+                    $"            ///\n" +
+                    $"            ///     default: {(prop.Default == "" ? "''" : prop.Default)}\n" +
+                    $"            ///     importance: {prop.Importance}\n" +
+                    $"            /// </summary>\n" +
+                    $"            public const string {ConfigNameToDotnetName(prop.Name)} = \"{prop.Name}\";\n\n";
             }
         }
 
@@ -486,7 +485,7 @@ namespace Confluent.Kafka
             codeText += $"    /// <summary>\n";
             codeText += $"    ///     {docs}\n";
             codeText += $"    /// </summary>\n";
-            codeText += $"    public class {name}{(derive ? " : ClientConfig" : " : Config")}\n";
+            codeText += $"    public partial class {name}{(derive ? " : ClientConfig" : " : Config")}\n";
             codeText += $"    {{";
             return codeText;
         }
--- a/src/ConfigGen/Program.cs
+++ b/src/ConfigGen/Program.cs
@@ -443,7 +443,7 @@ namespace Confluent.Kafka
         /// <summary>
         ///     Initialize a new <see cref=""{name}"" /> instance copying
         ///     an existing key/value pair collection.
-        ///     This will make a copy of the provided values i.e. the original will be NOT modified
+        ///     This will make a copy of the provided values i.e. operations on this class will NOT modify the original
         /// </summary>
         public {name}(IEnumerable<KeyValuePair<string, string>> config) : base(config) {{ }}
 ";
--- a/src/ConfigGen/Program.cs
+++ b/src/ConfigGen/Program.cs
@@ -322,6 +322,33 @@ namespace Confluent.Kafka
             return result;
         }
 
+        static string createConfigKeys(IEnumerable<PropertySpecification> props)
+        {
+            var codeText = "\n";
+            codeText += $"    public partial class Config\n";
+            codeText += $"    {{\n";
+            codeText += $"        /// <summary>\n";
+            codeText += $"        ///     A reference to the configuration key names\n";
+            codeText += $"        /// </summary>\n";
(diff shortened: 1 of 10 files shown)


mhowlett:
thanks! looks good at quick glance

it's not breaking though (right?) and too big to merge before 1.0, will do after.

AndyPook:
not breaking, 100% compatible both in interface an behavior.
Just uses ctor chaining to simplify; removes the Dictionary/ToDictionary over cloning; and the new bit allows use of Config to Get/Set on a dictionary in-place (ie doesn't clone the original).

It's the last bit I'd really like. Saves me from having to duplicate key strings; better looking code in places like service startup config validation.

100% compatible, 0% risk 😄

mhowlett:
good candidate for 1.0.1 along with e.g. my #876 which i also desperately want. only breaking changes are going in 1.0 from here. Your tostring change probably should have been left open, but it was trivial.

src/ConfigGen/Program.cs
```diff
@@ -254,7 +254,7 @@ static string parseType(string type)
         static string createFileHeader(string branch)
         {
             return
-@"// *** Auto-generated from librdkafka " + branch + @" *** - do not modify manually.
+@"// *** Auto-generated from librdkafka branch " + branch + @" *** - do not modify manually.
```
mhowlett: took the word 'branch' out before you forked evidently. should have no diff here.

src/ConfigGen/Program.cs
```diff
@@ -418,26 +418,40 @@ static string createClassHeader(string name, string docs, bool derive)
...
         ///     an existing key/value pair collection.
         /// </summary>
-        public ConsumerConfig(IEnumerable<KeyValuePair<string, string>> config) { this.properties = new Dictionary<string, string>(config.ToDictionary(a => a.Key, a => a.Value)); }
+        public {name}(IEnumerable<KeyValuePair<string, string>> config) : base(config) {{ }}
+
+        /// <summary>
+        ///     Initialize a new <see cref=""{name}"" /> wrapping
+        ///     an existing key/value pair collection.
+        /// </summary>
+        public {name}(IDictionary<string, string> config) : base(config) {{ }}
```
mhowlett: The reasoning behind using `IEnumerable<KeyValuePair<string, string>>` is that it's the most general, so covers the most cases. In particular, `IDictionary<string,string>` extends this. Why is another constructor overload necessary?

That said, people would find `IDictionary` more friendly, and i might have made an error in judgement if `IEnumerable` isn't useful in practice (i can't recall).

src/Confluent.Kafka/Config_gen.cs
```diff
@@ -223,6 +223,29 @@ public enum Acks : int
     public class ClientConfig : Config
     {
 
+        /// <summary>
+        ///     Initialize a new empty <see cref="ClientConfig" /> instance.
+        /// </summary>
+        public ClientConfig() { }
```
mhowlett: yeah, ctor chaining seems a worthwhile improvement.

src/ConfigGen/Program.cs
mhowlett: Oh, if you pass in an `IDictionary` , then there's no copy so you could modify the dictionary independently and the changes would be reflected in the config classes. This does seem like it could be useful and a good change. But I don't like that it works like that is you pass a dictionary and doesn't if you pass an `IEnumerable`. I think it would be better to change the get/set methods to work with the IEnumerable rather than a dictionary and change the `properties` field to be an `IEnumerable`. This would be O(N) of course, but efficiency doesn't matter here at all, and linear scan isn't going to be meaningfully slow anyway since there's so few items. As an added bonus, there are less constructor overloads.

mhowlett:
thanks for the PR! have a couple of requested changes, and then more than happy to merge.

src/ConfigGen/Program.cs
AndyPook: indeed this is the idea behind this ctor ie values are changed in-place. I'll add some xmldoc to highlight this difference.

It is not possible to do the same thing with the IEnumerable.
KeyValuePair is a "value type", is it's immutable. It is not possible to change the values in-place.

So the ctors that take IEnumerable _must_ do a copy

src/ConfigGen/Program.cs
AndyPook: removed the "Branch" text

src/ConfigGen/Program.cs
AndyPook: I've also made the `Config(Config ...)` ctor behave "in-place" too. It is a change in approach but I don't think this will affect any "normal" code code/samples I've seen.

Also added a `Clone()` helper method for when someone _really_ needs the "copy" style.

src/Confluent.Kafka/Config_gen.cs
AndyPook: 😄

src/ConfigGen/Program.cs
mhowlett: ahh crap, yeah, ok - your modifications make sense.

src/Confluent.Kafka/Config.cs
```diff
@@ -43,14 +43,35 @@ public class Config : IEnumerable<KeyValuePair<string, string>>
         /// <summary>
         ///     Initialize a new <see cref="Config" /> instance based on
         ///     an existing <see cref="Config" /> instance.
+        ///     <para>
```
mhowlett: appreciate that using `<para>` is a step forward, but i haven't used it anywhere else, so i think it would be good to remove these for consistency. does this result in meaningfully better docs anywhere?

src/Confluent.Kafka/Config.cs
```diff
@@ -43,14 +43,35 @@ public class Config : IEnumerable<KeyValuePair<string, string>>
...
+        ///     </para>
+        /// </summary>
+        public Config(IEnumerable<KeyValuePair<string, string>> config) { this.properties = config.ToDictionary(a => a.Key, a => a.Value); }
+
+        /// <summary>
+        /// Creates a copy of this instance.
         /// </summary>
-        public Config(IEnumerable<KeyValuePair<string, string>> config) { this.properties = new Dictionary<string, string>(config.ToDictionary(a => a.Key, a => a.Value)); }
+        /// <returns>A new instance of this config with a copy of the values from this instance.</returns>
+        public Config Clone() => new Config(properties.ToDictionary(c => c.Key, c => c.Value));
```
mhowlett: I think we can leave `Clone` out - don't think it's useful enough.

And yeah, although your change to the `Config` constructor is technically breaking, as you say, very remote chance that'll affect anyone.

src/ConfigGen/Program.cs
```diff
@@ -418,26 +418,51 @@ static string createClassHeader(string name, string docs, bool derive)
...
+        ///     </para>
         /// </summary>
-        public ConsumerConfig(IEnumerable<KeyValuePair<string, string>> config) { this.properties = new Dictionary<string, string>(config.ToDictionary(a => a.Key, a => a.Value)); }
+        public {name}(IDictionary<string, string> config) : base(config) {{ }}
 
+        /// <summary>
+        ///     Initialize a new <see cref=""{name}"" /> instance copying
+        ///     an existing key/value pair collection.
+        ///     <para>
+        ///     This will make a copy of the provided values i.e. the original will be NOT modified
```
mhowlett: i think this is clearer:
'This will make a copy of the provided values. i.e. operations on this class will NOT modify the original.'

src/ConfigGen/Program.cs
```diff
@@ -418,26 +418,51 @@ static string createClassHeader(string name, string docs, bool derive)
...
+        ///     </para>
+        /// </summary>
+        public {name}(ClientConfig config) : base(config) {{ }}
 
         /// <summary>
-        ///     Initialize a new <see cref=""ConsumerConfig"" /> instance based on
+        ///     Initialize a new <see cref=""{name}"" /> instance wrapping
         ///     an existing key/value pair collection.
+        ///     <para>
+        ///     This will change the values ""in-place"" i.e. the original will be modified
```
mhowlett: Maybe:
'Note: operations on this class will modify the provided collection'

src/ConfigGen/Program.cs
```diff
@@ -418,26 +418,51 @@ static string createClassHeader(string name, string docs, bool derive)
...
 
         /// <summary>
-        ///     Initialize a new <see cref=""ConsumerConfig"" /> instance based on
+        ///     Initialize a new <see cref=""{name}"" /> instance wrapping
         ///     an existing <see cref=""ClientConfig"" /> instance.
         /// </summary>
-        public ConsumerConfig(ClientConfig config) { this.properties = new Dictionary<string, string>(config.ToDictionary(a => a.Key, a => a.Value)); }
+        /// <summary>
+        ///     <para>
+        ///     This will change the values ""in-place"" i.e. the original will be modified
```
mhowlett: Maybe:
'Note: operations on this class will modify the provided collection'

src/ConfigGen/Program.cs
mhowlett: it's the weekend and i have't checked this super carefully, but a few comments and then it's looking pretty good to me. thanks for your efforts!

src/Confluent.Kafka/Config.cs
AndyPook: We use it all over the place. But only really of use if generating MSDN style doc sites.
I'll take them out for now

src/ConfigGen/Program.cs
AndyPook: consolidated versions of the wording

src/ConfigGen/Program.cs
AndyPook: consolidated wording

src/ConfigGen/Program.cs
AndyPook: consolidated wording

src/Confluent.Kafka/Config.cs
AndyPook: xmldoc consolidated

src/ConfigGen/Program.cs
AndyPook: I think the other changes have resolved this?

mhowlett:
the increase in complexity is still bothering me a bit unfortunately - before these changes, the config class 'owns' the configuration. after the changes, sometimes it does, sometimes it doesn't, and even if we depreciate the `IEnumerable` constructor overload, that will still the case (since we definitely want the default constructor to remain). usually, I think making something more complex in this sort of way is a sign the object modeling isn't perfect. that said, API UX is ultimately the most important thing.

> Last one is handy for config validation, setting defaults etc in dotnetCore IOptions stuff

specific example? can this be achieved another way, for example with a `SetRange` or `Merge(overwriteYesNo)` method or something?

AndyPook:
"complexity"? It has always been a decorator on top of a dictionary.

In dotnet Core it is standard practice to create an "options" class to define the configuration required for a component. Something like (simplified)
```
public class KafkaOptions
{
  public Dictionary<string, string> Configuration {get;set;}
  public string TopicName {get;set;}
}
```
This is bound to config sourced form json, env vars ...
It is also common to add PostConfig and Validation steps to the Configuration system. As a simple example of a PostConfig step.
```
 public void PostConfigure(KafkaOptions o)
{
  var config = new ConsumerConfig(o.Configuration)
  if(confg.GroupId == null)
    config.GroupId = myServiceName;
}
```
"myServiceName" is pulled out of the HostEnvironment. So the developer can elide "group.id" from the appSettings as it defaults to the name of the service.
A few conventions like this and config becomes much simplified.

Without this PR, that very simple and expressive PostConfigure needs to understand dictionary.TryGet know about the "group.id" magic string etc etc rather than just be able to use a property and have the original dictionary be updated.

This is a very normal pattern I've used in many teams without confusion.
I think `SetRange` or `Merge(override)` would be far more likely to raise questions. I'm not even sure what they might do 🤔 

Maybe we could make the ctor's explicitly decorators. Remove the IEnumerable ctor and have a couple of `CopyFrom` static factory methods?

mhowlett:
librdkafka style property names aren't idiomatic here (right?). my thinking was that in practice no one should/would ever use them - it would be better/more natural in the context of c# to just use the `ConsumerConfig` property names in external configuration (but shh, don't tell @edenhill i said that). So instead of `Dictionary<string, string>`, my thinking was you'd use `ConsumerConfig`. It's compatible with the asp.net configuration stuff.

The pattern you use does work pretty well, but seems a bit odd, and i was trying to avoid odd. is there a reason it's preferable to what i suggest above?

Note: I haven't thought this through / used extensively, so there could easily be holes in my thinking.

The increased structural complexity is because currently the config classes always own state, but with these changes they can either own state, or not.

AndyPook:
We really do think about things differently...
I'm confused that you see this as "complexity" or "odd". Facades and decorators are some of the oldest patterns. I don't see "ownership" as anything to worry about in this context. There are a bunch of MS types that "own" or not some object depending on the ctor used (ie some of the Stream types).

There is a strong preference that Options classes are simple DTOs with no dependencies on types from outside parts. The classes are often used in ctors of "business" classes. These are generally kept in projects that are preferred to not have dependencies on implementation details, rather on abstractions. 
So if the options class has a property of a Config type, anything that uses it also has a dep on Kafka.

There's also the O in SOLID. Keep as much as possible Open, flexible. If it can be used as a true facade then I can use it in a useful way. If not then it can only be used in a constrained way and I have to jump through hoops and re-include a bunch of the magic strings.

Which reminds me. The ConfigGen really ought to create a static class of const string props for all the config key names. Maybe I'll do a separate PR.

mhowlett:
> We really do think about things differently...

TBH I've been trying to work out how best to think about this. Some of the issues are not things I've needed to be up close with. Sorry to be a bit slow - would prefer to be slow than make mistakes.

> I'm confused that you see this as "complexity" or "odd". Facades and decorators are some of the oldest patterns

i say 'odd' because i can't think of anything exactly analogous this in the framework.
i say 'complex' because structural complexity has increased. this is a minor point though.

> There are a bunch of MS types that "own" or not some object depending on the ctor used (ie some of the Stream types).

that's a good comparison.

 >  These are generally kept in projects that are preferred to not have dependencies on implementation details, rather on abstractions. So if the options class has a property of a Config type, anything that uses it also has a dep on Kafka.

That's a great and important point, very convincing.

> Which reminds me. The ConfigGen really ought to create a static class of const string props for all the config key names. Maybe I'll do a separate PR.

sounds good.

ok, I think i'm convinced. i'll review it closely again before merge though.

thanks for persisting!

src/ConfigGen/Program.cs
```diff
@@ -418,26 +418,44 @@ static string createClassHeader(string name, string docs, bool derive)
...
         }
 
-        static string createConsumerSpecific()
+        static string createClassConstructors(string name)
         {
-            return
-@"        /// <summary>
-        ///     Initialize a new empty <see cref=""ConsumerConfig"" /> instance.
+            var codeText = $@"
+        /// <summary>
```
mhowlett: It was probably me who caused you to put this in here, but I don't think `CopyFrom` is necessary.

AndyPook:
OK, thanks for looking into it.

I've just pushed a change that fixes the return types of the `CopyFrom` methods (oops).
However, if you are "happier" with this facade approach then it may be "better" to revert them to  just being ctors which would feel more "nature" to me.

src/ConfigGen/Program.cs
```diff
@@ -322,6 +322,33 @@ static string EnumNameToDotnetName(string enumName)
             return result;
         }
 
+        static string createConfigKeys(IEnumerable<PropertySpecification> props)
```
mhowlett: Thanks for these additions!

There's an existing class `ConfigPropertyNames` (which predates `Config`) with some dotnet specific names in it. We shouldn't have this and what you've done. It's better under `Config` as you are doing it, but I prefer `PropertyNames` over `KeyNames`. I think we should depreciate `ConfigPropertyNames` (and move to `PropertyNames` under `Config`).

But a problem with your additions is they lets you to do, e.g. `ProducerConfig.GroupId`, which isn't correct. The nested static classes in `ConfigPropertyNames` solves this.

AndyPook: I've reworked the codegen. Could you confirm that I'm using `CPorA` the right way ?

AndyPook: If we want to "deprecate" `ConfigPropertyNames`, we'd need to also rework the current uses within this project. There are 19 references. Simple enough to do.
Would you want that done in this PR? Or done in a follow-up?

AndyPook: Decided to see what it would look like.
Quite neat. Apart from inside the Consumer/Producer classes where the internal `Config` classes mean that `Config.PropertyNames` class needs to be disambiguated. I think it's ok.
Thoughts?

AndyPook: ping?

mhowlett:
sorry for the delay, this is getting big.
I had a quick look before and saw that e.g. the producer property names were accessible via the consumer config class. I don't know if i said something that implied the structure should be like that, but if i did, i didn't mean it ... not great.
btw. i'm still ok with the dictionary changes.

AndyPook:
I was just following the same style as the `ConfigPropertyNames` class. Thought it'd be neat to roll them into the gen'd classes and get rid of the extra class. It still has the structure to show which names are for which config type.
It's just a bunch of strings 😄 nothing stopping anyone from using whichever literal or const string they like.

I've re-worked it again. But if this isn't good then I think I'd rather just back out all this PropertName stuff and just leave the new ctor bits. With these there's not really any need for the consts anyway.

AndyPook:
#

AndyPook:
oops!

AndyPook:
please see #979 for a more simple pr that just contains the Config ctor changes without the KeyNames work since around May 2

AndyPook:
As #979 is now accepted...

Issue #886, opened 2019-04-15
Consumer constructor is marked internal


dbrownxc:
Description
===========
Confluent.Kafka 1.0 RC4

Consumer has one constructor that is marked as internal, making it impossible to extend.  Is this intended?  This forces a developer to wrap the class to associate data with the consumer (changing all the code from consumer.Consume() to consumerWrapper.Consumer.Consume(), for example).  

One option is to make the constructor public.  

Another option is to seal the class or mark it as internal, only allowing the interface to be used externally, making it clear that the class cannot be inherited.

How to reproduce
================

The following will not build because the constructor is marked internal.
```
public class ExtendedConsumer<TKey,TValue> : Consumer<TKey,TValue>
{
    public ExtendedConsumer(ConsumerBuilder<TKey,TValue> builder):base(builder){};
}
```

Possibly related issues:
================
https://github.com/confluentinc/confluent-kafka-dotnet/issues/151
https://github.com/confluentinc/confluent-kafka-dotnet/issues/777

mhowlett:
that's a good point, the `Producer`/`Consumer` classes should be internal. Another relevant thing here is the `AdminClient` does have a public constructor (the reasoning being a `DependentAdminClientBuilder` would provide no value). However, it would actually provide value in allowing `AdminClient` to be made internal, limiting the API surface area (including need to double up on the docs) and avoiding the confusion you point out. I think this should be changed.

I don't think the inability to extend the client classes is a downside (composition is better in all the cases i've thought of), though I'm interested in hearing the specifics of what you want to do.

dbrownxc:
We just want to add some typed metadata to the go along with the configuration.  We are always passing the same metadata around with the consumer, so we end up wrapping that metadata along with the consumer and calling that wrapper a "consumer".

mhowlett:
i can see that that would be useful. we avoided introducing the builder pattern for a long time (to try and keep things simple), but when we finally added it, it solved so many problems, it was immediately clear it was a good idea. given it's integral to client construction, it doesn't really make sense to expose the constructors (which would only be for a very limited set of use cases). It makes more sense to hide the classes altogether as you suggest. Unfortunately, this makes what you want to do more difficult.

dbrownxc:
Understood.  Thanks for the explanation!

mhowlett:
#888 actions this issue

PR #888, opened 2019-04-15
make client classes internal

mhowlett:
address #886 - an elementary error on my part which i was blind to having looked over the API now so many times.


mhowlett:
note: force-push is re-base on to 1.0.x.

PR #891, opened 2019-04-17
Fix exception

xqrzd:
Small typo

PR #895, opened 2019-04-19
Fix exception

xqrzd:
Small typo. Replaces #891


mhowlett:
thanks, will include in 1.0.1

PR #915, opened 2019-05-02
added netstandard2.0 target

mhowlett:
see #914


mhowlett:
yes, that would be ideal, though i think the risk is very small. there are other cases missed as well and also some other things to think about / look into - targeting netcoreapp2.1 to get netcoreapp only APIs for perf improvement is not ideal, particularly as there is now 2.2 and 3.0.

PR #916, opened 2019-05-06
removed redundant access properties from ConsumeResult<TKey, TValue>

enzian:
I stumbled over this rather weird piece of API when we had the issue, that our application packaged into linux containers produces `NullReference` Exceptions when consuming messages, which it did not on our Windows development environments. We quickly figured that it was actually the partition EOF message but the reason it was taking us a long time to find out what was wrong in the first place was, that the issue manifested where it should not have!

Here is our consume code before the refactoring:

``` csharp
while (!stoppingToken.IsCancellationRequested)
{
    try
    {
        var msg = consumer.Consume(stoppingToken);
        if (msg != null)
        {
            logger.LogDebug($"Received message from topic '{msg.Topic}:{msg.Partition}' with offset: '{msg.Offset}[{msg.TopicPartitionOffset}]'");

            using (var scope = serviceProvider.CreateScope())
            {
                var handler = scope.ServiceProvider.GetService<IKafkaMessageHandler<TKey, TValue>>();
                if (handler == null)
                {
                    logger.LogError("Failed to resolve message handler. Did you add it to your DI setup.");
                    continue;
                }
                try
                {
                    // Invoke the handler
                    await handler.Handle(msg);
                }
                catch (Exception e)
                {
                    logger.LogError(e, "Message handler failed", e);
                    continue;
                }
            }
        }
        else
        {
            logger.LogDebug("No messages received");
        }
    }
}
```
Seemed fine to us until we got a nullref exception deeper down in the stack of `handler.Handle(msg)`. After some debugging we found that the properties `Key`, `Value` and `Timestamp` on `ConsumeResult<,>` were just wrappers around the same properties of the object in the `Message` property. We had checks for the cases that `Key` or `Value` could be null, but we did not expect `if (msg.Key == null)` to throw a nullref. We fixed our code in [this commit](https://github.com/leuchterag/Microsoft.Extensions.Hosting.Kafka/commit/9708c95579b2aa4ab2b15193cb0e53517d9c514f) and just opted for an additional check if there is actually a message present in the `ConsumeResult`.

Now, we could not figure out why it was implemented this way but we figured that duplicating functionality like that is an antipattern and this is also the reason I submit this PR. I know, that 1.0.0 has just been released and the time for breaking API changes is actually over, but if we don't merge this PR because of the fact it introduces a breaking change by removing the redundant properties, we should at least change the code so that it does not throw `NullReferenceException`s when accessing properties...

### DIFF
--- a/examples/AvroBlogExamples/Program.cs
+++ b/examples/AvroBlogExamples/Program.cs
@@ -124,7 +124,7 @@ namespace AvroBlogExample
 
                             Console.WriteLine(
                                 consumeResult.Message.Timestamp.UtcDateTime.ToString("yyyy-MM-dd HH:mm:ss")
-                                + $": [{consumeResult.Value.Severity}] {consumeResult.Value.Message}");
+                                + $": [{consumeResult.Message?.Value.Severity}] {consumeResult.Message?.Value.Message}");
                         }
                         catch (ConsumeException e)
                         {
--- a/examples/AvroGeneric/Program.cs
+++ b/examples/AvroGeneric/Program.cs
@@ -77,7 +77,7 @@ namespace Confluent.Kafka.Examples.AvroGeneric
                             {
                                 var consumeResult = consumer.Consume(cts.Token);
 
-                                Console.WriteLine($"Key: {consumeResult.Message.Key}\nValue: {consumeResult.Value}");
+                                Console.WriteLine($"Key: {consumeResult.Message.Key}\nValue: {consumeResult.Message?.Value}");
                             }
                             catch (ConsumeException e)
                             {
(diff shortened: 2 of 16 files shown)


enzian:
[clabot:check]

mhowlett:
Thanks for point this out - we should throw a more informative exception in the case `Message` is `null` (which happens only on partition eof events, which you must explicitly enable to get). In a previous version of the library `Message` could never be null - partition eof's were delivered via a C# `event`, however it's very inconvenient if they're not inline with the messages, so we arrived at the current design. Regardless, I think we want the convenience properties you're removing (since they're convenient!). Also, we can't remove them without depreciating first since it's a breaking change. Technically changing the exception type is also a breaking change, but i think that's minor enough for us to do it.

mhowlett:
closing since we want to address this as described above. will open an issue.

mhowlett:
actually, we can just check if `Message` is `null` and return `null` in that case (don't throw any exception). 
also, we probably don't want setters on these convenience properties, just getters.

enzian:
> actually, we can just check if Message is null and return null in that case (don't throw any exception).
exactly what I meant by:

> we should at least change the code so that it does not throw NullReferenceExceptions when accessing properties...

PR #930, opened 2019-05-13
Feature/filte by header

pitming:
Add the ability to filter a message based on the header before deserializing the key and values


pitming:
[clabot:check]

mhowlett:
thanks for the contribution. a couple of comments:
1. we won't add this functionality to the `Consumer` class - we want to keep it foundational. Can I suggest implementing this functionality as an extension method? Unfortunately we can't accept that change as a PR right now either, since we are still in the process of working what higher level APIs on top of the core classes are going to look like, and we need to make sure this will fit with the plan.
2. lots of white-space here, commits shouldn't have white-space changes.

pitming:
I thx for your comment.

I'll try to be carefull about spaces next time.

I was trying to mimic this functionality from spring: https://github.com/spring-cloud/spring-cloud-stream/issues/796.
Do you have any suggestion on the extension method ? I do not really see how I can stop de process of the message from an extrenal method.
I plan to go with a kind of lazy serializer.

regards,

PR #936, opened 2019-05-14
docker scripts

mhowlett:
the docker scripts i use for testing / development.

they should possibly be in their own / another repo as they're more generally useful, however eventually the .net integration tests will be automated and depend on these, so I think that is justification enough to leave them here.

also, they could be documented better, but other priorities, and better merged with a little documentation than not merged at all.


mhowlett:
i'm now getting cold feet on basing the images off a download of the kafka .tar. It would be better to just use off-the-shelf confluent docker images rather than require the images be built. The current strategy allows for prometheus monitoring, but there's little benefit to this in a dev/test environment which is what these scripts are intended for. some history: the scripts are pretty much re-purposed from another project where they make more sense (hence the implementation). changing this will require a bit of effort.

PR #945, opened 2019-05-18
Added 'Headers' property to SerializationContext

PaulHatch:
This PR adds a `Headers` property to `SerializationContext` and populates it in the `Producer` and `Consumer` classes so that headers can be inspected during serialization and deserialization.

With the 1.0 release, `Produce` now supports assigning headers when a message is created. This property will be null by default and as such a serializer will be unable to assign headers if a headers instance was not created with the message. While this is easy to work around, it could be slightly confusing.

This could be resolved by providing a default headers instance if the message headers are null during message production, I can add this if it is agreed this is a good idea.


PaulHatch:
[clabot:check]

AndyPook:
Nice!

Might be worth adding more doc to the SyncContext.Headers prop to explain that it might be null if `enableHeaderMarshaling` is false ?

AndyPook:
... I'm thinking that it should not create an empty headers. That would change the opt-in style. Effectively there'd always be a Headers instance. Both in the Consumer and in Message.Headers. I think I've seen code that checks for `Header==null` which would now never be true 🤔 

Would it be "sane" that if you know your serdes require Headers, that you would also be responsible to configuring the P&C's appropriately such that `enableHeaderMarshalling` was true?
So, maybe the comment about more xmldoc on the new `.Headers` prop was "right" 😄

PaulHatch:
So something like this?

> This property will be null during serialization if the message being produced did not set a value for the Headers property. It will also be null during deserialization if ConsumeResultFields does not include the "headers" field.

AndyPook:
Sure, that works...
maybe something about "By default Headers will be provided. However, may be null depending on how 'dotnet.consumer.consume.result.fields' is configured." or something.
Just my 2c... Matt will have the final say.

src/Confluent.Kafka/SerializationContext.cs
```diff
@@ -37,20 +37,34 @@ public static SerializationContext Empty
         /// <param name="topic">
         ///     The topic the data is being written to or read from.
         /// </param>
-        public SerializationContext(MessageComponentType component, string topic)
+        /// <param name="header">
+        ///     The headers associated with the message being processed.
+        /// </param>
+        public SerializationContext(MessageComponentType component, string topic, Headers header)
```
mhowlett: header -> headers

mhowlett:
Thanks for the contribution. I've done a lot of work in a local branch somewhere that incorporates headers into the serialization API (including serialization / deserialization of the header values), which isn't merged due to time constraints. These changes look good, but i might wait to get back in that head space before merging in case I'm forgetting something important.

PaulHatch:
Alright, sounds good

richard-axual:
I was wondering, has there been any progress on this issue?
I'd really like access to the headers from the serializers, as this will make our serializer and deserializer solutions easier.

PaulHatch:
Would be great to get this in, either with this or another approach.

PR #956, opened 2019-05-23
Expose additional compatibility methods on Schema Registry client

dstelljes:
New methods on the client that had already been implemented on the REST service:

- `IsCompatibleAsync` (overload for specific version)
- `GetCompatibilityAsync`/`SetCompatibilityAsync`
- `GetGlobalCompatibilityAsync`/`SetGlobalCompatibilityAsync`

### DIFF
--- a/src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
+++ b/src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
@@ -44,9 +44,15 @@ namespace Confluent.SchemaRegistry
     ///      - <see cref="CachedSchemaRegistryClient.LookupSchemaAsync(string, Schema, bool)" />
     ///      - <see cref="CachedSchemaRegistryClient.GetLatestSchemaAsync(string)" />
     ///      - <see cref="CachedSchemaRegistryClient.GetAllSubjectsAsync" />
+    ///      - <see cref="CachedSchemaRegistryClient.GetCompatibilityAsync(string)" />
+    ///      - <see cref="CachedSchemaRegistryClient.GetGlobalCompatibilityAsync" />
     ///      - <see cref="CachedSchemaRegistryClient.GetSubjectVersionsAsync(string)" />
     ///      - <see cref="CachedSchemaRegistryClient.IsCompatibleAsync(string, Schema)" />
+    ///      - <see cref="CachedSchemaRegistryClient.IsCompatibleAsync(string, int, Schema)" />
     ///      - <see cref="CachedSchemaRegistryClient.IsCompatibleAsync(string, string)" />
+    ///      - <see cref="CachedSchemaRegistryClient.IsCompatibleAsync(string, int, string)" />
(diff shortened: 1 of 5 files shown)


dstelljes:
[clabot:check]

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -323,6 +323,41 @@ public async Task<bool> IsCompatibleAsync(string subject, string schema)
...
+        ///     Refer to <see cref="Confluent.SchemaRegistry.ISchemaRegistryClient.GetCompatibilityAsync(string)" />
+        /// </summary>
+        public async Task<Compatibility> GetCompatibilityAsync(string subject)
+            => await restService.GetCompatibilityAsync(subject).ConfigureAwait(continueOnCapturedContext: false);
+
+
+        /// <summary>
+        ///     Refer to <see cref="Confluent.SchemaRegistry.ISchemaRegistryClient.SetCompatibilityAsync(string, Compatibility)" />
+        /// </summary>
+        public async Task<Compatibility> SetCompatibilityAsync(string subject, Compatibility compatibility)
```
mhowlett: I think this should be `UpdateCompatibilityAsync` to match the java client and also because it more precisely describes what's going on.

I also think this shouldn't return `Compatibility` (even though this is what the java client does). I think it makes sense for the PUT api call to do this, and also for the `RestService` that backs this method to do it (since it's a direct mapping to the API), however this is a higher level abstraction, and I think returning compatibility does more to harm the API than good because people may think some other value may be returned.

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -323,6 +323,41 @@ public async Task<bool> IsCompatibleAsync(string subject, string schema)
...
+        ///     Refer to <see cref="Confluent.SchemaRegistry.ISchemaRegistryClient.GetGlobalCompatibilityAsync()" />
+        /// </summary>
+        public async Task<Compatibility> GetGlobalCompatibilityAsync()
+            => await restService.GetGlobalCompatibilityAsync().ConfigureAwait(continueOnCapturedContext: false);
+
+
+        /// <summary>
+        ///     Refer to <see cref="Confluent.SchemaRegistry.ISchemaRegistryClient.SetGlobalCompatibilityAsync(Compatibility)" />
+        /// </summary>
+        public async Task<Compatibility> SetGlobalCompatibilityAsync(Compatibility compatibility)
```
mhowlett: Same comment as above - `UpdateGlobalCompatibilityAsync`, and just return `Task`.

mhowlett:
Thanks for the PR @dstelljes - sorry it took such a long time to get to. Couple of things to change, else LGTM and would love to get it in 1.3.

dstelljes:
done, thanks!

dstelljes:
@mhowlett, anything else you want adjusted here? Just rebased onto master.

dstelljes:
Closing now that https://github.com/confluentinc/confluent-kafka-dotnet/pull/2097 has merged

Issue #939, opened 2019-05-15
ProduceAsync Locking Up


Coder3333:
Description
===========
When I call await ProduceAsync, it locks up indefinitely.  I only wait for 1 minute.  This sounds a lot like https://github.com/confluentinc/confluent-kafka-dotnet/issues/545, but I am using the 1.0.0 Confluent NuGet.

I only see this when I run from my unit tests, which is  xUnit on .Net Core 2.1.  When I run the same code from a web application, it never locks up.  Also, this does not happen 100% of the time.  If I have a test that publishes 3 times synchronously, it could lock up on any one of the 3, or none at all.  Oddly, it seems to happen less often when I run the test in Debug mode with VS, as opposed to when I run without debugging.

I am not getting any logs or errors from the producer during this lock up period.

How to reproduce
================

Call await ProduceAsync from an xUnit test in .Net Core.

mhowlett:
My guess is tests are running concurrently and you're using a SyncrhonizationContext with a small number of threads. This could easily deadlock. You could try using `.ConfigureAwait(false)` to not enforce that tasks be completed on the same synchronization context. I don't know too much about xUnit, but the following link is consistent with the above theory: https://stackoverflow.com/questions/50918647/why-does-this-xunit-test-deadlock-on-a-single-cpu-vm

Coder3333:
I thought the xunit deadlock solution sounded good, but I tried it and it has not had any affect.

I have written a ton of async framework code and I have never heard of ConfigureAwait until I started researching this ProduceAsync problem.  I tried adding it in a few places, but I think it will be tough to get them all.  It did not help when I just added it in a few spots.

mhowlett:
a WPF SynchronizationContext is single threaded, guaranteeing a deadlock.

levdimov:
We have the same issue after upgrading to 1.0.0 from 0.11.4: `Confluent.Kafka.Producer` async code doesn't have any `ConfigureAwait(false)` statements (ex: https://github.com/confluentinc/confluent-kafka-dotnet/blob/91d71ec8487aaa4afdb26b516cd93b076942dbd7/src/Confluent.Kafka/Producer.cs#L679). Before 1.0.0 Producer code was returning tasks rather than awaiting them inside implementation.

[...]

PR #967, opened 2019-05-30
Correct task awaiting with ConfigureAwait(false) in Producer.

levdimov:
This fix resolves an issue when client code waits ProduceAsync synchronously (using `.Result` or `.GetAwaiter().GetResult()`). This fix is very useful for keeping backward compatability with legacy code which you cannot easily migrate to async. As I wrote [here](https://github.com/confluentinc/confluent-kafka-dotnet/issues/939#issuecomment-497264799) before 1.0.0 Producer code was returning tasks rather than awaiting them inside implementation.

I also added ConfigureAwaitChecker.Analyzer package so you will have warnings in future


levdimov:
[clabot:check]

mhowlett:
looks good except for unnecessary white space changes (these make it difficult to track what changes happened when). also, I'm not keen on the `ConfigureAwaitChecker.Analyzer` dependency. I don't know why it would be a problem, but as a rule of thumb, prefer to not to have dependencies unless critical. if you make those changes, LGTM. Thanks!

levdimov:
Sorry about whitespace changes, my IDE is removing trailing whitespaces automatically and I hide all whitespace changes in Git UI :)

Also I've removed `ConfigureAwaitChecker.Analyzer`. Personally, I don't see any problems with private dependencies, but you are the boss :)

Thank you!

mhowlett:
thanks! I also don't see any problem, but I don't have time to give it any thought and not including it is definitely 100% safe.

levdimov:
@mhowlett, may I ask you when these changes are going to be released?

mhowlett:
I've opened #968 so we can release asap - so hopefully soon, but there is some internal debate as to how to best do rapid releases moving forward. this is a test case for that.

PR #968, opened 2019-05-31
-> v1.0.1.1

mhowlett:
@edenhill - i realize this numbering scheme for you is probably like looking at a font with the kerning off or something, but it's supported by nuget and there is precedent with other libraries, so let's give it a go.

the 1.0.1 implies librdkafka 1.0.1, allowing people easily refer to the correct librdkafka docs / release notes etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+# 1.0.1.1
+
+## Changes
+
+- Applied `ConfigureAwait(false)` to all internal `await`ed calls, which resolves deadlock issues in
+synchronization contexts with limited numbers of threads.
+
+
 # 1.0.1
 
 ## Enhancements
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# 1.0.1.1
+
+## Changes
+
+- Applied `ConfigureAwait(false)` to all internal `await`ed calls, which resolves deadlock issues in
+synchronization contexts with limited numbers of threads.
```
edenhill: Want to reference an issue here?

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# 1.0.1.1
```
edenhill: Why is there an IP address here?

Issue #993, opened 2019-06-27
CachedSchemaRegistry getting 404 response codes


jonathansant:
Description
===========

I was playing around with the following example: https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/examples/AvroSpecific/Program.cs using our schema registry.

I started getting a 404 from the schema registry.

How to reproduce
================

- Set `SchemaRegistryUrl` to `https://xxx.dev/schema-registry`
- Run the program
- Examine the requests using Fiddler:
-- Expected: https://xxx.dev/schema-registry/subjects/sucrose-external-value/versions
-- Actual: https://xxx.dev/subjects/sucrose-external-value/versions

![screenshot](https://user-images.githubusercontent.com/6423827/60269534-ea2d8180-98ee-11e9-8f7e-b32d47d1f0a8.png)

I tracked down the issue and it seems that there's some weird functionality when using `HttpClient`. Basically when adding prepending the `/` here: https://github.com/confluentinc/confluent-kafka-dotnet/blob/98edaaf8a746d1a38e85418933da34c2e3a220c2/src/Confluent.SchemaRegistry/Rest/RestService.cs#L234

causes `HttpClient` to discard part of the url.

Checklist
=========

Please provide the following information:
 - [ ] A complete (i.e. we can run it), minimal program demonstrating the problem. No need to supply a project file.
 - [ ] Confluent.Kafka nuget version.
 - [ ] Apache Kafka version.
 - [ ] Client configuration.
 - [ ] Operating system.
 - [ ] Provide logs (with "debug" : "..." as necessary in configuration).
 - [ ] Provide broker log excerpts.
 - [x] Critical issue.

jonathansant:
Any news on this, please?

mhowlett:
thanks for the PR! closing this issue in favor of that.

PR #996, opened 2019-06-28
fix(Schema Registry): append '/' to `RestService.BaseAddress` and remove '/' from subject requests

jonathansant:
fixes: https://github.com/confluentinc/confluent-kafka-dotnet/issues/993

### DIFF
--- a/src/Confluent.SchemaRegistry/Rest/RestService.cs
+++ b/src/Confluent.SchemaRegistry/Rest/RestService.cs
@@ -22,7 +22,6 @@ using System.Linq;
 using System.Net;
 using System.Net.Http;
 using System.Text;
-using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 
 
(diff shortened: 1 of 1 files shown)


jonathansant:
[clabot:check]

src/Confluent.SchemaRegistry/Rest/RestService.cs
```diff
@@ -52,23 +53,29 @@ internal class RestService : IRestService
...
+                    if (authorizationHeader != null) { client.DefaultRequestHeaders.Authorization = authorizationHeader; }
+                    return client;
+                })
                 .ToList();
         }
 
+        private static string SanitizeUri(string uri)
+        {
+            var sanitized = uri.StartsWith("http", StringComparison.Ordinal) ? uri : $"http://{uri}";
+            return $"{Regex.Replace(sanitized, "(/$)+", string.Empty)}/";
```
mhowlett: can you change this to not use `System.Text.RegularExpressions`? i think this might result in an assembly being loaded that otherwise wouldn't be, and the logic's simple enough that using a regex doesn't add much.

mhowlett:
thanks! I think a number of people have hit this issue.
sorry for the delay in reviewing (end of quarter madness). 
lgtm except for 1 minor change request.

src/Confluent.SchemaRegistry/Rest/RestService.cs
jonathansant: done

PR #1037, opened 2019-08-23
regen config classes + version -> 1.2.0-RC2

mhowlett:


src/Confluent.Kafka/Config_gen.cs
```diff
@@ -1,4 +1,4 @@
-// *** Auto-generated from librdkafka v1.0.0 *** - do not modify manually.
+// *** Auto-generated from librdkafka master *** - do not modify manually.
```
edenhill: nit: Would prefer a proper tag / sha1 here.

src/Confluent.Kafka/Config_gen.cs
```diff
@@ -659,23 +723,23 @@ public class ClientConfig : Config
         public string SaslKerberosPrincipal { get { return Get("sasl.kerberos.principal"); } set { this.SetObject("sasl.kerberos.principal", value); } }
 
         /// <summary>
-        ///     Full kerberos kinit command string, %{config.prop.name} is replaced by corresponding config object value, %{broker.name} returns the broker's hostname.
+        ///     Shell command to refresh or acquire the client's Kerberos ticket. This command is executed on client creation and every sasl.kerberos.min.time.before.relogin (0=disable). %{config.prop.name} is replaced by corresponding config object value.
```
edenhill: This stuff was released in v1.1.0.
Maybe update your RELEASE.md guidelines to include the re-gen step so it is not forgotten in the future.

src/Confluent.Kafka/Config_gen.cs
```diff
@@ -659,23 +723,23 @@ public class ClientConfig : Config
...
         ///
-        ///     default: kinit -S "%{sasl.kerberos.service.name}/%{broker.name}" -k -t "%{sasl.kerberos.keytab}" %{sasl.kerberos.principal}
+        ///     default: kinit -R -t "%{sasl.kerberos.keytab}" -k %{sasl.kerberos.principal} \|\| kinit -t "%{sasl.kerberos.keytab}" -k %{sasl.kerberos.principal}
```
edenhill: Remove escaping backslashes.
Is this stuff generated from CONFIGURATION.md and not rd_kafka_conf_properties_show() ?

mhowlett: yes, it's from Configuration.md

src/Confluent.Kafka/Config_gen.cs
mhowlett: can't update, don't have

src/Confluent.Kafka/Config_gen.cs
mhowlett: v. good idea re: sha1. ideally we want the proper tag of the actual librdkafka release, but that doesn't exist yet so it would result in just an extra step that isn't worth it just for a comment.

src/Confluent.Kafka/Config_gen.cs
edenhill: Highly recommend we make one, for when we need to make a release when you're in Mexico.

mhowlett:
can you approve so I can test?

src/Confluent.Kafka/Config_gen.cs
mhowlett: for additional context: the config regen never made it into 1.1 because of the dependency on https://github.com/edenhill/librdkafka/pull/2416 (which i didn't pick up on before the 1.1 librdkafka release)

PR #1070, opened 2019-10-03
Added support for subject name strategies

mhowlett:


src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
```diff
@@ -85,6 +111,16 @@ public static class PropertyNames
             ///     default: "" (no authentication).
             /// </summary>
             public const string SchemaRegistryBasicAuthUserInfo = "schema.registry.basic.auth.user.info";
+
+            /// <summary>
+            ///     Key subject name strategy.
+            /// </summary>
+            public const string SchemaRegistryKeySubjectNameStrategy = "schema.registry.key.subject.name.strategy";
```
mhowlett: The `SchemaRegistry` prefix on property names here, and in the config class itself are pretty annoying. I'm thinking we should depreciate them and add variants without this prefix (however the naming of the underlying property name `schema.registry.key.subject.name.strategy` should remain). There's plenty of color to add to all this, available verbally to any reviewer on request...

test/Confluent.SchemaRegistry.Serdes.IntegrationTests/Tests/ProduceConsume.cs
```diff
@@ -118,7 +115,63 @@ public static void ProduceConsume(string bootstrapServers, string schemaRegistry
...
+                }
+            }
         }
 
+        /// <summary>
+        ///     Test that messages produced with the Avro serializer can be consumed with the
+        ///     Avro deserializer (topic name strategy)
+        /// </summary>
+        [Theory, MemberData(nameof(TestParameters))]
+        private static void ProduceConsume_Topic(string bootstrapServers, string schemaRegistryServers)
```
mhowlett: I'm sure it's possible in xunit to parameterize a single test rather than have these 3 variants, but i was too lazy to look it up, and xunit is super inconvenient for integration tests anyway, and at some point i want to get rid of it.

src/Confluent.SchemaRegistry/ISchemaRegistryClient.cs
```diff
@@ -150,10 +150,14 @@ public interface ISchemaRegistryClient : IDisposable
         /// <param name="topic">
         ///     The topic name.
         /// </param>
+        /// <param name="recordType">
+        ///     The fully qualified avro record type or null if the key is not
```
edenhill: Avro, not avro

src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
edenhill: We need something to separate the client and SR config namespaces, right?

test/Confluent.SchemaRegistry.IntegrationTests/Tests/IsCompatible.cs
```diff
@@ -24,25 +23,25 @@ namespace Confluent.SchemaRegistry.IntegrationTests
...
         {
-            var topicName = Guid.NewGuid().ToString();
+            var sr = new CachedSchemaRegistryClient(new SchemaRegistryConfig { SchemaRegistryUrl = config.Server });
 
             var testSchema1 = 
                 "{\"type\":\"record\",\"name\":\"User\",\"namespace\":\"Confluent.Kafka.Examples.AvroSpecific" +
                 "\",\"fields\":[{\"name\":\"name\",\"type\":\"string\"},{\"name\":\"favorite_number\",\"type\":[\"i" +
                 "nt\",\"null\"]},{\"name\":\"favorite_color\",\"type\":[\"string\",\"null\"]}]}";
 
-            var sr = new CachedSchemaRegistryClient(new SchemaRegistryConfig { SchemaRegistryUrl = config.Server });
```
edenhill: This changes an existing test, possibly allowing for regression to slip through.
Instead extend or duplicate the test to cover both the old and new test cases.

test/Confluent.SchemaRegistry.Serdes.IntegrationTests/Tests/ProduceConsume.cs
edenhill: I'm all for test duplication, it is much better than the alternative.

test/Confluent.SchemaRegistry.Serdes.IntegrationTests/Tests/ProduceConsumeGeneric.cs
```diff
@@ -52,48 +46,54 @@ public static void ProduceConsumeGeneric(string bootstrapServers, string schemaR
             );
 
             var config = new ProducerConfig { BootstrapServers = bootstrapServers };
-            var schemaRegistryConfig = new SchemaRegistryConfig { SchemaRegistryUrl = schemaRegistryServers };
+            var schemaRegistryConfig = new SchemaRegistryConfig
+            {
+                SchemaRegistryUrl = schemaRegistryServers,
+                // Test SchemaRegistryKeySubjectNameStrategy here,
+                // and SchemaRegistryValueSubjectNameStrategy in ProduceConsume.
+                SchemaRegistryKeySubjectNameStrategy = nameStrategy
```
edenhill: This reduces test coverage, would rather like to see an additional test that explicitly sets the parameter.

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -121,6 +133,21 @@ public CachedSchemaRegistryClient(IEnumerable<KeyValuePair<string, string>> conf
                 throw new ArgumentException($"Invalid value '{basicAuthSource}' specified for property '{SchemaRegistryConfig.PropertyNames.SchemaRegistryBasicAuthCredentialsSource}'");
             }
 
+            var keySubjectNameStrategyString = Convert.ToString(config.FirstOrDefault(prop => prop.Key.ToLower() == SchemaRegistryConfig.PropertyNames.SchemaRegistryKeySubjectNameStrategy).Value) ?? "";
```
rigelbm: config is a IEnumerable<KeyValuePair<string, string>>. I don't think this Convert.ToString is needed anymore.

test/Confluent.SchemaRegistry.Serdes.IntegrationTests/Tests/ProduceInvalid.cs
```diff
@@ -0,0 +1,153 @@
...
+// Refer to LICENSE for more information.
+
+using System;
+using Confluent.Kafka;
+using Xunit;
+
+
+namespace Confluent.SchemaRegistry.Serdes.IntegrationTests
+{
+    public static partial class Tests
```
rigelbm: Can you split each invalid case in a separate test?

test/Confluent.SchemaRegistry.UnitTests/CachedSchemaRegistryClient.cs
```diff
@@ -37,21 +36,75 @@ public void NoUrls()
...
-        public void ConstructKeySubjectName()
+        public void InvalidSubjectNameStrategy()
         {
             var config = new SchemaRegistryConfig { SchemaRegistryUrl = "irrelevanthost:8081" };
+            config.Set(SchemaRegistryConfig.PropertyNames.SchemaRegistryKeySubjectNameStrategy, "bad_value");
+            Assert.Throws<ArgumentException>(() => new CachedSchemaRegistryClient(config));
+        }
 
+        [Fact]
+        public void ConstructKeySubjectName()
```
rigelbm: Can you split each case a in a separate test?

test/Confluent.SchemaRegistry.UnitTests/CachedSchemaRegistryClient.cs
```diff
@@ -37,21 +36,75 @@ public void NoUrls()
...
         }
 
         [Fact]
         public void ConstructValueSubjectName()
         {
             var config = new SchemaRegistryConfig { SchemaRegistryUrl = "irrelevanthost:8081" };
-
             CachedSchemaRegistryClient src = new CachedSchemaRegistryClient(config);
-            Assert.Equal("mytopic-value", src.ConstructValueSubjectName("mytopic"));
+            Assert.Equal("mytopic-value", src.ConstructValueSubjectName("mytopic", "myschemaname"));
```
rigelbm: Can you split each case in a separate test?

test/Confluent.SchemaRegistry.Serdes.IntegrationTests/Tests/ProduceConsume.cs
rigelbm: Duplication is good. :)

src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
mhowlett: Only for configuration specified in a flat configuration file, which keeping the `schema.registry.key.subject.name.strategy` names enables. This is being overly accommodating though because no one's going to do that in practice in C#. Having `SchemaRegistry` in the property names as well as the enum (which always needs to be specified) is just unnecessarily doubling up.

I'll possibly open a follow up PR with the above change.

test/Confluent.SchemaRegistry.Serdes.IntegrationTests/Tests/ProduceConsume.cs
mhowlett: I mean there's almost certainly a short hand way of driving xunit to have exactly the same test coverage as this, but i'm too lazy to look in the manual.

src/Confluent.SchemaRegistry/ISchemaRegistryClient.cs
mhowlett: some would argue it's not worthy...

mhowlett: (did i say that out loud?)

test/Confluent.SchemaRegistry.Serdes.IntegrationTests/Tests/ProduceConsumeGeneric.cs
mhowlett: it also extends test coverage (to previously applicable cases that weren't covered). i'm making an effort / payoff tradeoff decision here.

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -132,7 +159,9 @@ public CachedSchemaRegistryClient(IEnumerable<KeyValuePair<string, string>> conf
                     property.Key != SchemaRegistryConfig.PropertyNames.SchemaRegistryRequestTimeoutMs &&
                     property.Key != SchemaRegistryConfig.PropertyNames.SchemaRegistryMaxCachedSchemas &&
                     property.Key != SchemaRegistryConfig.PropertyNames.SchemaRegistryBasicAuthCredentialsSource &&
-                    property.Key != SchemaRegistryConfig.PropertyNames.SchemaRegistryBasicAuthUserInfo)
+                    property.Key != SchemaRegistryConfig.PropertyNames.SchemaRegistryBasicAuthUserInfo &&
+                    property.Key != SchemaRegistryConfig.PropertyNames.SchemaRegistryKeySubjectNameStrategy &&
+                    property.Key != SchemaRegistryConfig.PropertyNames.SchemaRegistryValueSubjectNameStrategy)
```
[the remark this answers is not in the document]
mhowlett: yeah, it's getting a bit ridiculous. not sure i see the value in a collection though (wouldn't result in net code reduction).

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -121,6 +133,21 @@ public CachedSchemaRegistryClient(IEnumerable<KeyValuePair<string, string>> conf
...
+                !Enum.TryParse<SubjectNameStrategy>(keySubjectNameStrategyString, out KeySubjectNameStrategy))
+            {
+                throw new ArgumentException($"Unknown KeySubjectNameStrategy: {keySubjectNameStrategyString}");
+            }
+
+            if (valueSubjectNameStrategyString != "" &&
+                !Enum.TryParse<SubjectNameStrategy>(valueSubjectNameStrategyString, out ValueSubjectNameStrategy))
+            {
+                throw new ArgumentException($"Unknown ValueSubjectNameStrategy: {valueSubjectNameStrategyString}");
+            }
```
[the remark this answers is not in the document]
mhowlett: yep, makes sense.

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
mhowlett: removed elsewhere as well.

PR #1073, opened 2019-10-07
Depreciate SchemaRegistry property name prefix

mhowlett:
as noted in #1070, the `SchemaRegistry` prefixes on the config property names are redundant / pointless. Retaining them in the underlying property name strings (and their corresponding constants) though.

### DIFF
--- a/src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
+++ b/src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
@@ -126,7 +126,7 @@ namespace Confluent.SchemaRegistry
         /// <summary>
         ///     Specifies the configuration property(ies) that provide the basic authentication credentials.
         /// </summary>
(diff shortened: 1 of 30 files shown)


src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
```diff
@@ -145,54 +145,115 @@ public static class PropertyNames
             }
         }
 
+        /// <summary>
+        ///     Specifies the configuration property(ies) that provide the basic authentication credentials.
+        /// </summary>
+        [Obsolete("This property will be depreciated in a future version of this library in favor of the BasicAuthCredentialsSource property")]
```
rigelbm: Here and everywhere else: did you mean "deprecated"?

mhowlett: i did indeed

mhowlett: actually, I mean removed...

Issue #864, opened 2019-04-03
librdkafka segfault when topic == null


mhowlett:
`rd_kafka_topic_partition_list_add` segfaults when the topic parameter value is NULL. this is appropriate behavior in C but in C# we should never segfault. Instead, we should throw a `NullPointerException`. To achieve this we'll need to validate the parameters of `GetCTopicPartitionList`.

This is also very likely an issue in other places in the code. We should audit the entire API for this problem and fix.

callenWrench:
When we make this check (for null topic), is there any reason we shouldn't also check for an all-whitespace string? At least on non-critical performance paths.

PSanetra:
@mhowlett This issue still exists for the Methode `SafeKafkaHandle.Subscribe`

mhowlett:
thanks @PSanetra - this was resolved by https://github.com/confluentinc/confluent-kafka-dotnet/pull/1463 which is in the 1.6.x branch.

pranavrth:
The issue is fixed. Closing the issue.

PR #1078, opened 2019-10-10
#864: Added Null Reference Checks to SafeKafkaHandle

johnsoncore:
Partially addresses #864.

I've audited the AdminClient and ProducerClient for segfaults and replaced them with null reference exceptions. I placed the unit tests in their own test-class for null reference checks.

### DIFF
--- a/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
+++ b/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
@@ -236,6 +236,10 @@ namespace Confluent.Kafka.Impl
         internal int AddBrokers(string brokers)
         {
             ThrowIfHandleClosed();
+            if (brokers == null)
+            {
+                throw new NullReferenceException("List of brokers is required.");
+            }
             return (int)Librdkafka.brokers_add(handle, brokers);
         }
 
@@ -645,6 +649,12 @@ namespace Confluent.Kafka.Impl
                 }
                 foreach (var partition in partitions)
                 {
+                    if (partition.Topic == null)
+                    {
+                        Librdkafka.topic_partition_list_destroy(list);
+                        throw new NullReferenceException("Topic is required to assign partitions.");
+                    }
+
                     IntPtr ptr = Librdkafka.topic_partition_list_add(list, partition.Topic, partition.Partition);
                     Marshal.WriteInt64(
                         ptr,
@@ -1229,6 +1239,11 @@ namespace Confluent.Kafka.Impl
                 var increaseTo = newPartitionsForTopic.IncreaseTo;
                 var assignments = newPartitionsForTopic.ReplicaAssignments;
 
+                if (newPartitionsForTopic.Topic == null)
+                {
+                    throw new NullReferenceException("Cannot assign partition to null topic.");
+                }
+
                 IntPtr ptr = Librdkafka.NewPartitions_new(topic, (UIntPtr)increaseTo, errorStringBuilder, (UIntPtr)errorStringBuilder.Capacity);
                 if (ptr == IntPtr.Zero)
                 {
@@ -1282,11 +1297,16 @@ namespace Confluent.Kafka.Impl
             setOption_RequestTimeout(optionsPtr, options.RequestTimeout);
             setOption_OperationTimeout(optionsPtr, options.OperationTimeout);
             setOption_completionSource(optionsPtr, completionSourcePtr);
-
-            IntPtr[] deleteTopicsPtrs = new IntPtr[deleteTopics.Count()];
+            
+            IntPtr[] deleteTopicsPtrs = new IntPtr[deleteTopics.Count(t => t != null)];
             int idx = 0;
             foreach (var deleteTopic in deleteTopics)
             {
+                if (deleteTopic == null)
+                {
+                    continue;
+                }
+
                 var deleteTopicPtr = Librdkafka.DeleteTopic_new(deleteTopic);
                 deleteTopicsPtrs[idx] = deleteTopicPtr;
                 idx += 1;
--- a/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
+++ b/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
@@ -1297,7 +1297,7 @@ namespace Confluent.Kafka.Impl
             setOption_RequestTimeout(optionsPtr, options.RequestTimeout);
             setOption_OperationTimeout(optionsPtr, options.OperationTimeout);
             setOption_completionSource(optionsPtr, completionSourcePtr);
-            
+
             IntPtr[] deleteTopicsPtrs = new IntPtr[deleteTopics.Count(t => t != null)];
             int idx = 0;
             foreach (var deleteTopic in deleteTopics)
--- a/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
+++ b/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
@@ -238,7 +238,7 @@ namespace Confluent.Kafka.Impl
             ThrowIfHandleClosed();
             if (brokers == null)
             {
-                throw new NullReferenceException("List of brokers is required.");
+                throw new ArgumentNullException("brokers", "Argument 'brokers' must not be null.");
             }
             return (int)Librdkafka.brokers_add(handle, brokers);
         }
@@ -652,7 +652,7 @@ namespace Confluent.Kafka.Impl
                     if (partition.Topic == null)
                     {
                         Librdkafka.topic_partition_list_destroy(list);
-                        throw new NullReferenceException("Topic is required to assign partitions.");
+                        throw new ArgumentException("Cannot assign partitions because one or more have a null topic.");
                     }
 
                     IntPtr ptr = Librdkafka.topic_partition_list_add(list, partition.Topic, partition.Partition);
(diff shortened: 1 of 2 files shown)


src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -1282,11 +1297,16 @@ private void setOption_completionSource(IntPtr optionsPtr, IntPtr completionSour
             setOption_RequestTimeout(optionsPtr, options.RequestTimeout);
             setOption_OperationTimeout(optionsPtr, options.OperationTimeout);
             setOption_completionSource(optionsPtr, completionSourcePtr);
-
-            IntPtr[] deleteTopicsPtrs = new IntPtr[deleteTopics.Count()];
+
```
edenhill: involuntary (hopefully) whitespace change

test/Confluent.Kafka.IntegrationTests/Tests/AdminClient_NullReferenceChecks.cs
```diff
@@ -0,0 +1,123 @@
+// Copyright 2016-2017 Confluent Inc.
```
edenhill: 2019

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -236,6 +236,10 @@ internal int Flush(int millisecondsTimeout)
         internal int AddBrokers(string brokers)
         {
             ThrowIfHandleClosed();
+            if (brokers == null)
+            {
+                throw new NullReferenceException("List of brokers is required.");
```
mhowlett: `ArgumentNullException` would be better here and maybe text: "Argument 'brokers' must not be null". I'm thinking list is 'ambiguous' as is 'required'.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -1229,6 +1239,11 @@ private void setOption_completionSource(IntPtr optionsPtr, IntPtr completionSour
                 var increaseTo = newPartitionsForTopic.IncreaseTo;
                 var assignments = newPartitionsForTopic.ReplicaAssignments;
 
+                if (newPartitionsForTopic.Topic == null)
+                {
+                    throw new NullReferenceException("Cannot assign partition to null topic.");
```
mhowlett: Again, `ArgumentException` and maybe for the text "Cannot add partitions to a null topic." (word assign isn't appropriate).

this also leaks memory. i think we need a finally block like this:

```
finally {
    foreach (var newPartitionPtr in newPartitionsPtrs)
    {
        if (newPartitionPtr != IntPtr.Zero) {
            Librdkafka.NewPartitions_destroy(newPartitionPtr);
        }
    }
}
```

the current code is not really ideal because it too can leak memory, but this happens only in the event of an OOM error, so everything is screwed anyway.

what's different here is we leak memory on invalid args.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -1283,10 +1298,15 @@ private void setOption_completionSource(IntPtr optionsPtr, IntPtr completionSour
             setOption_OperationTimeout(optionsPtr, options.OperationTimeout);
             setOption_completionSource(optionsPtr, completionSourcePtr);
 
-            IntPtr[] deleteTopicsPtrs = new IntPtr[deleteTopics.Count()];
+            IntPtr[] deleteTopicsPtrs = new IntPtr[deleteTopics.Count(t => t != null)];
             int idx = 0;
             foreach (var deleteTopic in deleteTopics)
             {
+                if (deleteTopic == null)
```
mhowlett: Failing the entire operation with an exception rather than silently ignoring the problem seems better to me, because if the user managed to get a null in here, it's more likely than not they're screwed something up that makes the entire operation flawed. Suggest `ArgumentException` + "Cannot delete topics because one or more topic was specified as null."

johnsoncore: My original thought was to keep in line with the documentation stating this api is not transactional (some can fail, some can succeed), but your comment about it likely indicating other issues makes sense.  I'll incorporate these fixes tonight.

test/Confluent.Kafka.IntegrationTests/Tests/AdminClient_NullReferenceChecks.cs
```diff
@@ -0,0 +1,123 @@
...
+            // test creating a null topic throws a related exception
+            using (var producer = new ProducerBuilder<Null, Null>(new ProducerConfig { BootstrapServers = bootstrapServers }).Build())
+            using (var adminClient = new DependentAdminClientBuilder(producer.Handle).Build())
+            {
+                try
+                {
+                    adminClient.CreateTopicsAsync(new TopicSpecification[] { new TopicSpecification { Name = nullTopic, NumPartitions = 1, ReplicationFactor = 1 } }).Wait();
+                    Assert.True(false, "Expected exception.");
+                }
+                catch (KafkaException ex)
```
mhowlett: This is now inconsistent with `CreatePartitionsAsync`... I think we should change `CreateTopicsAsync` to explicitly check for null topics and throw `ArgumentException` not `KafkaException`. This is a breaking change, but minor enough that I think we should do it.

Feel free to audit the rest of the api for `KafkaException` where `ArgumentException` or derivative more appropriate. I think this approach to exceptions is best - not using `KafkaException` for argument related exceptions, or perhaps other very standard scenarios of which I can't think of any at the moment. But use it for everything else. Welcome feedback on that.

test/Confluent.Kafka.IntegrationTests/Tests/AdminClient_NullReferenceChecks.cs
```diff
@@ -0,0 +1,123 @@
...
+            {
+                adminClient.DeleteTopicsAsync(new List<string> { topicName1, nullTopic });
+                Assert.True(true);
+            }
+
+            // Asserts existing ListGroup behavior - null group is valid option
+            using (var producer = new ProducerBuilder<Null, Null>(new ProducerConfig { BootstrapServers = bootstrapServers }).Build())
+            using (var adminClient = new DependentAdminClientBuilder(producer.Handle).Build())
+            {
+                adminClient.ListGroup(null, TimeSpan.FromSeconds(10));
```
mhowlett: I think we should change this to throw an `ArgumentNullException`

test/Confluent.Kafka.IntegrationTests/Tests/AdminClient_NullReferenceChecks.cs
johnsoncore: Probably common knowledge, but I was curious, so I tracked down the reason for the different exception types. The root cause is that the CreatePartitions with null topic causes a segfault (so I had wrapped it in an exception.  CreateTopic on the other hand just returns a null pointer if the topic is null, so we were already wrapping it in a Kafka exception.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
johnsoncore: I've implemented this pattern on the methods touched by this pr. 

Glancing through a couple other methods (like DescribeConfigs), it looks like we might need to do a review for other areas needing this kind of disposal strategy as well.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
johnsoncore: Implemented as recommended.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -645,6 +649,12 @@ internal void Assign(IEnumerable<TopicPartitionOffset> partitions)
                 }
                 foreach (var partition in partitions)
                 {
+                    if (partition.Topic == null)
+                    {
+                        Librdkafka.topic_partition_list_destroy(list);
+                        throw new NullReferenceException("Topic is required to assign partitions.");
```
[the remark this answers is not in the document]
johnsoncore: Implemented as recommended. Uses ArgumentNullException where argument itself was null, and ArgumentException where a property of an argument was null.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
johnsoncore: I've implemented this strategy in most of the methods touched by this pull request.  

I saw one or two other methods, like DescribeConfigs, that I believe could also use this disposal strategy.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
johnsoncore: Implemented. Now throws argument exception.

test/Confluent.Kafka.IntegrationTests/Tests/AdminClient_NullReferenceChecks.cs
johnsoncore: Implemented as ArgumentException for both, with a test case to help keep the exceptions consistent between those two 'Create...' methods.

test/Confluent.Kafka.IntegrationTests/Tests/AdminClient_NullReferenceChecks.cs
johnsoncore: Implemented.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -236,6 +236,10 @@ internal int Flush(int millisecondsTimeout)
         internal int AddBrokers(string brokers)
         {
             ThrowIfHandleClosed();
+            if (brokers == null)
+            {
+                throw new ArgumentNullException("brokers", "Argument 'brokers' must not be null.");
```
mhowlett: interesting, i didn't realize this constructor overload existed. now there's redundant info in the exception, so something here is probably not ideal, but i don't think it's a big deal, so this is fine.

mhowlett:
thanks for doing this - great to get all this detail right. LGTM now.

these block level change diffs sure are hard to read!

PR #1083, opened 2019-10-10
Add a new config GetDouble function. Fix LingerMs

lorienpatt:
Added a  new GetDouble function and replaced GetEnum on LingerMs getter.
GetEnum throws an exception of ArgumentException when it tries System.Enum.Parse(typeof(double), "3.5", ignoreCase: true); ("3.5" is just used as an example).

### DIFF
--- a/src/ConfigGen/Program.cs
+++ b/src/ConfigGen/Program.cs
@@ -352,6 +352,9 @@ namespace Confluent.Kafka
                     case "bool":
                         codeText += $"GetBool(\"{prop.Name}\")";
                         break;
+                    case "double":
+                        codeText += $"GetDouble(\"{prop.Name}\")";
+                        break;
                     default:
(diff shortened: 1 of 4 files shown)


mhowlett:
thanks for the contribution @lorienpatt 

some white space issues (indentation is different in your changes to the rest of the code).

also we should make sure we get the Config_gen.cs class by running ConfigGen against librdkafka v1.2.1 tag.

feel free to fix these up, or i'll do this tomorrow. 

also this is important enough that we should also target the 1.2.x branch (not master) and I'll do a 1.2.2 release.

lorienpatt:
I have fixed the base, and the whitespace.  I will leave you to get the Config_gen.cs class by running ConfigGen against librdkafka v1.2.1 tag.

lorienpatt:
I had actually modified the Config_gen.cs class, so I have now modified the ConfigGen project.  @mhowlett  can you please check that I updated all the places I need to to add the double type?

PR #1102, opened 2019-10-30
WIP: #904 - DescribeConfig(ConsumerGroup) Segfaults

caj-johnson:
One of the member's of our team found this one last week. It looks like there's a segfault when calling AdminClient.DescribeConfig for ResourceType.Any or ResourceType.Group.   I couldn't find a root cause in the dotnet-library. 

I wanted to get input on how we wanted to handle this, so I've created an integration test to reproduce the issue and marked this PR as a work-in-progress.


caj-johnson:
[clabot:check]

mhowlett:
thanks for the PR. only broker and topic resources are supported by describe configs (kafka constraint, not specific to this client). we should fix this to throw an exception, not segfault.

Issue #1106, opened 2019-11-06
specify partition for one consumer in a group can not rebalance the others


wangjia184:
I have a topic with 150 partitions [0 - 149]
And I have three consumers joinning the same group. 

By default, each consumer is assigned with their own ranges.
* Consumer A : [0-49]
* Consumer B : [50-99]
* Consumer C : [100-149]

I want consumer B only consumes a specific partition. Hence within its `SetPartitionsRevokedHandler`,  only a specific one is returned.
```csharp
IEnumerable<TopicPartitionOffset> OnPartitionsAssigned(IConsumer<MessageKey, byte[]> consumer, List<TopicPartition> partitions)
 {
            return new TopicPartitionOffset[]
            {
                new TopicPartitionOffset(  new TopicPartition(_topic, 0), Offset.Unset),
            };
}
```

I expect consumer A and C would be assigned with partition [1 - 149]
I tested but it does not work in that way.   Consumer A & C are not rebalanced and they are still assigned with 50 partitions each.
* Consumer A : [0-49]
* Consumer C : [100-149]

Is this a bug of client?

Confluent.Kafka nuget version : 1.2.1
- Apache Kafka version : 0.10
- Client configuration :  EnableAutoCommit : false; AutoOffsetReset : Earliest

mhowlett:
this is operating as intended - the partitions you assign a client to read from are independent of the assignment given to you by the consumer group. the main use case (which is uncommon) of reading from a different set of partitions than that assigned is adding an additional 'control' partition to the set assigned to you by the CG. note that the java client doesn't let you specify the two independently.

you could achieve what you want with a custom partition assignment strategy, but that's currently not supported by the non-java clients.

finally, this request is unusual - feel free to give some more color so we can understand the reasoning behind the design.

wangjia184:
Thanks @mhowlett 
Our producer sends business data to Kafka.  For business reason, the order of the message is important. Hence the producer sends messages to specific partition.
e.g. Same client's messages are put into the same partition.

As a result,  the number of messages are not distributed well.  Some partition's size is 10x-100x of others.  Hence we want to use a dedicate consumer handling that particular partition.

Manual assignment would work in this case but hard to manage (especially considering one consumer is down)

* When dedicated consumer watchs on a specific partition;  Other consumers consume other partitions.
* When dedicated consumer is down,  other consumers consume all partitions.

ChadJessup:
[...]

PR #1133, opened 2019-12-04
Custom partitioner handler

ChadJessup:
First stab at allowing a Custom Partitioner Handler.

On ProducerBuilder...
```.SetPartitionRequestHandler((producer, partitionRequest) => Partition.Any)```

Only called when Message's partition is ```Partition.Any``` 

If any guidance could be provided for the below, that'd be rad:
1. I can't seem to get the ```Librdkafka.topic_conf_set_partitioner_cb``` callback to be triggered. More work will need to be done there, once that gets figured out.
The Java client seems to do the same thing that I'm doing, which is to have the Produce method just call into the Partitioner class...I'm not sure what the point of the librdkafka callback is for?

1. Custom Partitioner logic in Java receives a ```Cluster``` object:
```MetaData``` seems to functionally be the equivalent, but I'd be worried about calling that for each message, and I'm not sure how best that'd be cached.

### CODE REVIEW

**Comment by edenhill** [comment]
1. The partitioner_cb will be triggered from two places: from the current application thread inside produce() if the partition count of the topic is already known, or from a background librdkafka thread as metadata is received for the topic (this typically only happens at one time (but for all queued messages) per topic).
2. I suggest not providing a Cluster/Metadata object to the partitioner callback, all it really needs is the topic name and partition count, and they are already provided in the partitioner_cb.

File: src/Confluent.Kafka/PartitionRequest.cs [hunk 1]
```diff
@@ -0,0 +1,98 @@
+﻿//    Copyright 2019 Chad Jessup
```
edenhill: The copyright will need to be Confluent's.

src/Confluent.Kafka/PartitionRequest.cs
```diff
@@ -0,0 +1,98 @@
...
+{
+    /// <summary>
+    ///     Request to retrieve the partition to send a message to.
+    /// </summary>
+    /// <remarks>
+    ///     A request for the partition is only generated if the <seealso cref="Partition"/> is set to <seealso cref="Partition.Any"/>.
+    /// </remarks>
+    /// <typeparam name="TKey">The type of the Key.</typeparam>
+    /// <typeparam name="TValue">The type of the Value.</typeparam>
+    public class PartitionRequest<TKey, TValue>
```
edenhill: I'm  not too fond of this name, the word "Request" is already used to indicate a protocol request, which this is not, and there's already existing nomenclature for this, which is Partitioner.

src/Confluent.Kafka/Producer.cs
```diff
@@ -115,6 +116,22 @@ private void ErrorCallback(IntPtr rk, ErrorCode err, string reason, IntPtr opaqu
...
+        private Librdkafka.PartitionerDelegate partitionerCallbackDelegate;
+        private int PartitionerCallback(IntPtr rkt,
+            IntPtr keydata,
+            UIntPtr keylen,
+            int partition_cnt,
+            IntPtr rkt_opaque,
+            IntPtr msg_opaque)
+        {
+            // Ensure registered handlers are never called as a side-effect of Dispose/Finalize (prevents deadlocks in common scenarios).
+            if (ownedKafkaHandle.IsClosed) { return 0; }
```
edenhill: This effectively returns partition 0, you will want to use RD_KAFKA_PARTITION_UA (-1)

src/Confluent.Kafka/Producer.cs
```diff
@@ -660,6 +683,8 @@ internal Producer(ProducerBuilder<TKey, TValue> builder)
                     ex);
             }
 
+            topicPartition = GetPartition(topicPartition, message, message.Key, keyBytes, message.Value, valBytes);
```
edenhill: librdkafka shall trigger the partitioner, not this code.

src/Confluent.Kafka/Producer.cs
```diff
@@ -781,6 +806,8 @@ internal Producer(ProducerBuilder<TKey, TValue> builder)
                 );
             }
 
+            topicPartition = GetPartition(topicPartition, message, message.Key, keyBytes, message.Value, valBytes);
```
edenhill: dito.

src/Confluent.Kafka/Producer.cs
```diff
@@ -807,6 +834,22 @@ internal Producer(ProducerBuilder<TKey, TValue> builder)
             }
         }
 
+        private TopicPartition GetPartition(TopicPartition topicPartition, Message<TKey, TValue> message, TKey key, byte[] keyBytes, TValue value, byte[] valBytes)
```
edenhill: librdkafka will call the partitioner as needed, you should not forcibly call it from the dotnet code.

src/Confluent.Kafka/PartitionRequest.cs
ChadJessup: Thought as much, the LICENSE file in the repo was a little ambiguous, and I didn't want to claim copyright for a 3rd party without some representative stating as much.

src/Confluent.Kafka/PartitionRequest.cs
ChadJessup: I'm not too fond either...
How does ```SetPartionerHandler``` sound for the handler?  And I don't know if this class technically counts as the ```Partitioner``` itself. It's really just a class that holds the data to be passed to the handler...

mhowlett:
I think we probably want to support custom partitioners via a `SetPartitioner(topic, partitioner_cb)` method on the `ProducerBuilder` (need to think about it more though).

Separately, it would also be good to expose the in built partitioners in librdkafka. For example, you may want to align the sharding of your web app servers with kafka partitioning. To route users to the right servers, you'll need the partitioning algorithm used by kafka outside kafka.

ChadJessup:
> I think we probably want to support custom partitioners via a `SetPartitioner(topic, partitioner_cb)` method on the `ProducerBuilder` (need to think about it more though).
> 
> Separately, it would also be good to expose the in built partitioners in librdkafka. For example, you may want to align the sharding of your web app servers with kafka partitioning. To route users to the right servers, you'll need the partitioning algorithm used by kafka outside kafka.

@mhowlett  - by expose, do you mean have an override allowing the ```Paritioner``` enum alongside partitioner_cb?
Or recreating them in C# to be used directly?

How does this sound as an API - intermixed the patterns I can think of, alongside topic scenarios:

``` c#
ProducerBuilder<K, V>()
.SetPartitioner("", callback_cb) <- default topic config or .SetPartitioner(callback_cb)
.SetPartitioner("Topic1", new CustomParitioner()) <- easier for caller, since might be a large chain, or instance shared
.SetPartitioner("Topic2", (string topic, K key, V value, partition_cnt, byte[] topic_opaque, byte[] msg_opaque)) => return 2)
.SetPartitioner("^Topic3|Topic4", Confluent.Kafka.Partitioner.Murmur2)

```

Essentially, topic + an IPartitioner, lambda, or Confluent.Kafka.Partitioner enum.

Question: Opaque data doesn't seem to be exposed in the C# library - but, I've seen partitioners use that feature, should that be taken into account?


Java's Partitioner interface:
https://github.com/apache/kafka/blob/master/clients/src/main/java/org/apache/kafka/clients/producer/Partitioner.java


``` c#
public class IPartitioner<K, V> : IDisposable
{
    // Java's is Configurable, and the configuration system allows a partitioner to be configured like
    // everything else with prefixed key:values, can be removed and constructor can do what it needs
    // void Configure(Dictionary<string, sring> configs);

    int Partition(string topic, K key, V value, int partitionCount);

    // Removed since IDisposable
    // void Close();

    // came in KIP-480 for Sticky Partitioner...not sure if C# client has batches yet? Can come in later, of course, if needed
    void onNewBatch(String topic, Cluster cluster, int prevPartition) {
}
```

ChadJessup:
Ready for a quick look-see to see if this is a good direction for custom partitioners.

Done:
1. Reworked to be used by callback_cb, and allows IPartitioner to be passed in when using the ProducerBuilder.
2. Updated benchmark project to allow this codepath to be executed, with no noticeable difference on messages/sec. But, my dev-laptop setup isn't ideal for that amount of load.

Still need to do:
1. More unit tests
2. Integration tests.
3. Depending on @mhowlett feedback in my last question, probably more.

I left the key data being passed into the Partitioner as an ```IntPtr``` to the data and another for the data length. I felt that also asking the caller to provide Deserializers to the producer is a bigger change, than needed for this?
Also, since these are critical path for every message, I figured it might be good to allow the caller to manage that?  Feedback welcome on that aspect.

Hopefully, this is a good foundation to build the rest on.

ChadJessup:
@mhowlett - this has been waiting for review for a few months. With the closing of #1106 - should I just close this out? Or, should I update it to the newest version of the library?

mhowlett:
no, don't close, likely to get to this in the next few months, it's on the list.

mhowlett:
@ChadJessup - as you may suspect, I completely forgot about this PR (sorry!). We do want to include this feature sometime over the next few months. if you rebase onto master, that would be great, we'll use this work if possible.

ChadJessup:
> @ChadJessup - as you may suspect, I completely forgot about this PR (sorry!). We do want to include this feature sometime over the next few months. if you rebase onto master, that would be great, we'll use this work if possible.

No worries, world is kind of crazy right now.  I'll rebase and do any updates needed for the new version.

ChadJessup:
Going to close and reopen under a new PR.  I screwed up juggling a rebase and multiple remotes on my end.

mhowlett:
thanks!

PR #1136, opened 2019-12-04
Moving developer related docs to DEVELOPER.md

mhowlett:


DEVELOPER.md
```diff
@@ -0,0 +1,51 @@
+# Developer Notes
```
edenhill: Since we also need a CONTRIBUTING.md (how, legalities, attribution, etc, see librdkafka) it might be better to move this developer stuff in there as well.

mhowlett: sounds good. deferred to future iteration :-).

PR #1160, opened 2020-01-11
simplify P&C generics

AndyPook:
There is no need to use generics on the nested private  classes (the types always align).
Remove the <K, V> parts and just use the <TKey, TValue> from the parent

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -654,10 +654,7 @@ namespace Confluent.Kafka
         }
 
 
-        private ConsumeResult<K, V> ConsumeImpl<K,V>(
-            int millisecondsTimeout,
-            IDeserializer<K> keyDeserializer,
-            IDeserializer<V> valueDeserializer)
+        private ConsumeResult<TKey, TValue> ConsumeImpl(int millisecondsTimeout)
         {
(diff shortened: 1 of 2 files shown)


AndyPook:
Just taking it a step further...

Now it would make a lot of sense to expose `private ConsumeImpl(int)` as `public Consume(int)`. Change `IConsumer` to just have this. The other `Consume`s would then be implemented as extensions. (Or at least add `Consume(int)`).
This would mean changing the interface which is, technically, a breaking change (and there are a lot of xmldoc references :( ). But I don't think many would notice other than a recompile. IMO most do not do strict mocked testing at this level. More likely on whatever facade they have wrapped around it.

Thoughts...
I've seen some confusion around having your own config for timeout vs needing to discover "dotnet.cancellation.delay.max.ms".

As `int` is the intrinsic/internal value, needing to convert into TimeSpan then `Consume(TimeSpan)` converting back to int for every consume is a bit "odd".

Just exposing the `int` version on the interface then using extensions means you can provide the other helper overloads you feel are helpful. But also allows library users to also do their own thing.

mhowlett:
Yes, I think it makes sense to open up `ConsumeImpl` as `Consume(int)` (feel free to open that PR). I think we should just leave the other consumes as is to minimize change, though i agree extension methods seem better.

and yeah, I'm not concerned about 'breaking' the interface in this way. requiring a re-compile is fine by us.

PR #1163, opened 2020-01-15
expose `Consume(int)`

AndyPook:
related to #1160

This version just exposes the `Consume(int)` method.

Unfortunately moving the CancellationToken overload to an extension method does not work as it needs access to `cancellationDelayMaxMs` which is pulled from config in the `Consumer` impl. So not available to the extension in a "nice" way.
I did a search in github for  "dotnet.cancellation.delay.max.ms". Only a single hit outside Confluent.Kafka. I haven't seen anyone in "Corp world" use it either. 100ms seems like a reasonable value. We could just expose it as a default param on the method? And remove the config option, leave it up to the dev?

Also the `Consume(TimeStamp)` overload is referred to all over the place in xmldoc. Mostly codegen so probably not that big of a deal.

I do like extensions for this kind of thing. But not as nice as I'd thought in this case.
Keeping it simple for now. I may do another PR with the above changes, just to show what it would look like

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -654,7 +654,10 @@ namespace Confluent.Kafka
         }
 
 
-        private ConsumeResult<TKey, TValue> ConsumeImpl(int millisecondsTimeout)
+        /// <summary>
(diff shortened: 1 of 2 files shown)


mhowlett:
thanks! LGTM

PR #1156, opened 2020-01-02
Revert incorrect null argument check

mhowlett:
resolves #1154

PR #1169, opened 2020-01-30
ListGroups [BUG]

AndyPook:
`ListGroups(TimeSpan)` always fails due to passing group=`null`. Which then trips the null check.

This PR just moves the null check to `ListGroups(group, TimeSpan)`.

Failure and fix proved by running the example AdminClient with something like:
`dotnet .\AdminClient.dll localhost:29092 list-groups`


mhowlett:
thanks! this is exactly the same as #1156 which I can't merge because i didn't get an approval yet - however our process allows me to merge this, so merging...

PR #1189, opened 2020-02-19
Make Headers accessible during serialization

pascalconfluent:


pascalconfluent:
[clabot:check]

src/Confluent.Kafka/Producer.cs
```diff
@@ -664,12 +664,14 @@ internal Producer(ProducerBuilder<TKey, TValue> builder)
             Message<TKey, TValue> message,
             CancellationToken cancellationToken)
         {
+            Headers headers = message.Headers ?? new Headers();
```
mhowlett: ensuring headers is never `null` is a better API from the perspective that the typing is more constrained. however:
- this isn't consistent with `Message`, where headers can be `null`.
- it results in an allocation on every produce which in the vast majority of cases is unnecessary. we'd prefer to trade-off usability here imo.

src/Confluent.Kafka/SerializationContext.cs
```diff
@@ -37,10 +37,16 @@ public static SerializationContext Empty
         /// <param name="topic">
         ///     The topic the data is being written to or read from.
         /// </param>
-        public SerializationContext(MessageComponentType component, string topic)
+        /// <param name="headers">
+        ///     The collection of message headers (or null). Specifying null or an 
+        ///      empty list are equivalent. The order of headers is maintained, and
+        ///     duplicate header keys are allowed.
+        /// </param>       
+        public SerializationContext(MessageComponentType component, string topic, Headers headers)
```
mhowlett: add a default value for headers of `null`, that way it's backwards compatible (after re-compile).

mhowlett:
this looks good to me. with that said, before merging, I probably want a better holistic view of where this is going past this change, so I'll probably end up sitting on it a bit. i.e. we may also want message key and value to be in the SerializationContext, and we may also want to support serialization/deserialization of header values. it would be good to have clarity on what all that looks like ideally to make sure we don't introduce things we're then likely to want to break.

pascalconfluent:
@mhowlett I am using headers in Java to transport the event signature, or a wrapped AES key.
If you want we can jump on the Zoom call and talk about extended to Key/Value, .....

src/Confluent.Kafka/SerializationContext.cs
pascalconfluent: And done :)

src/Confluent.Kafka/Producer.cs
pascalconfluent: @mhowlett Instead of the headers I could change to the entire message. In this case that would be the responsibility of the serializer to allocate the headers if needed.

edenhill: Headers are additive and must not be overwritten: some headers could be added to the Produce() call, others by the serializer, some by producer interceptor plugins and even some by the producer itself.

michaelandrepearce:
This would be useful where someone wants to make a custom serde and instead on putting schema info in the byte[] they want to put it in headers to keep compatibility e.g. my protobuf scenario where existing flows will be having plain serialisation yet want to introduce schema repo without breaking large complex systems.

src/Confluent.Kafka/Producer.cs
mhowlett: following what @edenhill says, that the `Headers` class should probably really not have a `Remove` method. i think we can leave it as is though - the intent of adding headers here is to make them available to the serdes, not that the serdes should be able to change them.

mhowlett:
@michaelandrepearce - didn't notice that this was already in java.

src/Confluent.Kafka/Producer.cs
mhowlett: ahh right, i now see you want the headers to be able to be mutated by the serdes (and this feature makes a lot of sense).

i've now thought about this in depth, and think that your original implementation is the best compromise after all. There's an additional allocation, but short lived so it should be negligible. the api downsides seem significantly worse.

you have some white space issues on the `SerializationContext.Headers` comment - (bad indent and space after the 'an'). If you those whitespace issues and revert the 3rd commit, this PR is good to go. lets get it in 1.4!

michaelandrepearce:
Yes since 0.11 it was added when headers were added :)

src/Confluent.Kafka/Producer.cs
pascalconfluent: Just push a fix with indentation and space and revert to the right push.
Let me know if it's all good or not. :)

Thank you!

mhowlett:
i ended up tweaking this and adding integration tests in #1218 , which is now merged into 1.4.x.

PR #1190, opened 2020-02-21
Expose client name and version to brokers

mhowlett:
note: `AdminClient` utilizes a producer instance under the covers and doesn't need to be explicitly updated.

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -540,9 +540,8 @@ namespace Confluent.Kafka
                 throw new ArgumentException("'group.id' configuration parameter is required and was not specified.");
             }
 
-            var modifiedConfig = config
-                .Where(prop => prop.Key != ConfigPropertyNames.Consumer.ConsumeResultFields)
-                .Concat(Library.NameAndVersionConfig)
+            var modifiedConfig = Library.NameAndVersionConfig
+                .Concat(config.Where(prop => prop.Key != ConfigPropertyNames.Consumer.ConsumeResultFields))
                 .ToList();
 
             var enabledFieldsObj = config.FirstOrDefault(prop => prop.Key == ConfigPropertyNames.Consumer.ConsumeResultFields).Value;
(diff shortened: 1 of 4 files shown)


src/Confluent.Kafka/Consumer.cs
```diff
@@ -541,7 +541,9 @@ internal Consumer(ConsumerBuilder<TKey, TValue> builder)
             }
 
             var modifiedConfig = config
-                .Where(prop => prop.Key != ConfigPropertyNames.Consumer.ConsumeResultFields);
+                .Where(prop => prop.Key != ConfigPropertyNames.Consumer.ConsumeResultFields)
+                .Concat(Library.NameAndVersionConfig)
```
edenhill: The way I'm doing this in librdkafka and Go is to allow the application (or rather a high-high-level binding) to override the default client.software.*, I think the app/binding config should take precedence in this case.

src/Confluent.Kafka/Library.cs
```diff
@@ -125,5 +126,17 @@ public static bool Load(string path)
...
             => kafkaHandleCreateCount - kafkaHandleDestroyCount;
+
+        internal static List<KeyValuePair<string, string>> NameAndVersionConfig
+        {
+            get
+            {
+                return new List<KeyValuePair<string, string>>
+                {
+                    new KeyValuePair<string, string>( "client.software.name", "confluent-kafka-dotnet"),
+                    new KeyValuePair<string, string>( "client.software.version", VersionString )
```
edenhill: note: The version string should be without prefixing v

mhowlett: this is as reported by librdkafka, which looks to always be without a 'v'.

src/Confluent.Kafka/Consumer.cs
mhowlett: yeah, good point. i wasn't doing this because they're hidden, but it's a good feature. I've reversed the order of the concatenation (changes order `rd_kafka_conf_set` is called) which I believe should achieve this without relying on a librdkafka implementation detail that may change.

feel free to confirm before I merge (will hold off for a bit).

PR #1191, opened 2020-02-22
Add support for SSL authentication (certificate)

dinegri:
This PR adds support for SSL Authentication with Schema Registry

Example:

```
var schemaRegistryConfig = new SchemaRegistryConfig
{
    Url = "https://localhost:8082",
    RequestTimeoutMs = 5000,
    MaxCachedSchemas = 10,
    ValueSubjectNameStrategy = SubjectNameStrategy.TopicRecord,
    SslKeystoreLocation = "C:/Certificate.pfx",
    SslKeystorePassword = "changeit",
    SslCaLocation = "C:/CARoot.crt",
};
```

it also allows to use a self signed certificate for development purpose, using the option EnableSslCertificateVerification  = false

Example:

```
var schemaRegistryConfig = new SchemaRegistryConfig
{
    Url = "https://localhost:8082",
    RequestTimeoutMs = 5000,
    MaxCachedSchemas = 10,
    ValueSubjectNameStrategy = SubjectNameStrategy.TopicRecord,
    SslKeystoreLocation = "C:/Certificate.pfx",
    SslKeystorePassword = "changeit",
    SslCaLocation = "C:/CARoot.crt",
    EnableSslCertificateVerification = false,
};
```


dinegri:
[clabot:check]

mhowlett:
Nice! this is often requested - we'll target the release after 1.4 (currently in RC) to add this. we'll also be adding protobuf and json support in that version.

src/Confluent.SchemaRegistry/Confluent.SchemaRegistry.csproj
```diff
@@ -26,4 +26,8 @@
     <PackageReference Include="System.Net.Http" Version="4.3.0" />
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net452' ">
```
mhowlett: do you have an opinion on dropping `net452` support in schema registry entirely?

the official apache avro library, which we're migrating to away from our fork, targets `netstandard2.0` only. We debated this and plan to make the switch in the next release (v1.4) rather holding off and doing a major version bump (v2.0) since we think the negative impact will be relatively low.

this change has already been made on the 1.4.x branch. you might like to target that, though we won't incorporate your PR in 1.4, so it's probably better to leave it hanging for a bit.

dinegri: I think dropping <code>net452</code> is a good way to go,  in schema registry 

As soon as you release version 1.4, I will change the implementation for dropping support to it

mhowlett:
hi @dinegri - 1.4 is out and it'd be great to get this merged, can you remove the net452 code and re-target the 1.4.x branch? 

We should also have a test. if you're feeling super keen the way to do this I think is add a new SSL enabled SR to this docker compose file (note: no need for basic auth on that). https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/test/Confluent.SchemaRegistry.IntegrationTests/docker-compose.yaml

and add a test to `Confluent.SchemaRegistry.IntegrationTests`.

If that's too much effort, no worries at all, i will get to it.

note: the protobuf and json tests currently fail against the docker images - a docker image with protobuf/json enabled SR is not yet available.

dinegri:
Hi @mhowlett to test a real scenario we must have to create a cerificate signed by a trusted CA authority. I added a config to allow self signed certificate, so I am able to create a test using a self signed certificate.

What about create a test only for self signed certificate?

Just to let you know, I am using this solution in production using a trusted CA authority, but trusted CA belongs to the company I work.

dinegri:
Hi @mhowlett 

I added a test to `Confluent.SchemaRegistry.IntegrationTests.`, it's using self signed certificate

Do you see any problem to send jks to GitHub?

src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
```diff
@@ -127,11 +147,71 @@ public string Url
...
+            set { SetObject(SchemaRegistryConfig.PropertyNames.SslKeystorePassword, value.ToString()); }
+        }
+
+        /// <summary>
+        ///     Enable/Disable ssl verify
+        ///
+        ///     default: ''
+        ///     importance: low
+        /// </summary>
+        public bool? SslVerify
```
mhowlett, line 192: all of your property names are consistent with librdkafka except this one. I think it should be `EnableSslCertificateVerification`

mhowlett: (and the underlying string name also updated)

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -74,6 +75,11 @@ public class CachedSchemaRegistryClient : ISchemaRegistryClient, IDisposable
         /// </summary>
         public const int DefaultMaxCachedSchemas = 1000;
 
+        /// <summary>
+        ///     The default ssl verify for Schema Registry REST API calls.
+        /// </summary>
+        public const bool DefaultSslVerify = true;
```
mhowlett, line 81: also update this name

src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
```diff
@@ -127,11 +147,71 @@ public string Url
...
+            set { SetObject(SchemaRegistryConfig.PropertyNames.SslVerify, value); }
+        }
+
+        /// <summary>
+        ///     Specifies the timeout for requests to Confluent Schema Registry.
+        ///
+        ///     default: 30000
+        /// </summary>
+        [Obsolete("This property will be removed in a future version of this library in favor of the RequestTimeoutMs property")]
+        public int? SchemaRegistryRequestTimeoutMs
```
mhowlett, line 204: I think this has been accidentally added back in?

test/Confluent.SchemaRegistry.IntegrationTests/schema.registry.parameters.json
```diff
@@ -1,6 +1,11 @@
...
   "server": "localhost:8081",
   "server_with_auth": "localhost:8082",
+  "server_with_ssl": "https://localhost:8083",
   "username": "ckp_tester",
-  "password": "test_secret"
+  "password": "test_secret",
+  "keystore_location": "./secrets/schema-registry.keystore.jks",
+  "keystore_password": "cnf123",
+  "ca_location": "./secrets/schema-registry-ca.cer",
+  "ssl_verify": "false"
```
mhowlett, line 10: i'd love to see server certificate validation incorporated in the test as well to demonstrate the whole lot working.
how annoying is that request?

mhowlett:
thanks very much @dinegri - this is looking awesome. some minor things and one probably annoying request, what do you think about that? otherwise this is looking good to me.

btw: I have some notes on ssl from while ago which may be useful, though it seems like you know what you're doing. linking to them just in case. https://github.com/mhowlett/confluent-kafka-dotnet/tree/security/examples/Security

dinegri:
Hi @mhowlett ok, no problem

I have a domain (for 10 months I think) and I'll use Let's Encrypts, this way we can have certificates trusted by valid CA Authority (Let's Encrypts in this case), in this scenario I have to configure SR to not validate hostname `ssl.endpoint.identification.algorithm`

mhowlett:
is it possible to use a CA certificate that we generate? ( similar to what I describe here for kafka: https://github.com/mhowlett/confluent-kafka-dotnet/tree/security/examples/Security ), or is there a limitation with `HttpClient` that makes it difficult / impossible to use a custom certificate authority?

dinegri:
Hi, only if CA certificate that you generate is signed by a official Certificate Authority, unfortunally DefaultEnableSslCertificateVerification = false is ignoring error beyond Untrusted certicates and it must be used only for development. 

I am not sure but it could be a security issue let certificates signed by a official Certificate Authority on github. In this case my suggestion is use locally a valid certificate (signed by an entity that issues digital certificates - DigitCert / Let's Encrypt and so on)

src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
dinegri: done

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
dinegri: done

src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
dinegri: yes, my mistake
done

mhowlett:
> only if CA certificate that you generated is signed by a official Certificate Authority

do you know if / are you sure that this is a limitation of `HttpClient`?

can we implement our own `ServerCertificateCustomValidationCallback` to support this? Is that hard?

(these are questions i'm asking myself as much as you). i'm not a super expert on this stuff, hence all the questions...

I think we definitely don't want certificates signed by an official CA on github. another reason for that is we need effectively infinite expiration time so the tests won't break.

dinegri:
> do you know if / are you sure that this is a limitation of HttpClient?

there is no one using self signed certificates without using bypass (custom ServerCertificateCustomValidationCallback)

> can we implement our own ServerCertificateCustomValidationCallback to support this? Is that hard?

it would be good if I could only bypass self signed certificate but I don't know how. I think here we don't need to implement our own ServerCertificateCustomValidationCallback if we have a valid certificate (I mean a not self signed certificate)

Are you able to generate valid certificates signed by confluent CA root signed by a entity authority? You can test it locally disabling the bypass, it will work

mhowlett:
what i was proposing is using a self signed CA (which can be used to avoid the possibility of a man-in-the-middle attack, given we have access to both the server and client).

maybe the standard way to do this in .net/windows world is outside the bounds of the application - maybe you can just install that cert on the client at the machine level.

anyway, i'm starting to think that we should probably consider supporting this as an additional scope of work on top of this PR.

from here i'll have a chat with some security people at work and do a bit of manual verification. but i think we're good to go, or very close.

thanks again for your efforts!

dinegri:
what about dropping support for disabling ssl verify? As you said we can add it in a new PR

mhowlett:
i was thinking that, but we need that to for the integration tests (right?) and i think others will find it useful for test environments as well.

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -74,6 +75,11 @@ public class CachedSchemaRegistryClient : ISchemaRegistryClient, IDisposable
         /// </summary>
         public const int DefaultMaxCachedSchemas = 1000;
 
+        /// <summary>
+        ///     The default ssl verify for Schema Registry REST API calls.
```
edenhill, line 79: "The default SSL server certificate verification for .."

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -201,22 +207,31 @@ public CachedSchemaRegistryClient(IEnumerable<KeyValuePair<string, string>> conf
...
+            this.restService = new RestService(schemaRegistryUris, timeoutMs, username, password, SetSslConfig(config), sslVerify);
         }
 
 
         /// <remarks>
         ///     This is to make sure memory doesn't explode in the case of incorrect usage.
-        /// 
-        ///     It's behavior is pretty extreme - remove everything and start again if the 
+        ///
+        ///     It's behavior is pretty extreme - remove everything and start again if the
```
edenhill, line 232: Nit: avoid whitespace diffs of unrelated code.

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -233,12 +248,37 @@ private bool CleanCacheIfFull()
...
+            var certificates = new List<X509Certificate2>();
+
+            var certificateLocation = config.FirstOrDefault(prop => prop.Key.ToLower() == SchemaRegistryConfig.PropertyNames.SslKeystoreLocation).Value ?? "";
+            var certificatePassword = config.FirstOrDefault(prop => prop.Key.ToLower() == SchemaRegistryConfig.PropertyNames.SslKeystorePassword).Value ?? "";
+            if (!String.IsNullOrEmpty(certificateLocation))
+            {
+                certificates.Add(new X509Certificate2(certificateLocation, certificatePassword));
+            }
+
+            var caLocation = config.FirstOrDefault(prop => prop.Key.ToLower() == SchemaRegistryConfig.PropertyNames.SslCaLocation).Value ?? "";
```
edenhill, line 268: Will it use the system's default CA path if this is not specified?

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -472,4 +512,4 @@ protected virtual void Dispose(bool disposing)
         }
 
     }
-}
+}
```
edenhill, line 515: revert

src/Confluent.SchemaRegistry/Confluent.SchemaRegistry.csproj
```diff
@@ -30,4 +30,4 @@
     <PackageReference Include="System.Net.Http" Version="4.3.0" />
   </ItemGroup>
 
-</Project>
+</Project>
```
edenhill, line 33: revert

src/Confluent.SchemaRegistry/Rest/RestService.cs
```diff
@@ -77,6 +86,20 @@ private static string SanitizeUri(string uri)
...
         }
 
+        private static HttpClientHandler CreateHandler(List<X509Certificate2> certificates, bool enableSslCertificateVerification)
+        {
+            var handler = new HttpClientHandler();
+            handler.ClientCertificateOptions = ClientCertificateOption.Manual;
+
+            if (!enableSslCertificateVerification)
+            {
+                handler.ServerCertificateCustomValidationCallback = (httpRequestMessage, cert, cetChain, policyErrors) => { return true; };
```
edenhill, line 96: typo: cetChain

src/Confluent.SchemaRegistry/Rest/RestService.cs
```diff
@@ -366,4 +389,4 @@ protected virtual void Dispose(bool disposing)
         }
 
     }
-}
+}
```
edenhill, line 392: revert

src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
```diff
@@ -80,6 +80,26 @@ public static class PropertyNames
             /// </summary>
             [Obsolete("Subject name strategies should now be configured using the serializer's configuration. In the future, this configuration property will be removed from SchemaRegistryConfig")]
             public const string SchemaRegistryValueSubjectNameStrategy = "schema.registry.value.subject.name.strategy";
+
+            /// <summary>
+            ///     File path to CA certificate(s) for verifying the schema registry's key.
```
edenhill, line 85: Need to mention what the default behaviour is (hopefully to use the system CA certs)

src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
```diff
@@ -80,6 +80,26 @@ public static class PropertyNames
...
+            /// </summary>
+            public const string SslCaLocation = "schema.registry.ssl.ca.location";
+
+            /// <summary>
+            ///     Ssl keystore location.
```
edenhill, line 90: SSL

edenhill: Is this really a Keystore (which is a java thing), or some other format (PKCS#12?)?
The expected format needs to be specified

src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
```diff
@@ -80,6 +80,26 @@ public static class PropertyNames
...
+            /// </summary>
+            public const string SslKeystoreLocation = "schema.registry.ssl.keystore.location";
+
+            /// <summary>
+            ///     Ssl keystore password.
```
edenhill, line 95: SSL

src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
```diff
@@ -80,6 +80,26 @@ public static class PropertyNames
...
+            /// </summary>
+            public const string SslKeystorePassword = "schema.registry.ssl.keystore.password";
+
+            /// <summary>
+            ///     Ssl verify
```
edenhill, line 100: This doc string is insufficient, needs to say what it does, and why, and a warning to only disable verification in test/dev environments.

src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
```diff
@@ -127,11 +147,58 @@ public string Url
...
+        ///     importance: low
+        /// </summary>
+        public string SslCaLocation
+        {
+            get { return Get(SchemaRegistryConfig.PropertyNames.SslCaLocation); }
+            set { SetObject(SchemaRegistryConfig.PropertyNames.SslCaLocation, value.ToString()); }
+        }
+
+        /// <summary>
+        ///     Path to client's keystore (PKCS#12) used for authentication.
```
edenhill, line 163: ah, here it is! 

I don't really like the keystore naming since that is a Java thing.

src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
```diff
@@ -127,11 +147,58 @@ public string Url
...
+        ///     importance: low
+        /// </summary>
+        public string SslKeystorePassword
+        {
+            get { return Get(SchemaRegistryConfig.PropertyNames.SslKeystorePassword); }
+            set { SetObject(SchemaRegistryConfig.PropertyNames.SslKeystorePassword, value.ToString()); }
+        }
+
+        /// <summary>
+        ///     Enable/Disable SSL certificate verification
```
edenhill, line 187: "SSL server certificate .."

edenhill: Add warning about only using in contained test/dev envs.

test/Confluent.SchemaRegistry.IntegrationTests/Tests/SslAuth.cs
```diff
@@ -0,0 +1,78 @@
...
+            using (var sr = new CachedSchemaRegistryClient(conf))
+            {
+                var topicName = Guid.NewGuid().ToString();
+                var subject = SubjectNameStrategy.Topic.ToDelegate()(new SerializationContext(MessageComponentType.Value, topicName), null);
+                var id = sr.RegisterSchemaAsync(subject, testSchema1).Result;
+                var schema = sr.GetLatestSchemaAsync(subject).Result;
+                Assert.Equal(schema.Id, id);
+            }
+
+            // connect to authenticating with invalid ssl config. shouldn't work.
```
edenhill, line 60: "connect to authenticating" doesn't read right, missing a word?

src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
dinegri: Hi @edenhill, 

I agree, but after checking ProducerConfig or ConsumerConfig configs and to keep the same convention I changed to SslKeystoreLocation, both are PKCS#12 format

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
dinegri: yes, it will

src/Confluent.SchemaRegistry/SchemaRegistryConfig.cs
mhowlett: I think that since a precedent has been set, we should go with the keystore nomenclature, even though i don't like it either. Bikeshedding a bit.

edenhill: I think it was a mistake to call them Keystore (a specific format which is not PKCS12) in the first place, since they are not.
I would rather see us use a proper name for these new properties, and adding aliases in the kafka client with proper names and deprecating the keystore, rather than dig deeper down this hole.
Calling it keystore will not be helpful for our users since they think they can pass a JKS, with all kinds of troubles ahead of them.

edenhill: Matt reminded me that librdkafka has a ssl.keystore.location which takes a PKCS#12, very much like Certificate2, so I redact all my previous opinions and happily cheer on SslKeyStoreLocation, et.al.

mhowlett:
merging. follow up PR coming to fix up a few comments and add to the release notes. thanks for persevering with us on this @dinegri

PR #1197, opened 2020-02-26
Update AvroSpecific example to include a logical type field

mhowlett:


examples/AvroSpecific/User.avsc
```diff
@@ -4,6 +4,12 @@
  "fields": [
      {"name": "name", "type": "string"},
      {"name": "favorite_number",  "type": ["int", "null"]},
-     {"name": "favorite_color", "type": ["string", "null"]}
+     {"name": "favorite_color", "type": ["string", "null"]},
+     {"name": "hourly_rate", "type": {
```
edenhill: Why isnt this type nullable like the others?

mhowlett: no reason - i just pasted something from somewhere else.

you got me intrigued though and tried this case. Unfortunately, there's a bug in the `Avro` implementation that causes `avrogen` to produce code that won't even compile in this case.

mkellogg: I actually fixed that in this pull request (nullable value types) because I noticed the same thing when adding UUIDs, in case you were wondering

https://github.com/apache/avro/pull/822

Thanks for adopting the main avro repo 👍

mhowlett: @mkellogg - that is great news :-)

yep, we're very happy to be discontinuing the fork. it only ever existed so we could move quickly with avro support (as an interim measure). at the time, a fair bit of work was required to modernize the official repo, and it was unclear how much 'human' overhead there would be in making this happen. it's good to see the repo in a good place now, with healthy community contributions!

PR #1209, opened 2020-03-09
Changes for updated librdkafka txn API

mhowlett:
thanks for the cgmd serde api - i owe you a fosters

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -855,5 +855,23 @@ namespace Confluent.Kafka
         /// <inheritdoc/>
         public ConsumeResult<TKey, TValue> Consume(TimeSpan timeout)
             => Consume(timeout.TotalMillisecondsAsInt());
+
+
+        /// <inheritdoc/>
+        public IConsumerGroupMetadata ConsumerGroupMetadata
+        {
+            get
+            {
+                var cgmd = this.kafkaHandle.GetConsumerGroupMetadata();
+                try
+                {
+                    return new ConsumerGroupMetadata { serializedMetadata = this.kafkaHandle.SerializeConsumerGroupMetadata(cgmd) };
+                }
+                finally
+                {
+                    this.kafkaHandle.DestroyConsumerGroupMetadata(cgmd);
+                }
+            }
+        }
     }
 }
--- a/src/Confluent.Kafka/Error.cs
+++ b/src/Confluent.Kafka/Error.cs
@@ -17,6 +17,7 @@
 // Refer to LICENSE for more information.
 
 using System;
+using Confluent.Kafka.Impl;
 
 
 namespace Confluent.Kafka
@@ -33,7 +34,38 @@ namespace Confluent.Kafka
         /// <param name="error">
         ///     The error object to initialize from.
         /// </param>
-        public Error(Error error) : this(error.Code, error.Reason, error.IsFatal) { }
+        public Error(Error error)
+        {
+            this.reason = error.Reason;
+            this.IsFatal = error.IsFatal;
+            this.IsRetriable = error.IsRetriable;
+            this.IsTxnAbortable = error.IsTxnAbortable;
+            this.Code = error.Code;
+        }
+
+        /// <summary>
+        ///     Initialize a new Error instance from a native pointer to
+        ///     a rd_kafka_error_t object, then destroy the native object.
+        /// </summary>
+        internal Error(IntPtr error)
+        {
+            if (error == IntPtr.Zero)
+            {
+                Code = ErrorCode.NoError;
+                reason = null;
+                IsFatal = false;
+                IsRetriable = true;
+                IsTxnAbortable = true;
+                return;
+            }
+            
+            Code = Librdkafka.error_code(error);
+            IsFatal = Librdkafka.error_is_fatal(error);
+            IsTxnAbortable = Librdkafka.error_is_txn_abortable(error);
+            IsRetriable = Librdkafka.error_is_retriable(error);
+            reason = Librdkafka.error_string(error);
+            Librdkafka.error_destroy(error);
+        }
 
         /// <summary>
         ///     Initialize a new Error instance from a particular
@@ -51,6 +83,8 @@ namespace Confluent.Kafka
             Code = code;
             reason = null;
             IsFatal = code == ErrorCode.Local_Fatal;
+            IsRetriable = !IsFatal;
+            IsTxnAbortable = !IsFatal;
         }
 
         /// <summary>
@@ -79,6 +113,8 @@ namespace Confluent.Kafka
             Code = code;
             this.reason = reason;
             IsFatal = isFatal;
+            IsRetriable = !isFatal;
+            IsTxnAbortable = !isFatal;
         }
 
         /// <summary>
@@ -99,6 +135,8 @@ namespace Confluent.Kafka
             Code = code;
             this.reason = reason;
             IsFatal = code == ErrorCode.Local_Fatal;
+            IsRetriable = !IsFatal;
+            IsTxnAbortable = !IsFatal;
         }
 
         /// <summary>
@@ -110,6 +148,17 @@ namespace Confluent.Kafka
         ///     Whether or not the error is fatal.
         /// </summary>
         public bool IsFatal { get; }
+
+        /// <summary>
+        ///     Whether or not the error is retriable.
+        /// </summary>
+        public bool IsRetriable { get; }
+
+        /// <summary>
+        ///     Whether or not an open transaction is abortable
+        ///     following the error.
+        /// </summary>
+        public bool IsTxnAbortable { get; }
 
         private readonly string reason;
 
--- a/src/Confluent.Kafka/Error.cs
+++ b/src/Confluent.Kafka/Error.cs
@@ -83,8 +83,8 @@ namespace Confluent.Kafka
             Code = code;
             reason = null;
             IsFatal = code == ErrorCode.Local_Fatal;
-            IsRetriable = !IsFatal;
-            IsTxnAbortable = !IsFatal;
+            IsRetriable = false;
+            IsTxnAbortable = false;
         }
 
         /// <summary>
@@ -113,8 +113,8 @@ namespace Confluent.Kafka
             Code = code;
             this.reason = reason;
             IsFatal = isFatal;
-            IsRetriable = !isFatal;
-            IsTxnAbortable = !isFatal;
+            IsRetriable = false;
+            IsTxnAbortable = false;
         }
 
         /// <summary>
@@ -135,8 +135,8 @@ namespace Confluent.Kafka
             Code = code;
             this.reason = reason;
             IsFatal = code == ErrorCode.Local_Fatal;
-            IsRetriable = !IsFatal;
-            IsTxnAbortable = !IsFatal;
+            IsRetriable = false;
+            IsTxnAbortable = false;
         }
 
         /// <summary>
@@ -152,7 +152,7 @@ namespace Confluent.Kafka
         /// <summary>
         ///     Whether or not the operation that caused the error is retriable.
         /// </summary>
-        public bool IsRetriable { get; }
+        internal bool IsRetriable { get; }
 
         /// <summary>
         ///     Whether or not the current transaction is abortable
@@ -162,7 +162,7 @@ namespace Confluent.Kafka
         ///     This is only relevant for the transactional producer
         ///     API.
         /// </remarks>
-        public bool IsTxnAbortable { get; }
+        internal bool IsTxnAbortable { get; }
 
         private readonly string reason;
 
--- a/src/Confluent.Kafka/Error.cs
+++ b/src/Confluent.Kafka/Error.cs
@@ -54,8 +54,8 @@ namespace Confluent.Kafka
                 Code = ErrorCode.NoError;
                 reason = null;
                 IsFatal = false;
-                IsRetriable = true;
-                IsTxnAbortable = true;
+                IsRetriable = false;
+                IsTxnAbortable = false;
                 return;
             }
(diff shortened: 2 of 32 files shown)


src/Confluent.Kafka/Consumer.cs
```diff
@@ -855,5 +855,23 @@ internal Consumer(ConsumerBuilder<TKey, TValue> builder)
...
         public ConsumeResult<TKey, TValue> Consume(TimeSpan timeout)
             => Consume(timeout.TotalMillisecondsAsInt());
+
+
+        /// <inheritdoc/>
+        public IConsumerGroupMetadata ConsumerGroupMetadata
+        {
+            get
+            {
+                var cgmd = this.kafkaHandle.GetConsumerGroupMetadata();
```
edenhill: Do note that consumer_group_metadata() may return NULL if called on a producer instance or there is no group.id configured.

src/Confluent.Kafka/Error.cs
```diff
@@ -51,6 +83,8 @@ public Error(ErrorCode code)
             Code = code;
             reason = null;
             IsFatal = code == ErrorCode.Local_Fatal;
+            IsRetriable = !IsFatal;
```
edenhill: Huhm, I'm not sure this is completely correct and I think it would be better to be consistent with the underlying librdkafka errors in that we do not make all existing errors retriable.

Ignoring txnAbortable since it is specific to transactions, there are three error states:
 * fatal - is it not meaningful to continue operating the current client instance.
 * retriable - the error indicated a temporary issue, retry the operation
 * neither - the error is most likely permanent, but that doesn't mean the application needs to treat is a fatal error.

Most non-retriable errors are not fatal, the fatality in these cases is up to the program logic.
E.g., is an UnknownTopic for a QueryWatermarkOffsets() call fatal for the instance? No. Is it fatal for the application? Depends on the program logic.

To be consistent with the underlying librdkafka I think it is better to leave these as false.

The way I'm thinking with APIs that return the new rd_kafka_error_t type is instead of documenting specific error codes, the API docs should simply state what the application should based on the three error states above.
Not all APIs might return a retriable error, and skipping it from the doc string saves people from wrapping all calls in retry-loops.

src/Confluent.Kafka/Error.cs
```diff
@@ -111,6 +149,17 @@ public Error(ErrorCode code, string reason)
         /// </summary>
         public bool IsFatal { get; }
 
+        /// <summary>
+        ///     Whether or not the error is retriable.
```
edenhill: It is the operation that is retriable, not the error.

src/Confluent.Kafka/Error.cs
```diff
@@ -111,6 +149,17 @@ public Error(ErrorCode code, string reason)
...
+        /// </summary>
+        public bool IsRetriable { get; }
+
+        /// <summary>
+        ///     Whether or not an open transaction is abortable
```
edenhill: Add note that this is only relevant for the transactional producer API.
And I would phrase it as 'current transaction' as not to add a new state to transactions nomenclature (Open).

src/Confluent.Kafka/Error.cs
```diff
@@ -33,7 +34,38 @@ public class Error
...
+        ///     a rd_kafka_error_t object, then destroy the native object.
+        /// </summary>
+        internal Error(IntPtr error)
+        {
+            if (error == IntPtr.Zero)
+            {
+                Code = ErrorCode.NoError;
+                reason = null;
+                IsFatal = false;
+                IsRetriable = true;
```
edenhill: See other note below: these defaults don't match librdkafka.

src/Confluent.Kafka/Consumer.cs
mhowlett: neither of these cases possible - `rd_kafka_poll_set_consumer`, which is called in the consumer constructor, forces that `group.id` must be set, and there is no way for a user to call this method on a producer `rk`.

note that in the docs for `IConsumer.ConsumerGroupMetadata` I did document that this will be `null` if `GroupId` is not set, however I did that to ensure docs wouldn't go out of date in the event this got fixed (i considered the miss-documentation harmless, as obvious exceptions will be thrown on consumer startup in this case not causing confusion).

src/Confluent.Kafka/Error.cs
mhowlett: phrasing: yep.
i was thinking the note is redundant, but yep, might as well.

src/Confluent.Kafka/Error.cs
mhowlett: > The way I'm thinking with APIs that return the new rd_kafka_error_t type is instead of documenting specific error codes, the API docs should simply state what the application should/could do based on the three error states above.

yes, but they should be `true` i think (as i have them, the negation of fatal), not `false`, to indicate to the application that yes, this is retriable, and yes, this is txn abortable. UnknownTopic for QueryWatermarkOffsets or QueueFull on Produce - both should have IsRetriable and IsTxnAbortable `true` (and fatal `false`).

I was definitely thinking this was a dangerous approximation to be making, and also considered making these constructors obsolete so the dangerous approximation is not exposed outside the insides of the library (where it can be fixed as needed). that was more work though so I didn't. I might do that now.

edenhill: txnAbortable should definitely not be set by default, it does not even have meaning outside the transactional API and setting it to false will just be confusing.

As for retriable: I'd argue that UnknownTopic is not a retriable error, if the topic was not automatically created by referencing it the first time it is unlikely that another call to the same API will get better results, and it just tricks users into repeatedly calling APIs, which could have side effects.

QueueFull, Timed out, Transport failure, on the other hand are retriable, but we should not blindly base the retriable flag on the error code itself, but a combination of error and API (and underlying protocol request), and that is something only librdkafka should do since  the bindings don't have enough information to make that decision.

So, to play it safe I'd urge us to ease into this transition by only setting these flags where we're completely sure they are connect, which is currently on the transactional API.

edenhill: And, it is better to error on the side of caution.

mhowlett: > txnAbortable should definitely not be set by default, it does not even have meaning outside the transactional API and setting it to false will just be confusing.

This makes complete sense to me - if you *are* in a transaction, you're ok to abort it (if you're not, just ignore, it's clear what to do). The opposite of what you say is the most natural to me. The difference in the way we reationalize this underscores that this is a crap API. What happens if you do QueryWatermarkOffsets inside a transaction? IsTxnAbortable should be `true`, it won't be. Many other examples. People will take these values literally. Marking QueueFull as not-retriable will confuse people.

For now, I've just hidden these properties, which keeps the Error class API the same as before and not confusing. we can work out what to do later. This is a step up from providing non-meaningful values.

edenhill: txn_abortable() doesn't mean it is okay but optional to abort(), but that the transaction needs to be aborted for the producer to recover, and this will only be signalled from transactional APIs.
It doesn't make sense for other APIs (QueryWmarks..) to set that flag since it has nothing to do with transactions.

I think a TransactionError sub-class would be a good solution to avoid confusion here though and not keep txnAbortable on the main Error class.

> People will take these values literally. Marking QueueFull as not-retriable will confuse people.

Yes, but we need to do this organically to not change semantics later, and the easiest way to do that is to document that retriable(), et.al, are only relevant for the transactional API at this point.
Example:
https://github.com/confluentinc/confluent-kafka-go/blob/master/kafka/producer.go#L855
> // Returns nil on success or an error object on failure.
> // Check whether the returned error object permits retrying
> // by calling `err.(kafka.Error).IsRetriable()`, or whether an abortable
> // or fatal error has been raised by calling
> // `err.(kafka.Error).IsTxnAbortable()` or `err.(kafka.Error).IsFatal()`
> // respectively.

I.e., it explicitly tells the user how to handle the returned error object.

mhowlett: > txn_abortable() doesn't mean it is okay but optional to abort(), but that the transaction needs to be aborted for the producer to recover, and this will only be signalled from transactional APIs.

The name implies to me the semantics I assumed. If this were called `txn_requires_abort()`, I would have guessed the semantics correctly. change the name?

> `TransactionError`

maybe - for sure I'm thinking there would be a `KafkaTransactionException`. I think i would just put the `TxnRequiresAbort` property on that rather than create another error class as well (but would implement and see how it feels).

can you clarify exactly the intended semantics of `is_retriable` - i believe your plan that this be applicable everywhere?

If so, I think that should be a property on `Error`. What I can do is make this throw a `NotImplementedException` by default if accessed. For the cases where we have an `error_t`, i can populate with the value from that. For selected other cases I can make it behave correctly (maybe just do the QueueFull on produce as it's common). Alternatively, just leave this off.

> and the easiest way to do that is to document that retriable(), et.al, are only relevant for the transactional API at this point.

people won't read the docs - they will see the property exists in intellisense and make assumptions about how it works in every context. the more we require people read the docs the more difficult and time consuming it is to use and the more mistakes people will make.

src/Confluent.Kafka/IProducer.cs
```diff
@@ -339,6 +339,14 @@ public interface IProducer<TKey, TValue> : IClient
         ///     Any produce call outside an on-going transaction, or for a failed
         ///     transaction, will fail.
         /// </remark>
+        /// <exception cref="KafkaTxnRequiresAbortException">
```
mhowlett: can librdkafka cause this to occur here?

src/Confluent.Kafka/IProducer.cs
```diff
@@ -391,13 +407,16 @@ public interface IProducer<TKey, TValue> : IClient
         /// <param name="timeout">
         ///     The maximum length of time this method may block.
         /// </param>
+        /// <exception cref="KafkaException">
+        ///     Thrown on all errors.
+        /// </exception>
```
mhowlett: Is `KafkaTxnRequiresAbortException` possible here?

mhowlett:
notes on latest changes:
- can you double check where I assume the txn_requires_abort error can occur.
- ultimately, I plan to expose `IsRetriable` on `Error`. We may optionally create `KafkaFatalException` and `KafkaRetriableException` as well (but that decision can be deferred even longer). I don't want to do this now because I think inappropriate default values will do more harm than good.
- I don't plan to expose `TxnRequiresAbort` on `Error` ever. This error will only be exposed to users via the new `KafkaTxnRequiresAbortException` (which derives from `KafkaException`, so a lazy programmer can do a catch-all).
- These changes are conservative, because the only user facing change I've done is `KafkaTxnRequiresAbortException`, which I'm quite sure we want.

src/Confluent.Kafka/IProducer.cs
```diff
@@ -339,6 +339,14 @@ public interface IProducer<TKey, TValue> : IClient
         ///     Any produce call outside an on-going transaction, or for a failed
         ///     transaction, will fail.
         /// </remark>
+        /// <exception cref="TransactionRequiresAbortException">
```
edenhill: BeginTransaction will not raise ABORTABLE.
All the transaction APIs have doc strings to say which error types are raised:
https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L6699

src/Confluent.Kafka/IProducer.cs
edenhill: Nope, see https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L6883 and make sure your docs match.

src/Confluent.Kafka/TransactionRequiresAbortException.cs
```diff
@@ -0,0 +1,40 @@
...
+//
+// Refer to LICENSE for more information.
+
+using System;
+
+
+namespace Confluent.Kafka
+{
+    /// <summary>
+    ///     Represents an error that occured during an interaction
```
edenhill: It is more specific than that, it represents an error that caused the current transaction to fail and enter the abortable state...

src/Confluent.Kafka/IProducer.cs
edenhill: Nope, see https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L6699

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -492,54 +492,107 @@ internal ErrorCode PollSetConsumer()
...
+                throw new ArgumentException("groupMetadata object must be a value acquired via Consumer.ConsumerGroupMetadata.");
+            }
+            var serializedMetadata = ((ConsumerGroupMetadata)groupMetadata).serializedMetadata;
+            var cgmdPtr = this.DeserializeConsumerGroupMetadata(serializedMetadata);
+            try
+            {
+                var error = new Error(Librdkafka.send_offsets_to_transaction(this.handle, offsetsPtr, cgmdPtr, (IntPtr)millisecondsTimeout));
+                if (error.Code != ErrorCode.NoError)
+                {
+                    if (error.TxnRequiresAbort)
```
edenhill: Did we decide not to check retriable?

src/Confluent.Kafka/TransactionRequiresAbortException.cs
mhowlett: thanks

src/Confluent.Kafka/IProducer.cs
mhowlett: thanks

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
mhowlett: I wasn't going to, but you made me realize there's no reason not to. Done in exactly the same way as txn-requires-abort.

PR #1215, opened 2020-03-16
Made message convenience properties on ConsumeResult obsolete

mhowlett:
The message convenience properties on `ConsumeResult` were added at a time when it was never the case that `ConsumeResult.Message` could be `null`. Soon after, this became possible but the properties were never removed. They're a recurring source of confusion, so i'm marking them as obsolete.

@edenhill

### DIFF
--- a/test/Confluent.SchemaRegistry.Serdes.UnitTests/Configuration.cs
+++ b/test/Confluent.SchemaRegistry.Serdes.UnitTests/Configuration.cs
@@ -33,7 +33,7 @@ namespace Confluent.SchemaRegistry.Serdes.UnitTests
         {
             testTopic = "topic";
             var schemaRegistryMock = new Mock<ISchemaRegistryClient>();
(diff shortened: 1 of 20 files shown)


test/Confluent.SchemaRegistry.Serdes.UnitTests/Configuration.cs
```diff
@@ -33,7 +33,7 @@ public ConfigurationTests()
         {
             testTopic = "topic";
             var schemaRegistryMock = new Mock<ISchemaRegistryClient>();
-            schemaRegistryMock.Setup(x => x.RegisterSchemaAsync(testTopic + "-value", It.IsAny<string>())).ReturnsAsync(
+            schemaRegistryMock.Setup(x => x.RegisterSchemaAsync(testTopic + "-value", It.IsAny<Schema>())).ReturnsAsync(
```
edenhill: This one is different from the other changes in this PR.

mhowlett: i missed it in the SR PR.

PR #1218, opened 2020-03-19
Make headers available in SerializationContext

mhowlett:
minor tweaks to #1189

### DIFF
--- a/src/Confluent.Kafka/Producer.cs
+++ b/src/Confluent.Kafka/Producer.cs
@@ -664,18 +664,14 @@ namespace Confluent.Kafka
             Message<TKey, TValue> message,
             CancellationToken cancellationToken)
         {
-            Headers headers = message.Headers;
+            Headers headers = message.Headers ?? new Headers();
 
             byte[] keyBytes;
             try
             {
-                SerializationContext context = new SerializationContext(MessageComponentType.Key, topicPartition.Topic, headers);
-
                 keyBytes = (keySerializer != null)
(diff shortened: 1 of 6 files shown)


src/Confluent.Kafka/Producer.cs
```diff
@@ -664,12 +664,14 @@ internal Producer(ProducerBuilder<TKey, TValue> builder)
             Message<TKey, TValue> message,
             CancellationToken cancellationToken)
         {
+            Headers headers = message.Headers ?? new Headers();
```
edenhill: Do we really want to do this for each message when the new majority of users will not need headers in their serializers?
How about passing message.Headers by reference, allowing it to be null on input and serializers creating it if it does not exist?

Alternatively have the serializers optionally return a set of headers it wants appended to the headers list?
This is more convenient but it means serializers will not have access to the headers added by the user in the Produce() call.

edenhill: In the case where `new Headers` is called there doesn't seem to be an assignment of `message.Headers = headers`, meaning the headers would be lost, right?

mhowlett: it's `headers` which is used from this point on, not `message.Headers` - the logic is correct.

my reaction to the allocation was exactly the same as yours, but the perf hit won't be significant (it's short lived), and my preference swung to the better API.

it's most logical this is part of `SerializationContext` in my view. also, it's used by both producer and consumer. also, the producer may want the headers, we don't want to restrict that. also this closely matches java impl.

PR #1227, opened 2020-03-27
-> 1.4.0-RC5

mhowlett:


test/Confluent.Kafka.IntegrationTests/Tests/Consumer_OffsetsForTimes.cs
```diff
@@ -38,63 +38,66 @@ public void Consumer_OffsetsForTimes(string bootstrapServers)
...
             var consumerConfig = new ConsumerConfig
             {
                 GroupId = Guid.NewGuid().ToString(),
                 BootstrapServers = bootstrapServers
             };
 
-            var firstMessage = messages[0];
-            var lastMessage = messages[N - 1];
-            using (var consumer = new ConsumerBuilder<byte[], byte[]>(consumerConfig).Build())
+            using (var topic = new TemporaryTopic(bootstrapServers, 1))
```
mhowlett: This change is simply to use a temporary topic instead of the (old, to be discontinued) pre-created single partition topic. There was a bug in the test that assumed the topic was empty which I didn't pick up on previously because i didn't run the complete set of tests.

test/Confluent.Kafka.IntegrationTests/Tests/Producer_Poll.cs
```diff
@@ -47,7 +47,8 @@ public void Producer_Poll(string bootstrapServers)
                 // call to Poll (or if the poll method hasn't been called, over
                 // the lifetime of the producer).
                 Assert.True(producer.Poll(TimeSpan.FromMilliseconds(500)) >= 1);
-                Assert.True(sw.ElapsedMilliseconds < 1);
+                var elapsed = sw.ElapsedMilliseconds;
+                Assert.True(elapsed < 2);
```
mhowlett: This suddenly stopped working today. a lot has changed about my environment since yesterday (new distro, new .net install), so there's a good chance the failure is related to that. But please have a quick think about whether anything in librdkafka change that could make poll take longer.

CHANGELOG.md
```diff
@@ -2,19 +2,22 @@
 
 ## Enhancements
 
-- References librdkafka v1.4.0-RC4. Refer to the [release notes](https://github.com/edenhill/librdkafka/releases/tag/v1.4.0) for more information. Headline features:
-  - Producer support for transactions ([@edenhill](https://github.com/edenhill)). This is the final piece in the puzzle required to enable exactly once stream processing (EOS) in .NET.
-- Added a [WordCount](https://github.com/confluentinc/confluent-kafka-dotnet/tree/master/examples/Transactions) example demonstrating a streaming map-reduce application with exactly-once processing.
+- References librdkafka v1.4.0-RC5. Refer to the [release notes](https://github.com/edenhill/librdkafka/releases/tag/v1.4.0) for more information. Headline features:
+  - KIP-98: Producer support for transactions ([@edenhill](https://github.com/edenhill)). This is the final piece in the puzzle required to enable exactly once stream processing (EOS) in .NET.
```
edenhill: Exactly Once Semantics (EOS)   would be more correct.

test/Confluent.Kafka.IntegrationTests/Tests/Producer_Poll.cs
edenhill: Why is this asserting to less than 2 milliseconds when the comment says 500?

mhowlett: the timeout on poll is 500 milliseconds (the operation should be close to instantaneous). previously, this test has never failed. now it is failing reliably. either the precision of Stopwatch has changed somehow or librdkafka is taking longer in poll.

edenhill: Measure (and debug=protocol) and ye shall find.

Also, 2ms does not yield a very robust test.

mhowlett: that doesn't tell me.

agree the 2ms is not robust, but gonna leave it for now. Also, this is now passing with 1ms, underscoring your point.

Issue #1254, opened 2020-04-18
Consumer reassignment to Offset.End does not work as expected


AndrewBoklashko:
Description
===========
When consumer is assigned to the end of partition then unassigned and after some time again assigned to the end of partition, it will be consuming from the offset at which it was unassigned, while the expected behaviour is to consume from the end of partition.

How to reproduce
================

```C#
var producer =
    new ProducerBuilder<Null, string>(
            new ProducerConfig
            {
                BootstrapServers = kafkaBrokers,
                Acks = Acks.None
            })
        .Build();

Task.Run(
    async () =>
    {
        while (true)
        {
            producer.Produce(
                testTopic,
                new Message<Null, string>
                {
                    Value = DateTime.UtcNow.ToString()
                });
            await Task.Delay(1000);
        }
    });

var consumer =
    new ConsumerBuilder<Null, string>(
            new ConsumerConfig
            {
                GroupId = Guid.NewGuid().ToString(),
                BootstrapServers = kafkaBrokers
            })
        .Build();

Task.Run(
    () =>
    {
        while (true)
        {
            var consumeResult = consumer.Consume();
            logger.Info($"Consumed: {consumeResult.Value}");
        }
    });

consumer.Assign(new TopicPartitionOffset(testTopic, Offset.End));
logger.Info("Subscribed.");

Task.Delay(TimeSpan.FromSeconds(5)).ContinueWith(t =>
{
    consumer.Assign(Enumerable.Empty<TopicPartition>());
    logger.Info("Unsubscribed.");
});

Task.Delay(TimeSpan.FromSeconds(10)).ContinueWith(t =>
{
    consumer.Assign(new TopicPartitionOffset(testTopic, Offset.End));
    logger.Info("Subscribed.");
});

Task.Delay(TimeSpan.FromSeconds(15)).ContinueWith(t =>
{
    consumer.Assign(Enumerable.Empty<TopicPartition>());
    logger.Info("Unsubscribed.");
});
```

Actual result:
```
2020-04-18 13:27:29,853 INFO  [1]: Log - Subscribed.
2020-04-18 13:27:31,115 INFO  [5]: Log - Consumed: 18.04.2020 13:27:30
2020-04-18 13:27:32,159 INFO  [5]: Log - Consumed: 18.04.2020 13:27:31
2020-04-18 13:27:33,209 INFO  [5]: Log - Consumed: 18.04.2020 13:27:32
[...]

PR #1256, opened 2020-04-20
Added test for Consumer.Assign with Offset.Beginning/Offset.End

mhowlett:
test for #1254

### DIFF
--- a/test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Assign.cs
+++ b/test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Assign.cs
@@ -50,9 +50,9 @@ namespace Confluent.Kafka.IntegrationTests
                     Util.ProduceNullStringMessages(bootstrapServers, topic.Name, 1, N);
                     consumer.Resume(new List<TopicPartition> { new TopicPartition(topic.Name, 0) });
                     var cr3 = consumer.Consume(TimeSpan.FromSeconds(10));
-                    Assert.Equal(5, cr3.Offset);
+                    Assert.Equal(N, cr3.Offset);
                     var p = consumer.Position(new TopicPartition(topic.Name, 0));
-                    Assert.Equal(6, p.Value);
+                    Assert.Equal(N+1, p.Value);
                     consumer.Assign(new TopicPartitionOffset(topic.Name, 0, Offset.Beginning));
                     var cr4 = consumer.Consume(TimeSpan.FromSeconds(10));
                     Assert.Equal(0, cr4.Offset);
@@ -63,9 +63,8 @@ namespace Confluent.Kafka.IntegrationTests
             {
                 GroupId = Guid.NewGuid().ToString(),
                 BootstrapServers = bootstrapServers,
-                SessionTimeoutMs = 6000,
                 EnableAutoCommit = false,
-                AutoOffsetReset = AutoOffsetReset.Earliest
+                AutoOffsetReset = AutoOffsetReset.Error
             };
 
             using (var consumer = new ConsumerBuilder<byte[], byte[]>(consumerConfig).Build())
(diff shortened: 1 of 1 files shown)


test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Assign.cs
```diff
@@ -0,0 +1,87 @@
...
+                    Assert.Equal(0, cr1.Offset);
+                    // test that if consumer is paused and reassigned, this does not affect offset.
+                    consumer.Pause(new List<TopicPartition> { new TopicPartition(topic.Name, 0) });
+                    consumer.Assign(new TopicPartitionOffset(topic.Name, 0, Offset.End));
+                    var cr2 = consumer.Consume(1000);
+                    Assert.Null(cr2);
+                    Util.ProduceNullStringMessages(bootstrapServers, topic.Name, 1, N);
+                    consumer.Resume(new List<TopicPartition> { new TopicPartition(topic.Name, 0) });
+                    var cr3 = consumer.Consume(TimeSpan.FromSeconds(10));
+                    Assert.Equal(5, cr3.Offset);
```
edenhill, line 53: nit: mix of constant and not: 5 should be N here, and 6 be N + 1.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Assign.cs
```diff
@@ -0,0 +1,87 @@
...
+                    var cr4 = consumer.Consume(TimeSpan.FromSeconds(10));
+                    Assert.Equal(0, cr4.Offset);
+                }
+            };
+
+            var consumerConfig = new ConsumerConfig
+            {
+                GroupId = Guid.NewGuid().ToString(),
+                BootstrapServers = bootstrapServers,
+                SessionTimeoutMs = 6000,
```
edenhill, line 66: not needed since you subscribe is not used.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Assign.cs
```diff
@@ -0,0 +1,87 @@
...
+                EnableAutoCommit = false,
+                AutoOffsetReset = AutoOffsetReset.Earliest
```
edenhill, line 68: Set this to .Error to make sure the Assign():ed offset is used and not auto.offset.reset.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Assign.cs
mhowlett: good point

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Assign.cs
mhowlett: good point, cut 'n paste error.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Assign.cs
mhowlett: don't you think it's a better test as is? (since the behavior if `Earliest` should still be to honor the offset passed into `Assign`, and there is probably more chance the code doesn't work this way than with `Error`).

edenhill: Since you are passing Beginning (same as Earliest) to assign there is no way to distinguish if it was the offset passed to Assign() or auto.offset.reset that was used.
Setting it to Error will raise an error if the auto.offset.reset is used, which is a stricter (and thus better) test case.

mhowlett: ahh right, i see your point now.

PR #1290, opened 2020-05-15
UrlEncode subject name

### LINKED ISSUE #1258: ProtobufSerializer produces reference subject names which contain slashes
The situation is that I have a proto file `foo1.proto` which imports another file in a subfolder `Protolib/foo2.proto`. I then use Grpc.Tools to generate messages from these and would like to produce these to Kafka in protobuf format using a schema registry. The problem comes in registering the schemata via the ProtobufSerializer. In my case the serialiser tries to register the schema in `/subjects/foo1.proto/versions` (which is OK) and the dependency in `/subjects/protolib/foo2.proto/versions` (which causes a 404 error in the registry). 

An example project is here: [ProtobufProducerError.zip](https://github.com/confluentinc/confluent-kafka-dotnet/files/4522827/ProtobufProducerError.zip)

It seems that in ProtobufSerializer there is no way to modify the ReferenceSubjectNameStrategy, which at the moment just uses the FileDescriptor.Name, and can contain `/`. Are there are plans to extend the strategies available? Ideally, it would be great to be able to set a custom subject for messages and their references.

mhowlett:
this will hopefully resolve #1258 
regardless, we should be doing this.

### DIFF
--- a/src/Confluent.SchemaRegistry/Rest/RestService.cs
+++ b/src/Confluent.SchemaRegistry/Rest/RestService.cs
@@ -279,32 +279,32 @@ namespace Confluent.SchemaRegistry
                         .ConfigureAwait(continueOnCapturedContext: false);
 
         public async Task<List<int>> GetSubjectVersionsAsync(string subject)
-            => await RequestListOfAsync<int>($"subjects/{subject}/versions", HttpMethod.Get)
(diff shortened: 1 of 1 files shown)


mhowlett:
integration test == opportunity cost

Issue #871, opened 2019-04-08
Support for OAuthBearer SASL authentication mechanism


kuehlmeyer:
Description
===========

It seems there's no support for the SASL mechanism OAuthBearer (see https://docs.confluent.io/current/kafka/authentication_sasl/authentication_sasl_oauth.html#) in the .NET version of the Confluent.Kafka lib. The class SaslMechanism has only the members Gssapi, Plain, ScramSha256, and ScramSha512 in 1.0.0-RC3.

Are the plans to support OAuthBearer in the near future? Or is there another way to use it for .NET producers and consumers?

How to reproduce
================

See source of SaslMechanism class.

Checklist
=========

Please provide the following information:
 - [ ] A complete (i.e. we can run it), minimal program demonstrating the problem. No need to supply a project file.
 - [1.0.0-RC3 ] Confluent.Kafka nuget version.
 - [ ] Apache Kafka version.
 - [ ] Client configuration.
 - [Windows 10] Operating system.
 - [ ] Provide logs (with "debug" : "..." as necessary in configuration).
 - [ ] Provide broker log excerpts.
 - [ ] Critical issue.

mhowlett:
yes, a PR for this is in the late stages of review in librdkafka, you can expect it in a post-1.0 release (which will occur more frequently than 0.11.6 -> 1.0!): https://github.com/edenhill/librdkafka/pull/2189

Fallibilist:
@mhowlett 
Any progress on this?

mhowlett:
it's here: https://github.com/edenhill/librdkafka/releases/tag/v1.1.0

I omitted to regenerate the strongly typed .NET Config classes for the v1.1 release, but you can still use the relevant config via librdkafka style properties. Will fix this in the upcoming 1.2 release. 

https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md

edenhill:
For OAUTHBEARER to be meaningful the application must provide a token refresh callback and token acquisition logic, the built-in unsecure JWT is just for testing, and not for actual production use.

I believe the .NET client currently lacks the new APIs required to perform the token operations.

Fallibilist:
@edenhill 
So it is currently possible with the .NET client? It is straightforward enough to create the callback to fetch the JWT, but I'm not certain how to pass the callback to the producer config. 

[...]

PR #1295, opened 2020-05-20
Feature/871/oauthbearer

thtp:
fixes https://github.com/confluentinc/confluent-kafka-dotnet/issues/871

### DIFF
--- a/src/ConfigGen/Program.cs
+++ b/src/ConfigGen/Program.cs
@@ -106,7 +106,7 @@ namespace ConfigGen
         /// <summary>
         ///     OAUTHBEARER
         /// </summary>
-        OauthBearer
+        OAuthBearer
     }
 ";
 
@@ -128,7 +128,7 @@ namespace ConfigGen
                 if (r == ""PLAIN"") { return Confluent.Kafka.SaslMechanism.Plain; }
                 if (r == ""SCRAM-SHA-256"") { return Confluent.Kafka.SaslMechanism.ScramSha256; }
                 if (r == ""SCRAM-SHA-512"") { return Confluent.Kafka.SaslMechanism.ScramSha512; }
-                if (r == ""OAUTHBEARER"") { return Confluent.Kafka.SaslMechanism.OauthBearer; }
+                if (r == ""OAUTHBEARER"") { return Confluent.Kafka.SaslMechanism.OAuthBearer; }
                 throw new ArgumentException($""Unknown sasl.mechanism value {r}"");
             }
             set
@@ -138,7 +138,7 @@ namespace ConfigGen
                 else if (value == Confluent.Kafka.SaslMechanism.Plain) { this.properties[""sasl.mechanism""] = ""PLAIN""; }
                 else if (value == Confluent.Kafka.SaslMechanism.ScramSha256) { this.properties[""sasl.mechanism""] = ""SCRAM-SHA-256""; }
                 else if (value == Confluent.Kafka.SaslMechanism.ScramSha512) { this.properties[""sasl.mechanism""] = ""SCRAM-SHA-512""; }
-                else if (value == Confluent.Kafka.SaslMechanism.OauthBearer) { this.properties[""sasl.mechanism""] = ""OAUTHBEARER""; }
+                else if (value == Confluent.Kafka.SaslMechanism.OAuthBearer) { this.properties[""sasl.mechanism""] = ""OAUTHBEARER""; }
                 else throw new ArgumentException($""Unknown sasl.mechanism value {value}"");
             }
         }
--- a/src/ConfigGen/Program.cs
+++ b/src/ConfigGen/Program.cs
@@ -123,6 +123,7 @@ namespace ConfigGen
                 if (r == ""PLAIN"") { return Confluent.Kafka.SaslMechanism.Plain; }
                 if (r == ""SCRAM-SHA-256"") { return Confluent.Kafka.SaslMechanism.ScramSha256; }
                 if (r == ""SCRAM-SHA-512"") { return Confluent.Kafka.SaslMechanism.ScramSha512; }
+                if (r == ""OAUTHBEARER"") { return Confluent.Kafka.SaslMechanism.OauthBearer; }
                 throw new ArgumentException($""Unknown sasl.mechanism value {r}"");
             }
             set
@@ -132,6 +133,7 @@ namespace ConfigGen
                 else if (value == Confluent.Kafka.SaslMechanism.Plain) { this.properties[""sasl.mechanism""] = ""PLAIN""; }
                 else if (value == Confluent.Kafka.SaslMechanism.ScramSha256) { this.properties[""sasl.mechanism""] = ""SCRAM-SHA-256""; }
                 else if (value == Confluent.Kafka.SaslMechanism.ScramSha512) { this.properties[""sasl.mechanism""] = ""SCRAM-SHA-512""; }
+                else if (value == Confluent.Kafka.SaslMechanism.OauthBearer) { this.properties[""sasl.mechanism""] = ""OAUTHBEARER""; }
                 else throw new ArgumentException($""Unknown sasl.mechanism value {value}"");
             }
         }
--- a/src/ConfigGen/Program.cs
+++ b/src/ConfigGen/Program.cs
@@ -102,6 +102,11 @@ namespace ConfigGen
         ///     SCRAM-SHA-512
         /// </summary>
         ScramSha512
+
+        /// <summary>
+        ///     OAUTHBEARER
+        /// </summary>
+        OauthBearer
     }
 ";
 
--- a/src/Confluent.Kafka/AdminClient.cs
+++ b/src/Confluent.Kafka/AdminClient.cs
@@ -502,18 +502,6 @@ namespace Confluent.Kafka
             => kafkaHandle.AddBrokers(brokers);
 
         /// <summary>
-        ///     Refer to <see cref="Confluent.Kafka.IClient.OauthBearerSetToken(string, long, string, IDictionary{string, string})" />
-        /// </summary>
-        public void OauthBearerSetToken(string tokenValue, long lifetimeMs, string principalName, IDictionary<string, string> extensions)
-            => kafkaHandle.OauthBearerSetToken(tokenValue, lifetimeMs, principalName, extensions);
-        
-        /// <summary>
-        ///     Refer to <see cref="Confluent.Kafka.IClient.OauthBearerSetTokenFailure(string)" />
-        /// </summary>
-        public void OauthBearerSetTokenFailure(string errstr)
-            => kafkaHandle.OauthBearerSetTokenFailure(errstr);
-
-        /// <summary>
         ///     Refer to <see cref="Confluent.Kafka.IClient.Name" />
         /// </summary>
         public string Name
--- a/src/Confluent.Kafka/AdminClient.cs
+++ b/src/Confluent.Kafka/AdminClient.cs
@@ -501,6 +501,17 @@ namespace Confluent.Kafka
         public int AddBrokers(string brokers)
             => kafkaHandle.AddBrokers(brokers);
 
+        /// <summary>
+        ///     Refer to <see cref="Confluent.Kafka.IClient.OauthBearerSetToken(string, long, string, string[])" />
+        /// </summary>
+        public void OauthBearerSetToken(string tokenValue, long lifetimeMs, string principalName, string[] extensions)
+            => kafkaHandle.OauthBearerSetToken(tokenValue, lifetimeMs, principalName, extensions);
+        
+        /// <summary>
+        ///     Refer to <see cref="Confluent.Kafka.IClient.OauthBearerSetTokenFailure(string)" />
+        /// </summary>
+        public void OauthBearerSetTokenFailure(string errstr)
+            => kafkaHandle.OauthBearerSetTokenFailure(errstr);
 
         /// <summary>
         ///     Refer to <see cref="Confluent.Kafka.IClient.Name" />
--- a/src/Confluent.Kafka/AdminClient.cs
+++ b/src/Confluent.Kafka/AdminClient.cs
@@ -502,9 +502,9 @@ namespace Confluent.Kafka
             => kafkaHandle.AddBrokers(brokers);
 
         /// <summary>
-        ///     Refer to <see cref="Confluent.Kafka.IClient.OauthBearerSetToken(string, long, string, string[])" />
+        ///     Refer to <see cref="Confluent.Kafka.IClient.OauthBearerSetToken(string, long, string, IDictionary{string, string})" />
         /// </summary>
-        public void OauthBearerSetToken(string tokenValue, long lifetimeMs, string principalName, string[] extensions)
+        public void OauthBearerSetToken(string tokenValue, long lifetimeMs, string principalName, IDictionary<string, string> extensions)
             => kafkaHandle.OauthBearerSetToken(tokenValue, lifetimeMs, principalName, extensions);
         
         /// <summary>
--- a/src/Confluent.Kafka/ConsumerBuilder.cs
+++ b/src/Confluent.Kafka/ConsumerBuilder.cs
@@ -211,7 +211,7 @@ namespace Confluent.Kafka
         /// <summary>
         /// Set SASL/OAUTHBEARER token refresh callback in provided conf object.
         /// The SASL/OAUTHBEARER token refresh callback is triggered via rd_kafka_poll() whenever OAUTHBEARER is the SASL mechanism and a token needs to be retrieved, typically based on the configuration defined in sasl.oauthbearer.config.
-        /// The callback should invoke <see cref="IClient.OauthBearerSetToken"/> or <see cref="IClient.OauthBearerSetToken"/> to indicate success or failure, respectively.
+        /// The callback should invoke <see cref="IClient.OauthBearerSetToken"/> or <see cref="IClient.OauthBearerSetTokenFailure"/> to indicate success or failure, respectively.
         ///
         /// An unsecured JWT refresh handler is provided by librdkafka for development and testing purposes,
         /// it is enabled by setting the enable.sasl.oauthbearer.unsecure.jwt property to true and is mutually exclusive to using a refresh callback.
--- a/src/Confluent.Kafka/ConsumerBuilder.cs
+++ b/src/Confluent.Kafka/ConsumerBuilder.cs
@@ -52,6 +52,11 @@ namespace Confluent.Kafka
         internal protected Action<IConsumer<TKey, TValue>, string> StatisticsHandler { get; set; }
 
         /// <summary>
+        ///     The configured OauthBearer Token Refresh handler.
+        /// </summary>
+        internal protected Action<IConsumer<TKey, TValue>, string> OauthBearerTokenRefreshHandler { get; set; }
+
+        /// <summary>
         ///     The configured key deserializer.
         /// </summary>
         internal protected IDeserializer<TKey> KeyDeserializer { get; set; }
@@ -78,7 +83,7 @@ namespace Confluent.Kafka
 
         internal Consumer<TKey,TValue>.Config ConstructBaseConfig(Consumer<TKey, TValue> consumer)
         {
-            return new Consumer<TKey,TValue>.Config
+            return new Consumer<TKey, TValue>.Config
             {
                 config = Config,
                 errorHandler = this.ErrorHandler == null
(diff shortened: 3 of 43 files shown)


thtp:
[clabot:check]

d-ulyanov:
Thanks for this PR @thtp! Waiting for this feature.
@mhowlett could we include it in the next release please?

src/Confluent.Kafka/ConsumerBuilder.cs
```diff
@@ -200,6 +208,29 @@ public ConsumerBuilder(IEnumerable<KeyValuePair<string, string>> config)
             return this;
         }
 
+        /// <summary>
+        /// Set SASL/OAUTHBEARER token refresh callback in provided conf object.
+        /// The SASL/OAUTHBEARER token refresh callback is triggered via rd_kafka_poll() whenever OAUTHBEARER is the SASL mechanism and a token needs to be retrieved, typically based on the configuration defined in sasl.oauthbearer.config.
```
edenhill, line 213: doc string must be updated to match .NET API rather than C.

src/Confluent.Kafka/ConsumerBuilder.cs
```diff
@@ -200,6 +208,29 @@ public ConsumerBuilder(IEnumerable<KeyValuePair<string, string>> config)
...
+        /// The callback should invoke <see cref="IClient.OauthBearerSetToken"/> or <see cref="IClient.OauthBearerSetToken"/> to indicate success or failure, respectively.
```
edenhill, line 214: the second one should be SetTokenFailure

src/Confluent.Kafka/IClient.cs
```diff
@@ -75,5 +75,31 @@ public interface IClient : IDisposable
...
+        /// <param name="lifetimeMs">when the token expires, in terms of the number of milliseconds since the epoch.</param>
+        /// <param name="principalName">the mandatory Kafka principal name associated with the token.</param>
+        /// <param name="extensions">
+        /// optional SASL extensions key-value array with extensions_size elements (number of keys * 2),
+        /// where [i] is the key and [i+1] is the key's value, to be communicated to the broker as additional key-value pairs during the initial client response as per https://tools.ietf.org/html/rfc7628#section-3.1.
+        /// The key-value pairs are copied.
+        /// </param>
+        /// 
+        /// <seealso cref="OauthBearerSetTokenFailure"/>
+        void OauthBearerSetToken(string tokenValue, long lifetimeMs, string principalName, string[] extensions);
```
edenhill, line 95: have extensions as an optional map instead

thtp:
> Missing tests (trivup can be used to set up an oauthbearer cluster)
@edenhill, is it correct that integration tests are not run CI builds, and supposed to be run manually on local environment? I mean, should I do something to ensure test environment is configured properly (some scripts or whatever), or I can just 'git push' test assuming there already is a configured oauthbearer cluster on localhost?

edenhill:
@mhowlett can you explain the integration test setup?

thtp:
> @mhowlett can you explain the integration test setup?

Well, I already have a cluster with configured oauth to play with. The question is, do I have to do something to set up a test cluster for CI?

thtp:
Oh, found docker-compose. It must be the answer )

thtp:
@edenhill Regarding integration tests, I have to ask again, because I’m not sure that I understood correctly.
Should I provide any sort of docker-compose files or scripts for trivup, or will the tests themselves be enough? If i must provide something beside tests, what exactly should it be?

edenhill:
@mhowlett Can you chime in on how integration test setups are done with the dotnet client?

src/Confluent.Kafka/ConsumerBuilder.cs
```diff
@@ -200,6 +208,30 @@ public ConsumerBuilder(IEnumerable<KeyValuePair<string, string>> config)
             return this;
         }
 
+        /// <summary>
+        /// Set SASL/OAUTHBEARER token refresh callback in provided conf object.
```
mhowlett, line 212: pedantic, but... can you follow indentation convention set in other methods

src/Confluent.Kafka/AdminClient.cs
```diff
@@ -501,6 +501,17 @@ public Metadata GetMetadata(string topic, TimeSpan timeout)
         public int AddBrokers(string brokers)
             => kafkaHandle.AddBrokers(brokers);
 
+        /// <summary>
+        ///     Refer to <see cref="Confluent.Kafka.IClient.OauthBearerSetToken(string, long, string, IDictionary{string, string})" />
+        /// </summary>
+        public void OauthBearerSetToken(string tokenValue, long lifetimeMs, string principalName, IDictionary<string, string> extensions)
```
mhowlett, line 507: I don't like having these methods exposed at the client class level because they aren't generally applicable - putting them here messys up the name space for all users of the clients, who mostly won't care about them. I think it would be better to pass in two additional delegate arguments into the refresh handler instead.

edenhill: The idea is that tokens can be set/failed at finer granularity than the refresh frequency, so these need to be exposed as standalone methods.

src/Confluent.Kafka/ConsumerBuilder.cs
mhowlett: also line length

src/Confluent.Kafka/AdminClient.cs
mhowlett: yes, but better if these methods are passed into the handler to limit their scope.

mhowlett: oh, that's annoying

mhowlett: is the requirement simply that a higher rate of refresh than the default assumed by librdkafka may be desired (which is 80% of lifetime)? or might the user want to do this on an ad-hoc basis? If it's the former, this could be achieved via configuration (and i like it better). if the latter, what's the scenario?

edenhill: Look at the original librdkafka PR for motivation.

mhowlett:
tests: we have a docker compose file for the schema registry tests. we don't currently have an equivalent for the Confluent.Kafka integration tests (note: I just run them locally against a locally running kafka, not in CI). I think making a docker compose file modeled after the schema registry one is what we want here.

mhowlett:
btw: thanks a lot for the PR @thtp - definitely something we've been needing to get done.

src/Confluent.Kafka/AdminClient.cs
mhowlett: that has 321 comments.. i think we can have a chat.

mhowlett: @thtp - i discussed with @edenhill and we agreed my suggestion is the more conservative. if it turns out the client level methods are required, we can retrofit easily. do you specifically require this yourself?

so on the builder classes, i'm thinking something like:

```
public ConsumerBuilder<TKey, TValue> SetOauthBearerTokenRefreshHandler(Action<IConsumer<TKey, TValue>, SetTokenDelegate, SetTokenErrorDelegate, string> oauthBearerTokenRefreshHandler)
```

(if we ever add oauth methods to the client we could overload this with a `SetOauthBearerTokenRefreshHandler` as you have it as well).

thtp: My main thoughts was that someone might have scenarios with reactive token revocation, or might need to update claims, or something like this. I mean, there might be reasons to update or revoke token besides its expiration.
Also, IClient wraps RdKafka HandleImpl, and oauthbearer_set_token is declared in than HandleImpl class. So it just feels natural to place it there (just mimicking original api).
The last one - just clean and easy to use method signature. Donno, I just like how it looks. Less code written, less delegates declaration, less ctrl+clicking around the solution.

Well, I bet you've discussed all that already...
So, delegates it be? Final decision? 146% sure? ) @mhowlett ?

p.s: I hope my attempts to express my thoughts in English are at least understandable )

mhowlett: i will think about this more, give me a day or two.

mhowlett: some thoughts:

> The last one - just clean and easy to use method signature. Donno, I just like how it looks. Less code written, less delegates declaration, less ctrl+clicking around the solution.

What you suggest is more straightforward / simpler for people to get, and that is absolutely a big advantage. What I proposed does add UX complexity for the sake of a cleaner separation of concerns. This is not always/necessarily a good tradeoff.

> My main thoughts was that someone might have scenarios with reactive token revocation, or might need to update claims, or something like that. I mean, there might be reasons to update or revoke token besides its expiration.

My assumption was this isn't a super important capability because AFAICT, the java implementation doesn't allow for this (am i missing something here?). that's not to say it isn't useful though. The scheme i proposed does actually allow for this, since the delegate instance lifetime can extend beyond that of the handler. not very good UX (but if not commonly needed, not something to optimise for).

> Also, IClient wraps RdKafka HandleImpl, and oauthbearer_set_token is declared in than HandleImpl class. So it just feels natural to place it there (just mimicking original api).

It's not an is-a relationship.

src/Confluent.Kafka/AdminClient.cs
```diff
@@ -501,6 +501,17 @@ public Metadata GetMetadata(string topic, TimeSpan timeout)
...
+            => kafkaHandle.OauthBearerSetToken(tokenValue, lifetimeMs, principalName, extensions);
+        
+        /// <summary>
+        ///     Refer to <see cref="Confluent.Kafka.IClient.OauthBearerSetTokenFailure(string)" />
+        /// </summary>
+        public void OauthBearerSetTokenFailure(string errstr)
```
mhowlett, line 513: i think the param name should probably be just `error`, `errstr` isn't very idiomatic c#

src/Confluent.Kafka/AdminClient.cs
mhowlett: Wondered if extension methods can help. Maybe, but it's not very clean.

First you'd have:

```
namespace Confluent.Kafka.Oauthbearer
{
    public static class OauthBearerProducerExtensions
    {
        public static void OauthBearerSetToken<TKey, TValue>(this IProducer<TKey, TValue> producer, string tokenValue, long lifetimeMs, string principalName, IDictionary<string, string> extensions)
        {
            // ...
        }

        public static void OauthBearerSetTokenFailure<TKey, TValue>(this IProducer<TKey, TValue> producer, string error)
        {
            // ...
        }
    }
```

etc.

the client instances have `Handle`, which have a `SafeKafkaHandle`. However you can't get at it in the extension method without using reflection because it's internal. You could use reflection, or define a helper class as follows (which can access internal properties) and use that instead. I think it'd probably go reflection, but it feels a bit horrible.

```
    public static class OauthbearerUtils
    {
        public static void OauthBearerSetToken<TKey, TValue>(Handle handle, string tokenValue, long lifetimeMs, string principalName, IDictionary<string, string> extensions)
        {
            // ...
        }

        public static void OauthBearerSetTokenFailure(Handle handle, string error)
        {
            // ...
        }
    }
}
```

Then we have the problem of how to specify the refresh handler on the builder classes. Again, we have the problem of `internal` properties. i think you could get around that somewhat by having an extension method on the `ProducerBuilder` etc. for adding the refresh handler, which returns `OauthBearerProducerBuilder`, but this a bit of a hack and isn't great because the same trick won't work if we want extensibility for something else. Instead, I think we'd need a general public method on the builder classes: `SetHandler(string name, object handler)`, but this is not very good / overly complex as well.

mhowlett: current thinking:

1. extension methods don't help.
2. i don't like either approach.
3. i do still currently prefer what i suggested, but i'm not very confident about it.

please bear with me a bit longer ...

mhowlett: another option is for the handler to return an object which represents either a valid token or an error. 

```
public ConsumerBuilder<TKey, TValue> SetOAuthBearerTokenRefreshHandler(Func<IConsumer<TKey, TValue>, string, OAuthBearerRefreshResult> oAuthBearerTokenRefreshHandler)

OAuthBearerToken : OAuthBearerRefreshResult
OAuthBearerError: : OAuthBearerRefreshResult
```

this doesn't allow for reactive token renewal / cancellation, but it's a step down in UX complexity. It's also arguably an improvement since it forces the behavior in the handler that is required - a single action to set the token or set the token error.

also, there is nothing preventing us from expanding the capability in the future if required - this doesn't cut off any avenues for extensibility.

this is currently my favourite option.

src/ConfigGen/Program.cs
```diff
@@ -102,6 +102,11 @@ public enum SaslMechanism
         ///     SCRAM-SHA-512
         /// </summary>
         ScramSha512
+
+        /// <summary>
+        ///     OAUTHBEARER
+        /// </summary>
+        OauthBearer
```
mhowlett, line 109: should be `OAuthBearer` to follow C# naming conventions. Also similar changes elsewhere.

src/Confluent.Kafka/AdminClient.cs
edenhill: I think we'll want to be cautious in having one of the librdkafka-based clients do things differently from the others, and only do so on solid grounds.
In this particular case I don't see any reason, other than aesthetics, from deviating from librdkafka, and in fact it reduces functionality of a feature that by nature requires a lot of flexibility.

thtp: Actually, extensions might be a good solution. I mean, extensions might be provided by this library. So, internal methods will not be a problem, we can leave IClient without any mods, and still have an ability to set token using IClient instance.

thtp:
> tests: we have a docker compose file for the schema registry tests. we don't currently have an equivalent for the Confluent.Kafka integration tests (note: I just run them locally against a locally running kafka, not in CI). I think making a docker compose file modeled after the schema registry one is what we want here.

My bet was on the "default, builtin AuthenticateCallbackHandler implementation" promised as default token validation mechanism for OAuthBearerLoginModule. I beleave, unsecured JWT is exactly what we need here. But I have no success in getting it work. I mean, all my attempts to start a broker with unsecured JWT are failng. Maybe you know someone who can help with creating a docker-compose for a Kafka broker that will use unsecured JWT? @mhowlett ?

P.S.: I Can provide a docker-compose with keycloak as an identity server ) That will work, but I'm not sure that it is a good solution. I mean, we well have to have a code to get an actual token just for couple of tests. And those tests will also test a keycloak which is completely out of a scope.

edenhill:
Highly recommend using trivup to create test clusters, for oauthbearer it should be as simple as:
```bash
$ pip install trivup
..

$ python -m trivup.clusters.KafkaCluster --sasl OAUTHBEARER --version 2.5.0
# Interactive mode
# - Waiting for cluster to go operational in /home/maglun/old/kafka/KafkaCluster/90659419
# - Client configuration in /home/maglun/old/kafka/KafkaCluster/90659419/rdkafka.conf
# - Connect to cluster with bootstrap.servers SASL_PLAINTEXT://localhost:39809,SASL_PLAINTEXT://localhost:38125,SASL_PLAINTEXT://localhost:35643
# - You're now in an interactive sub-shell, type 'exit' to stop the cluster and exit back to your shell.

To run a command as administrator (user "root"), use "sudo <command>".
See "man sudo_root" for details.

$ cat $RDKAFKA_TEST_CONF 
bootstrap.servers=SASL_PLAINTEXT://localhost:39809,SASL_PLAINTEXT://localhost:38125,SASL_PLAINTEXT://localhost:35643
broker.address.family=v4
security.protocol=SASL_PLAINTEXT
sasl.mechanism=OAUTHBEARER
enable.sasl.oauthbearer.unsecure.jwt=True
sasl.oauthbearer.config=scope=requiredScope principal=admin

$ run-your-tests
...

# terminate cluster
$ exit
```

edenhill:
There should not be a default OAUTHBEARER handler in any of the clients, it does not make sense since all oauthbearer deployments will be different, and we don't want to use the unsecure default of JWT.

thtp:
Default handlers are documented [here](http://kafka.apache.org/20/javadoc/index.html?org/apache/kafka/common/security/oauthbearer/OAuthBearerLoginModule.html). Since it is an official feature designed for testing and development, I don't see any contraindications in using it for testing and development ) Furthermore, using that feature will keep us from testing anything but what should be tested. The problem is - i haven't yet figured out how to set up broker to use it properly. Or, may be, I just misunderstood what are all those default handlers about?

edenhill:
I misremembered the default stance, you're right.

src/Confluent.Kafka/AdminClient.cs
mhowlett: @thtp i was excited by extension methods too. unfortunately it turns out they don't have access to internal members :-(. we could use reflection in both cases to get access to internal members. this is a bit horrible, but the user is not aware it's happening.

@edenhill - take your point. I think it too, i'm pretty torn on this.

(currently letting this sit for a bit - some chance what is right will present itself with certainty).

mhowlett:
we only need basic testing of this stuff i think in the .net client. trivup may indeed be what we ultimately want to use for integration tests, but we don't currently and if we can avoid locking that in now that would be good imo.

mhowlett:
@thtp - unfortunately i've never set this up before, so can't give any wisdom ...

src/Confluent.Kafka/AdminClient.cs
mhowlett: ok, i'm coming around to thinking as is wins (80% sure). let's let it sit another day though.

thtp: @mhowlett check out extensions implementation. Looks like perfect compromise ) (will revert if it's bulls..t)

mhowlett: interesting. i didn't think that would compile (based on what i was reading), but it does. we'll need to do the same trick on the builder classes, because it doesn't make sense to be able to set the refresh handler if the methods don't exist on the clients to implement them...

it is seeming good, yes. i think it's worth implementing those builder class changes (i look forward to taking it for a spin and seeing how it feels).

note: I do think it's worth dwelling on these things for a bit, because once you decide on something you're stuck with it and sometimes seemingly insignificant structural decisions can end up ballooning into much bigger things in the future.

thtp: Well, modifying the builders might be a little bit much more difficult... 
I mean, builders does not have a reference to a Handle, so we can't call SetWhateverHandler in a builder extension. We have to store delegate somewhere before it can be called on a Handler instance. Extension properties does not make it to the C#8, so the best we can do is storing delegates as a Dictionary<object, object> in the static field of an extension class. And we still will need to call a method to set those handlers in the ctor of the client.
Another way is to make it an extension on IClient. So the user will have to create an instance of a client with builder, and then on that instance call SetWhateverHandler().
Both options look weird.
To make it less weird we will have to change builders dramatically. It might become just a chane of Action<TClient> delegates that will initialize  an instance of a client or something like that, whatever... Anyway, it have to be changed if you want to remove any knowledge about all the handlers from a builders.

mhowlett: the builders could have an internal OAuthBearerTokenRefreshHandler property, but there's no way to set it without the extension method is what i was thinking

thtp:
Okay, now I can confirm that my simple produce-consume test is working with a trivup setup provided above. So, if docker-compose is not mandatory, I may say that tests status might be considered as 'ready for review' )
Really, how badly do you want me to prepare docker-compose? I've already spent a plenty of time trying, and almost lost my willing to live ) I afraid, one more evening is all I can stand )

mhowlett:
@thtp - thanks very much for trying :-). what you've done is fine by me, but leave the oauthbearer bootstrap servers config blank (it changes every time).

i actually feel that regression testing of this feature is of fairly low value (compared to other tests that don't exist, and would be easier to create so have roi, so we should be doing those instead) because regressions are unlikely to occur in the C# layer. but it's good to have the test committed, so we can enable it when the testing framework gets better.

thtp:
Is there anything else I should change? Or can I expect  these changes in some of the nearest releases? )

mhowlett:
i would like to get this in v1.5, which is coming pretty soon. the API is still causing me to pause though. First, I feel sure we shouldn't have the extension method on the client classes and not on the builder classes (the current state). Using extension methods on internal properties/methods actually feels like a bit of a mis-use of them to me in general, but i may be wrong, and it is quite good UX if we can make this work on the builder classes as well. Are you sure it's not possible (am i wrong below?):

> the builders could have an internal OAuthBearerTokenRefreshHandler property, but there's no way to set it without the extension method is what i was thinking

So that's the first option, if possible. The second option is how you originally had it (which may be the most pragmatic, but i'm still not very fond of it). And the third option is using return values from the refresh method.

Something I feel i want to know is how java does it, and if anyone finds this deficient.

Finally, we could always mark the API as experimental (giving us the option to change our minds later, but get the feature out now). Good to avoid it, but we can do it if we must.

thtp:
> the builders could have an internal OAuthBearerTokenRefreshHandler property, but there's no way to set it without the extension method is what i was thinking

I don't think it's actually a good idea. It looks like such an implementation violates encapsulation and does not add any value. There might be some reason if we will be able to remove handler from the builder completely. But without complete rewriting, it will only add unnecessary complexity (imo).

> Something I feel i want to know is how java does it, and if anyone finds this deficient.

As far as I understand, java client uses the same approach, as a server configuration. Setting a handler implementation class name to the corresponding client configuration property. Like `sasl.whatever.handler=some.java.namespace.WhatevrHandlerImplemetation` in the client config file. But that's not for sure.

thtp:
Any news? Something? Anything?

mhowlett:
the use of extension methods on the client classes violates encapsulation in the same way as the builder classes (which is why i say using extension methods seems like misusing them) and doing the extension method trick on the client and not the builder classes seems inconsistent to me.

the java client doesn't have any additional methods on the client classes. my current status on this pr is i need to find time to get to the bottom of that.

mhowlett:
believe it or not (!), LGTM, rebasing to 1.5.x and merging.

A few minor formatting things I'll change in a follow up PR. Release forthcoming.

PR #1297, opened 2020-05-25
Initial custom partitioner work

ChadJessup:
Initial stab at Custom Partitioners.
Partitioners are added on via the `ProducerBuilder`'s `SetPartitioner` method.
Custom partitioners are inherited from the new `IPartitioner` interface.

Basic unit tests, and integration with Benchmark project as well (add `use-partitioner` to benchmark run args).

This is really just locking down an API at this point, other than test/benchmark partitioners, no custom logic is created.

Prior PR on this had an open question to @mhowlett about what overloads we should allow on the `SetPartitioner` - some options below, first option is what's in this PR.

```
ProducerBuilder<K, V>()
.SetPartitioner("Topic1", new CustomPartitioner()) <- easier for caller, since might be a large chain, or instance shared
.SetPartitioner("Topic2", (string topic, K key, V value, partition_cnt, byte[] topic_opaque, byte[] msg_opaque)) => return 2)
.SetPartitioner("^Topic3|Topic4", Confluent.Kafka.Partitioner.Murmur2)
```


src/Confluent.Kafka/IPartitioner.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// <seealso cref="Confluent.Kafka.Partition"/> is set (that is, <seealso cref="Confluent.Kafka.Partition.Any"/>).
+    /// </summary>
+    public interface IPartitioner : IDisposable
+    {
+        /// <summary>
+        /// This method may be called in any thread at any time,
+        /// it may be called multiple times for the same message/key.
+        /// 
+        /// Partitioner function constraints:
+        ///   - MUST NOT call any rd_kafka_*() functions except:
```
edenhill, line 32: Change doc-string to cover .NET methods, not C

src/Confluent.Kafka/IPartitioner.cs
```diff
@@ -0,0 +1,48 @@
...
+        ///       rd_kafka_topic_partition_available()
+        ///   - MUST NOT block or execute for prolonged periods of time.
+        ///   - MUST return a value between 0 and partition_cnt-1, or the
+        ///     special <seealso cref="Confluent.Kafka.Partition.Any"/> value if partitioning
+        ///     could not be performed.
+        /// </summary>
+        /// <param name="topic">The topic to partition for.</param>
+        /// <param name="keydata">Pointer to the Key's data.</param>
+        /// <param name="keylen">The lenght of the Key's data.</param>
```
edenhill, line 41: typo

src/Confluent.Kafka/IPartitioner.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <param name="partition_cnt">The known count of known partitions for the Topic.</param>
+        /// <param name="rkt_opaque">The rkt_opaque argument is the opaque set by rd_kafka_topic_conf_set_opaque().</param>
```
edenhill, line 43: Should not be exposed to .NET

src/Confluent.Kafka/IPartitioner.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// <param name="msg_opaque">The msg_opaque argument is the per-message opaque passed to produce().</param>
```
edenhill, line 44: Should not be exposed to .NET, but some other way to pass an arbitrary object to the partitioner per message would be good.

ChadJessup:
@mhowlett  - to avoid a lot of noise, I was planning on bundling changes from your future feedback and @edenhill 's existing feedback.
Would you rather I do this in two passes?

This is also a little poke. :)

julesbovet:
Hello @ChadJessup, do you plan on continuing working on this feature? We're very interested by a custom partitioner support at my company, so I'd like to know if you are just waiting on mhowlett feedback, or if you'd like us to pick up the work on this.
Also @mhowlett, are you waiting on updates from ChadJessup? Just making sure there's no deadlock here :)

ChadJessup:
> Hello @ChadJessup, do you plan on continuing working on this feature? We're very interested by a custom partitioner support at my company, so I'd like to know if you are just waiting on mhowlett feedback, or if you'd like us to pick up the work on this.
> Also @mhowlett, are you waiting on updates from ChadJessup? Just making sure there's no deadlock here :)

@julesbovet - sorry for the delay on getting back to you.
I honestly have no idea what the status of this is. This is the second of these (same code, just went stale) that hasn't been looked at in much capacity, this is after moving it to the new version.

I have a feeling it's waiting until the work lines up with some timeline, or the state of the world gets less crazy?

mhowlett:
i'm expecting to get to this in the next few weeks.

mhowlett:
> i'm expecting to get to this in the next few weeks.

... is what I said 20 days ago! 

@ChadJessup - there are two ways we can do this: 1) I can take your branch, add commits / change it to my liking then get a code review by @edenhill. if we do this you will show up as joint contributor, and you don't need to do anything more 2) the two of us can iterate, me giving comments and you doing all the commits. that way you'll show up as sole contributor. this process is typically slower. 

what's your preference?

ChadJessup:
> > i'm expecting to get to this in the next few weeks.
> 
> ... is what I said 20 days ago!
> 
> @ChadJessup - there are two ways we can do this: 1) I can take your branch, add commits / change it to my liking then get a code review by @edenhill. if we do this you will show up as joint contributor, and you don't need to do anything more 2) the two of us can iterate, me giving comments and you doing all the commits. that way you'll show up as sole contributor. this process is typically slower.
> 
> what's your preference?

Hi @mhowlett! I won't have time for a bit, but if you're cool running with what's here - feel free to take over.

Can I be tagged on the final code review as well? I'd like to just see how things end up. :)

mhowlett:
sure - will tag you in the final review. i'll also comment here if i'm deviating significantly.

mhowlett:
continuing over in #1436

PR #1311, opened 2020-06-05
Test well known protobuf schema refs

mhowlett:


test/Confluent.SchemaRegistry.Serdes.IntegrationTests/Tests_Protobuf/Protobuf/compile_linux.sh
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 ~/.nuget/packages/grpc.tools/1.16.0/tools/linux_x64/protoc \
-    --proto_path=.\
+    --proto_path=. \
+    --proto_path=/usr/local/include \
```
edenhill, line 5: Since /usr/local/ is for manually installed things (rather than proper distro packages) it would be good to point out what it expects to find here and how to install it.

PR #1344, opened 2020-07-03
progress on nullable avro values [WIP, not for review]

mhowlett:
I just set out to add support for nullable avro values. unfortunately this is a little awkward because the generic type on the serdes can currently be int, bool etc (this is in fact explicitly supported) which aren't nullable. so we can't return a null value from the deserializer - the type system in C# isn't powerful enough. the best solution is probably to discontinue support for int, bool, and instead provide support for int?, bool?. This is all good except it's not backwards compatible (changes to code required). We could alternatively provide different serdes side by side.

this issues does not apply to the json and protobuf serdes, they can handle null values.

### DIFF
--- a/src/Confluent.SchemaRegistry.Serdes.Avro/AvroDeserializer.cs
+++ b/src/Confluent.SchemaRegistry.Serdes.Avro/AvroDeserializer.cs
@@ -35,7 +35,7 @@ namespace Confluent.SchemaRegistry.Serdes
     ///       bytes 1-4:        Unique global id of the Avro schema that was used for encoding (as registered in Confluent Schema Registry), big endian.
     ///       following bytes:  The serialized data.
     /// </remarks>
-    public class AvroDeserializer<T> : IAsyncDeserializer<T>
+    public class AvroDeserializer<T> : IAsyncDeserializer<T> where T : class
     {
         private IAvroDeserializerImpl<T> deserializerImpl;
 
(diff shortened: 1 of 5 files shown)


src/Confluent.SchemaRegistry.Serdes.Avro/AvroDeserializer.cs
```diff
@@ -35,7 +35,7 @@ namespace Confluent.SchemaRegistry.Serdes
     ///       bytes 1-4:        Unique global id of the Avro schema that was used for encoding (as registered in Confluent Schema Registry), big endian.
     ///       following bytes:  The serialized data.
     /// </remarks>
-    public class AvroDeserializer<T> : IAsyncDeserializer<T>
+    public class AvroDeserializer<T> : IAsyncDeserializer<T> where T : class
```
edenhill, line 38: Assuming `class` implies that the type is nullable?

mhowlett: yes. but this won't compile because we explicitly handle the `int` etc. cases. even if we didn't, it's not a backwards compatible change. it's likely people rely on this, since we talk about it in places, and I don't think we can change to `int?` until a major version bump. straightforward for people to update to, but it'll break things.

edenhill: How about providing a `ProduceTombstone(key)` and a `Message.IsTombstone()` ?

mhowlett: i don't think we want to go adding to the core client APIs just to get around this - would prefer to add a second set of avro serdes.

edenhill: Tombstones/compaction is a core Kafka feature, so I believe this abstraction should indeed be in the core clients rather than the serdes.
Tombstones being represented as Null is a protocol implementation detail and not necessarily something that needs to be exposed the same way to the application.

mhowlett: true, but we'd be exposing the same information redundantly (which I like less).

also, this does not actually do anything to address the problem which is at the serde level, not the client level.

Issue #905, opened 2019-04-24
Cannot produce / consume tombstone message


michael-huxtable:
Description
===========

You cannot produce/consume a `null` message with the `AvroSerializer`, to a produce a `null` message then a specific producer needs to be used with the value serializer set as `Serializers.Null`.

So far using the `AvroDeserializer` , I have not found a way to consume these `null` messages, currently when doing this I get the following exception:

```
Confluent.Kafka.ConsumeException: Local: Value deserialization error ---> System.IO.EndOfStreamException: Unable to read beyond the end of the stream.
   at System.IO.BinaryReader.ReadByte()
   at Confluent.SchemaRegistry.Serdes.SpecificDeserializerImpl`1.Deserialize(String topic, Byte[] array)
   at Confluent.SchemaRegistry.Serdes.AvroDeserializer`1.DeserializeAsync(ReadOnlyMemory`1 data, Boolean isNull, SerializationContext context)
   at Confluent.Kafka.Consumer`2.<>c__DisplayClass71_0.<<ConsumeViaBytes>b__1>d.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at Confluent.Kafka.Consumer`2.ConsumeViaBytes(Int32 millisecondsTimeout)
   --- End of inner exception stack trace ---
```

How to reproduce
================

Use a specific producer for building tombstone messages:

```
var tombstoner = new ProducerBuilder<int, Null>(_kafkaConfiguration.ProducerConfiguration)
	.SetKeySerializer(new AvroSerializer<int>(_schemaRegistryClient))
	.SetValueSerializer(Serializers.Null)
	.Build();

var tasks = properties.Select(property => tombstoner.ProduceAsync("test-topic",
	new Message<int, Null>
	{
		Key = 1,
		Value = null,
		Timestamp = Timestamp.Default
	}));
```

The consumer code is as the examples on the GitHub homepage, it's currently the simple consumer example. Single partition, etc. The other non-null messages get handled fine. A `ConsumeException` is thrown for the `null` messages.

Checklist
=========

Please provide the following information:
 - [Y] A complete (i.e. we can run it), minimal program demonstrating the problem. No need to supply a project file.
 - [1.0.0-RC3] Confluent.Kafka nuget version.
 - [Kafka 2.0.1-L0 (Landoop)] Apache Kafka version.
 - [N] Client configuration.
 - [Windows 10 x64] Operating system.
[...]

PR #1352, opened 2020-07-15
Support serializing and deserializing null in Avro serdes

YairHalberstadt:
Fixes https://github.com/confluentinc/confluent-kafka-dotnet/issues/905

Brings this up to par with the java implementation.
One limitation is that you cannot produce or consume nulls of value type primitives.


timtebeek:
@mhowlett Could you weigh in here? Would love to see this merged & released for our landscape of mixed Java/.Net services, so we can use the same patterns of tombstone records throughout.

kuduh:
@mhowlett When will the error of serialization and deserialization of null in Avro serdes be fixed?

src/Confluent.SchemaRegistry.Serdes.Avro/AvroDeserializer.cs
```diff
@@ -92,6 +92,19 @@ public async Task<T> DeserializeAsync(ReadOnlyMemory<byte> data, bool isNull, Se
         {
             try
             {
+                if (isNull)
+                {
+                    // Jit optimizes this check away
+                    if (default(T) == null)
```
mhowlett, line 98: nice

mhowlett:
Thanks @YairHalberstadt - sorry for the delay. This is looking good to me - the trick you're playing with `default` is something I think I didn't think of when I went to do this. I want to spend a bit of time playing with it to make sure though before merging. Targeting v1.9.0.

mhowlett (approved):
Great work, thanks for doing this. It'll be v1.9.0

ahmar-husain:
@mhowlett when v.1.9.0 will be available?

PR #1357, opened 2020-07-20
-> v1.5.0

mhowlett:
The watermark integration test is actually failing here: https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/test/Confluent.Kafka.IntegrationTests/Tests/WatermarkOffsets.cs#L62

This could be because the low wm offsets are automatically cached now (seems probable), or possibly an initialization error. do you recall making changes here?


edenhill:
I think the test is buggy, since it consumes the partition there will be a cached low offset.

mhowlett:
fixed - interesting thing is the behavior is new in v1.5.

PR #1376, opened 2020-08-09
Latency benchmark test

mhowlett:
Try not to be too picky on this - more important things to be focused on. Just trying to add some value quickly for something I needed on Friday.

### DIFF
--- a/test/Confluent.Kafka.Benchmark/Latency.cs
+++ b/test/Confluent.Kafka.Benchmark/Latency.cs
@@ -1,0 +1,193 @@
+// Copyright 2020 Confluent Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


test/Confluent.Kafka.Benchmark/Latency.cs
```diff
@@ -0,0 +1,193 @@
...
+                throw new Exception("Message rate must be >= 1000 msg/s");
+            }
+
+            var sw = new Stopwatch();
+            sw.Start();
+
+            var monitorObj = new Object();
+
+            var consumerTask = Task.Run(() => {
+                // Use middle results only to better estimate steady state performance.
```
edenhill, line 47: Second half, rather than middle, no?

edenhill (approved):

PR #1377, opened 2020-08-09
Added exception for SpecificRecord serializer/deserializer in case type has no _SCHEMA field.

BillyFromAHill:


src/Confluent.SchemaRegistry.Serdes.Avro/SpecificDeserializerImpl.cs
```diff
@@ -53,7 +53,13 @@ public SpecificDeserializerImpl(ISchemaRegistryClient schemaRegistryClient)
 
             if (typeof(ISpecificRecord).IsAssignableFrom(typeof(T)))
             {
-                ReaderSchema = (global::Avro.Schema)typeof(T).GetField("_SCHEMA", BindingFlags.Public | BindingFlags.Static).GetValue(null);
+                var schemaField = typeof(T).GetField("_SCHEMA", BindingFlags.Public | BindingFlags.Static);
+                if (schemaField is null)
+                {
+                    throw new ArgumentException("Deserializable type must have public static field with name '_SCHEMA'.");
```
mhowlett, line 59: I think we want more information, maybe something like: `($"ISpecificRecord implementation '{}' ...", typeof(T).FullName)` or something like that? The call stack will indicate this happened during deserialization, though that could also be in the message.

src/Confluent.SchemaRegistry.Serdes.Avro/SpecificSerializerImpl.cs
```diff
@@ -67,7 +67,13 @@ internal class SpecificSerializerImpl<T> : IAvroSerializerImpl<T>
             Type writerType = typeof(T);
             if (typeof(ISpecificRecord).IsAssignableFrom(writerType))
             {
-                writerSchema = (global::Avro.Schema)typeof(T).GetField("_SCHEMA", BindingFlags.Public | BindingFlags.Static).GetValue(null);
+                var schemaField = typeof(T).GetField("_SCHEMA", BindingFlags.Public | BindingFlags.Static);
+                if (schemaField is null)
+                {
+                    throw new ArgumentException("Serializable type must have public static field with name '_SCHEMA'.");
```
mhowlett, line 73: ditto

mhowlett:
LGTM after exception message changes - thanks @BillyFromAHill !

src/Confluent.SchemaRegistry.Serdes.Avro/SpecificDeserializerImpl.cs
BillyFromAHill: Done

src/Confluent.SchemaRegistry.Serdes.Avro/SpecificSerializerImpl.cs
BillyFromAHill: Done

PR #1378, opened 2020-08-09
Increb changes

mhowlett:


mhowlett:
this is ready for a review

src/Confluent.Kafka/Consumer.cs
```diff
@@ -177,37 +179,85 @@ private void LogCallback(IntPtr rk, SyslogLevel level, string fac, string buf)
...
+                            throw new InvalidOperationException("(Incremental)Assign/Unassign must not be called in the partitions assigned handler");
                         }
                     }
-                    Assign(assignTo);
+
+                    if (kafkaHandle.RebalanceProtocol == "COOPERATIVE")
+                    {
+                        if (assignTo.Count() != partitions.Count())
+                        {
+                            throw new InvalidOperationException("The partitions assigned handler must not return a different set of topic partitions than it was provided");
```
edenhill, line 219: Do we really have to enforce this?
Some applications want to consume the subscribed topics and a common command topic/partition that all consumers read.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -177,37 +179,85 @@ private void LogCallback(IntPtr rk, SyslogLevel level, string fac, string buf)
...
+                        }
+
+                        int i = 0;
+                        foreach (var p in assignTo)
+                        {
+                            if (p.TopicPartition != partitions[i++])
```
edenhill, line 225: Are these two always sorted? If not, sort them.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -220,19 +270,31 @@ private void LogCallback(IntPtr rk, SyslogLevel level, string fac, string buf)
...
-                    if (assignTo.Count() > 0) { Assign(assignTo); }
-                    else { Unassign(); }
+                    if (kafkaHandle.RebalanceProtocol == "COOPERATIVE")
+                    {
+                        // assignTo is always empty, not used in the COOPERATIVE case.
+                        IncrementalUnassign(partitions);
+                    }
+                    else
+                    {
+                        // This distinction is important because calling Assign whilst the consumer is being
```
edenhill, line 293: It is actually okay now, we just call unassign() internally.

src/Confluent.Kafka/ConsumerBuilder.cs
```diff
@@ -275,27 +290,41 @@ public ConsumerBuilder(IEnumerable<KeyValuePair<string, string>> config)
...
-        ///     special value (Beginning, End or Unset). If Unset, consumption will resume from the
-        ///     last committed offset for each partition, or if there is no committed offset, in accordance
-        ///     with the `auto.offset.reset` configuration property.
+        ///     of the handler. Partition offsets may be a specific offset, or special value (Beginning, End
+        ///     or Unset). If Unset, consumption will resume from the last committed offset for each
+        ///     partition, or if there is no committed offset, in accordance with the `auto.offset.reset`
+        ///     configuration property.
+        ///
+        ///     Kafka supports two rebalance protocols: EAGER (range and roundrobin assignors) and
+        ///     COOPERATIVE ("incremental") (sticky-cooperative assignor).
```
edenhill, line 303: cooperative-sticky

src/Confluent.Kafka/ConsumerBuilder.cs
```diff
@@ -275,27 +290,41 @@ public ConsumerBuilder(IEnumerable<KeyValuePair<string, string>> config)
...
+        ///
+        ///     ## EAGER Rebalancing
```
edenhill, line 305: Put (range, roundrobin) in the header too, for clarity

src/Confluent.Kafka/ConsumerBuilder.cs
```diff
@@ -275,27 +290,41 @@ public ConsumerBuilder(IEnumerable<KeyValuePair<string, string>> config)
...
-        ///     by the consumer group, but typically will. Partition offsets may be a specific offset, or
-        ///     special value (Beginning, End or Unset). If Unset, consumption will resume from the
-        ///     last committed offset for each partition, or if there is no committed offset, in accordance
-        ///     with the `auto.offset.reset` configuration property.
+        ///     of the handler. Partition offsets may be a specific offset, or special value (Beginning, End
+        ///     or Unset). If Unset, consumption will resume from the last committed offset for each
+        ///     partition, or if there is no committed offset, in accordance with the `auto.offset.reset`
+        ///     configuration property.
+        ///
+        ///     Kafka supports two rebalance protocols: EAGER (range and roundrobin assignors) and
```
edenhill, line 302: Should mention how the strategy is configured

src/Confluent.Kafka/ConsumerBuilder.cs
```diff
@@ -275,27 +290,41 @@ public ConsumerBuilder(IEnumerable<KeyValuePair<string, string>> config)
...
+        ///     The set of partitions returned from your handler may differ from that provided by the
+        ///     group (though they should typically be the same). These partitions are the
+        ///     entire set of partitions to consume from. There will be exactly one call to the
+        ///     partitions revoked or partitions lost handler (if they have been set using
+        ///     SetPartitionsRevokedHandler / SetPartitionsLostHandler) corresponding to every call to
+        ///     this handler.
+        ///
+        ///     ## COOPERATIVE (Incremental) Rebalancing
+        ///
+        ///     The set of partitions returned from your handler must match that provided by the
```
edenhill, line 316: Don't like this constraint, is there an actual reason for it?

src/Confluent.Kafka/ConsumerBuilder.cs
```diff
@@ -275,27 +290,41 @@ public ConsumerBuilder(IEnumerable<KeyValuePair<string, string>> config)
...
+        ///     group. These partitions are an incremental assignment - are in additional to those
```
edenhill, line 317: the stuff after the hyphen reads weird.
at least change additional -> addition

src/Confluent.Kafka/ConsumerBuilder.cs
```diff
@@ -311,23 +340,36 @@ public ConsumerBuilder(IEnumerable<KeyValuePair<string, string>> config)
...
-        ///     the partitions revoked handler (if one has been set using SetPartitionsRevokedHandler").
+        ///     Specify a handler that will be called when a new consumer group partition assignment has
+        ///     been received by this consumer.
+        ///
+        ///     Following execution of the handler, consumption will resume from the last committed offset
+        ///     for each partition, or if there is no committed offset, in accordance with the
+        ///     `auto.offset.reset` configuration property.
+        ///
+        ///     Kafka supports two rebalance protocols: EAGER (range and roundrobin assignors) and
+        ///     COOPERATIVE ("incremental") (sticky-cooperative assignor).
```
edenhill, line 351: cooperative-sticky

edenhill: drop the quotes

src/Confluent.Kafka/ConsumerBuilder.cs
```diff
@@ -311,23 +340,36 @@ public ConsumerBuilder(IEnumerable<KeyValuePair<string, string>> config)
...
+        ///     Partitions passed to the handler represent the entire set of partitions to consume from.
+        ///     There will be exactly one call to the partitions revoked or partitions lost handler (if
+        ///     they have been set using SetPartitionsRevokedHandler / SetPartitionsLostHandler)
+        ///     corresponding to every call to this handler.
         ///
-        ///     Consumption will resume from the last committed offset for each partition, or if there is
-        ///     no committed offset, in accordance with the `auto.offset.reset` configuration property.
+        ///     ## COOPERATIVE (Incremental) Rebalancing
+        ///
+        ///     Partitions passed to the handler are an incremental assignment - are in additional to those
```
edenhill, line 362: dito dito

src/Confluent.Kafka/IConsumer.cs
```diff
@@ -271,6 +271,89 @@ public interface IConsumer<TKey, TValue> : IClient
...
+        ///     to the current assignment.
+        /// </summary>
+        /// <param name="partition">
+        ///     The additional partition to consume from.
+        ///     Consumption will resume from the last committed
+        ///     offset, or according to the 'auto.offset.reset'
+        ///     configuration parameter if no offsets have been
+        ///     committed yet.
+        /// </param>
+        void IncrementalAssign(TopicPartition partition);
```
edenhill, line 285: Why do we need this variant?
I think it will trick people into calling IncrementalAssign() in a loop, rather than a single call, which is not the way the underlying C API is intended to be use (preferably one *assign call per rebalance).

src/Confluent.Kafka/IConsumer.cs
```diff
@@ -271,6 +271,89 @@ public interface IConsumer<TKey, TValue> : IClient
...
+
+        /// <summary>
+        ///     Incrementally remove <paramref name="partitions" />
+        ///     to the current assignment.
+        /// </summary>
+        /// <param name="partitions">
+        ///     The set of partitions to remove from the current
+        ///     assignment.
+        /// </param>
+        void IncrementalUnassign(IEnumerable<TopicPartition> partitions);
```
edenhill, line 343: dito

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -780,17 +783,47 @@ internal void Assign(IEnumerable<TopicPartitionOffset> partitions)
                 }
             }
 
-            ErrorCode err = Librdkafka.assign(handle, list);
+            ErrorCode err = ErrorCode.NoError;
+            Error error = null;
+
+            if (assignMethodErr != null) { err = assignMethodErr(handle, list); }
+            if (assignMethodError != null) { error = new Error(assignMethodError(handle, list)); }
```
edenhill, line 790: else if ?

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -780,17 +783,47 @@ internal void Assign(IEnumerable<TopicPartitionOffset> partitions)
...
+
             if (list != IntPtr.Zero)
             {
                 Librdkafka.topic_partition_list_destroy(list);
             }
-            if (err != ErrorCode.NoError)
+
+            if (err != ErrorCode.NoError) { throw new KafkaException(CreatePossiblyFatalError(err, null)); }
+            if (error != null && error.Code != ErrorCode.NoError) { throw new KafkaException(error); }
```
edenhill, line 798: error.Code will never be NoError

edenhill, line 798: else if

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -780,17 +783,47 @@ internal void Assign(IEnumerable<TopicPartitionOffset> partitions)
...
+                return (Librdkafka.assignment_lost(handle) != IntPtr.Zero);
             }
         }
 
+        internal string RebalanceProtocol
+        {
+            get
+            {
+                ThrowIfHandleClosed();
+                return Util.Marshal.PtrToStringUTF8(Librdkafka.rebalance_protocol(handle));
```
edenhill, line 824: It will return NULL if used after calling Close()

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_CooperativeRebalance_1.cs
```diff
@@ -0,0 +1,95 @@
...
+                // call should be incremental (checked in the handler).
+                consumer.Subscribe(new List<string> { topic1.Name, topic2.Name });
+
+                Util.ProduceNullStringMessages(bootstrapServers, topic2.Name, 1, 1);
+                var cr2 = consumer.Consume(TimeSpan.FromSeconds(10));
+                Assert.NotNull(cr2);
+                Assert.Single(cr2.Message.Value);
+
+                // Exceed MaxPollIntervalMs => lost partitions.
+                Thread.Sleep(8);
```
edenhill, line 82: Should also verify the counters prior to this sleep/maxpoll triggering.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_CooperativeRebalance_2.cs
```diff
@@ -0,0 +1,97 @@
...
+                }
+                catch (ConsumeException e)
+                {
+                    Assert.Equal(ErrorCode.Local_MaxPollExceeded, e.Error.Code);
+                }
+
+                // The second call should trigger the revoked handler (because a lost
+                // handler has not been specified).
+                consumer.Consume(TimeSpan.FromSeconds(1));
+                Assert.Equal(1, revokeCount);
```
edenhill, line 85: check revokeCount == 0 prior to consume()  (defensive programming/testing)

edenhill (changes requested):
Needs some negative tests, e.g., erroneous calls/usage.

edenhill:
Don't forget to update the examples.

src/Confluent.Kafka/Consumer.cs
mhowlett: I think so. The application logic in the cooperative case will get pretty complex, i think we just don't want to go there.

This API we have for realizing command topic/partition IMO is bad:

1. structurally, it should be represented to the user as a separate thing - the two things shouldn't be as interleaved as they are. it is not the clearest way of modelling the requirements.
2. the terminology is ambiguous / tricky and it makes the API harder to understand.
3. Java client does not allow this.
4. It makes the CGRP implementation harder! 

I think best practice currently for people who want this pattern is to just use two consumers. Ideally we'd be able to use a single instance for this (but I think two consumers is not a big deal for most users).

src/Confluent.Kafka/Consumer.cs
mhowlett: they are, provided the user doesn't change the ordering of what they are given if they choose to return a collection with different offsets. which usually they wouldn't. Ok, i'll sort the user returned collection. I guess only O(n) in the usual case, so no big deal.

src/Confluent.Kafka/Consumer.cs
mhowlett: changing this to just call `Unassign`.

src/Confluent.Kafka/ConsumerBuilder.cs
mhowlett: the reason is we don't want to encourage people to do it - the logic to track it isn't super hard, but it's getting a little involved, and it's just a bad API IMO.

src/Confluent.Kafka/IConsumer.cs
mhowlett: yeah, I agree with that. I was following the assign/unssign variants. i don't think there's a good reason to keep this

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
mhowlett: i've changed it but it's superfluous because execution can't get here if the first condition is true (exception thrown).

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
mhowlett: ... my point exactly

mhowlett: oh, above comment was meant for here. made else anyway here too.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
mhowlett: changed to return null in that case.

src/Confluent.Kafka/ConsumerBuilder.cs
mhowlett: because to realize a ctrl topic, they don't need to adjust the group assignment in the rebalance cb (it will happen outside rebalance_cb), and I hate that this is possible.

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
mhowlett: this logic is actually incorrect. if assignMethodError returns IntPtr.Zero, this should set error to null, not an Error object. i have no idea how the tests were working - i'm quite sure they were ...

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_CooperativeRebalance_1.cs
```diff
@@ -0,0 +1,103 @@
...
+
+                // Subscribe to a second one partition topic, the second assign
+                // call should be incremental (checked in the handler).
+                consumer.Subscribe(new List<string> { topic1.Name, topic2.Name });
+
+                Util.ProduceNullStringMessages(bootstrapServers, topic2.Name, 1, 1);
+                var cr2 = consumer.Consume(TimeSpan.FromSeconds(10));
+                Assert.NotNull(cr2);
+                Assert.Single(cr2.Message.Value);
+
```
edenhill, line 84: Call consume again with 1s timeout to make sure cr1 is not received again.

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_IncrementalAssign.cs
```diff
@@ -0,0 +1,94 @@
...
+                Assert.Equal(0, (int)cr4.Partition);
+                Assert.Equal(3, cr4.Message.Value.Length);
+
+                consumer.IncrementalAssign(new List<TopicPartition> { new TopicPartition(topic1.Name, 0) });
+                var cr5 = consumer.Consume(TimeSpan.FromSeconds(10));
+                Assert.NotNull(cr5);
+                Assert.Equal(1, cr5.Offset);
+                Assert.Equal(topic1.Name, cr5.Topic);
+                Assert.Equal(0, (int)cr5.Partition);
+
```
edenhill, line 85: Consume with short timeout to make sure no further messages are seen.

test/Confluent.Kafka.IntegrationTests/Tests/Transactions_Abort.cs
```diff
@@ -40,11 +40,16 @@ public void Transactions_Abort(string bootstrapServers)
             {
                 producer.InitTransactions(defaultTimeout);
                 producer.BeginTransaction();
-                producer.Produce(topic.Name, new Message<string, string> { Key = "test key 0", Value = "test val 0" });
-                Thread.Sleep(1000); // ensure the abort ctrl message makes it into the log.
+                producer.Produce(topic.Name, new Message<string, string> { Key = "test key 0", Value = "test val 0" }, (dr) => {
+                    Assert.Equal(0, dr.Offset);
+                });
+                Thread.Sleep(4000); // ensure the abort ctrl message makes it into the log.
```
edenhill, line 46: Shouldn't that be after the AbortTransaction call then?

edenhill (approved):
LGTM, left some small comments.

I'm thinking it might be a good idea to have a separate incremental assignment handler API to make sure applications truly support the new semantics and not only set the new strategy.

examples/Consumer/Program.cs
```diff
@@ -62,15 +63,25 @@ public static void Run_Consume(string brokerList, List<string> topics, Cancellat
                 .SetStatisticsHandler((_, json) => Console.WriteLine($"Statistics: {json}"))
                 .SetPartitionsAssignedHandler((c, partitions) =>
```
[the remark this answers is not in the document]
mhowlett: I like it because it protects the user against themselves. downsides: 1. `SetPartitionsIncrementallyAssignedHandler` or `SetPartitionsAssignedIncrementalHandler` or similar seem pretty awkward (better name?). This is the main thing that's detracting from it for me atm 2. The current API matches the Java API better. Not that we've let that stop us before.

I feel inclined to just leave it, but i like your thinking. I'll procrastinate on this for a bit.

PR #1383, opened 2020-08-16
Type schema creation refactoring.

BillyFromAHill:


mhowlett:
does this fix a problem? after a quick look, it looks like the suggested factoring is probably better, but every change comes with a risk of breaking something and I need to review 3rd party changes very carefully (so this PR will probably remain unactioned due to time prioritization). that said, thanks for the contribution!

BillyFromAHill:
> does this fix a problem? after a quick look, it looks like the suggested factoring is probably better, but every change comes with a risk of breaking something and I need to review 3rd party changes very carefully (so this PR will probably remain unactioned due to time prioritization). that said, thanks for the contribution!

This PR removes necessity of static _Schema field for the type and some of the copypaste around schema resolving. We needed it for our project, but in result we made some workarounds.

PR #1408, opened 2020-09-17
Move from DateTime to DateTimeOffset

gliljas:
DateTime is a weird type, and DateTimeOffset provides a much nicer experience. The PR is mainly about the Timestamp type, but I changed in other places as well. Perhaps the places where "Now" is used to keep track of elapsed time should use other constructs instead, such as Stopwatch or CancellationTokens with scheduled set time.

### DIFF
--- a/examples/AvroBlogExamples/Program.cs
+++ b/examples/AvroBlogExamples/Program.cs
@@ -122,7 +122,7 @@ namespace AvroBlogExample
                             var consumeResult = consumer.Consume(cts.Token);
 
                             Console.WriteLine(
-                                consumeResult.Message.Timestamp.UtcDateTime.ToString("yyyy-MM-dd HH:mm:ss")
+                                consumeResult.Message.Timestamp.DateTimeOffset.UtcDateTime.ToString("yyyy-MM-dd HH:mm:ss")
                                 + $": [{consumeResult.Message.Value.Severity}] {consumeResult.Message.Value.Message}");
                         }
                         catch (ConsumeException e)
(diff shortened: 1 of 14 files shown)


gliljas:
[clabot:check]

mhowlett:
The timestamp in the Kafka message format is just a long value which has a 1:1 correspondence with `DateTime` - using `DateTimeOffset` would be adding to the size of the data type without adding information. I think `DateTime` best satisfies the requirements here (which include memory / performance), except it is liable to misuse as you note.

Regardless of what is best, we are reluctant to make breaking changes to the API without very good reason as this impacts a lot of people, so we won't be changing this.

Thanks for the contribution though!

gliljas:
The PR doesn't introduce any breaking changes. The DateTime properties are still there, although marked as obsolete (which could be changed, of course). And it doesn't add to the size of the data type. It's just a masquerading property.

mhowlett:
The timestamp in Kafka has inherent ambiguity, in that it's just a long without a timezone, and this matches well to `DateTime` (deficiencies and all). We thought about DateTimeOffset when making the class, and decided against it. I agree there is an argument for it, but i'm not convinced it completely solves the problem or adds enough value (we also prefer a smaller API surface area over different ways to do the same thing).

The PR marks some methods as obsolete (meaning the intention is to make breaking changes when we do a major version bump).

gliljas:
Since the actual data is a UTC timestamp (I wouldn't call it ambiguous, or is it?), there will never be any timezone, or even an offset<>0, so we really don't get any added value functionality with DateTimeOffset, but the ambiguity (indeed reduced by the name "UtcDateTime") is what's so bad about DateTime, especially once it's relayed further through other API:s.

Anyway, you've thought about it and decided against it, so fair enough :)

mhowlett:
again, thanks for the input - if we get the same feedback a lot, we'll reconsider.

PR #1430, opened 2020-10-17
Producer.Produce improve error handling

joostas:
In `Producer.ProduceAsync` method inner exception is included when creating `ProduceException` exception
```csharp
catch (Exception ex)
{
    throw new ProduceException<TKey, TValue>(
        new Error(ErrorCode.Local_KeySerialization),
        new DeliveryResult<TKey, TValue>
        {
            Message = message,
            TopicPartitionOffset = new TopicPartitionOffset(topicPartition, Offset.Unset)
        },
        ex); // <-- inner exception is included
}
```
But in synchronous version `Producer.Produce` it is not.

This change makes both methods to handle exception in same way.


joostas:
[clabot:check]

mhowlett (approved):
LGTM - thanks. will be in 1.6.0

PR #1431, opened 2020-10-20
Json serialization

appleseed7even:
Parameterless constructor restriction removed. No need to restrict/forced to create contracts with a parameterless constructor. Since Newtonsoft Json is used, the library is perfectly capable of (de)serializing contracts with private setters and constructor parameters following a naming convention, parameter names names correspond to prop names.

This change will allow serialization of contracts with private setters (immutable) and parameterized constructors.

### DIFF
--- a/src/Confluent.SchemaRegistry.Serdes.Json/JsonSerializer.cs
+++ b/src/Confluent.SchemaRegistry.Serdes.Json/JsonSerializer.cs
@@ -53,7 +53,7 @@ namespace Confluent.SchemaRegistry.Serdes
     ///     integration of System.Text.Json and JSON Schema, so this
     ///     is not yet supported by the serializer.
     /// </remarks>
-    public class JsonSerializer<T> : IAsyncSerializer<T>
+    public class JsonSerializer<T> : IAsyncSerializer<T> where T : class
     {
         private const int DefaultInitialBufferSize = 1024;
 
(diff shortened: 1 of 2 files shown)


appleseed7even:
[clabot:check]

remcobosman (approved):

src/Confluent.SchemaRegistry.Serdes.Json/JsonSerializer.cs
```diff
@@ -53,7 +53,7 @@ namespace Confluent.SchemaRegistry.Serdes
     ///     integration of System.Text.Json and JSON Schema, so this
     ///     is not yet supported by the serializer.
     /// </remarks>
-    public class JsonSerializer<T> : IAsyncSerializer<T>  where T : new()
```
skwasjer, line 56: Should still perhaps be class constraint?

appleseed7even: Fair point, change has been made

skwasjer (approved):

mhowlett:
thanks @appleseed7even , @skwasjer - I don't recall why this constraint is there - there would have been a reason at one point, but I don't any reason for it now.

mhowlett (approved):

PR #1436, opened 2020-10-23
Custom Partitioners

mhowlett:


src/Confluent.Kafka/Impl/LibRdKafka.cs
```diff
@@ -182,6 +182,7 @@ static bool SetDelegates(Type nativeMethodsClass)
             _topic_conf_destroy = (Action<IntPtr>)methods.Single(m => m.Name == "rd_kafka_topic_conf_destroy").CreateDelegate(typeof(Action<IntPtr>));
             _topic_conf_set = (Func<IntPtr, string, string, StringBuilder, UIntPtr, ConfRes>)methods.Single(m => m.Name == "rd_kafka_topic_conf_set").CreateDelegate(typeof(Func<IntPtr, string, string, StringBuilder, UIntPtr, ConfRes>));
             _topic_conf_set_partitioner_cb = (Action<IntPtr, PartitionerDelegate>)methods.Single(m => m.Name == "rd_kafka_topic_conf_set_partitioner_cb").CreateDelegate(typeof(Action<IntPtr, PartitionerDelegate>));
+            _topic_conf_set_opaque = (Action<IntPtr, IntPtr>)methods.Single(m => m.Name == "rd_kafka_topic_conf_set_opaque").CreateDelegate(typeof(Action<IntPtr, IntPtr>));
```
mhowlett, line 185: this is unused, but we might as well leave the binding here, it might come in useful.

src/Confluent.Kafka/Impl/NativeMethods/NativeMethods.cs
```diff
@@ -287,8 +290,7 @@ internal class NativeMethods
 
         [DllImport(DllName, CallingConvention = CallingConvention.Cdecl)]
         internal static extern SafeTopicHandle rd_kafka_topic_new(
-                IntPtr rk,
-                [MarshalAs(UnmanagedType.LPStr)] string topic,
+                IntPtr rk, IntPtr topic,
```
mhowlett, line 293: since Kafka topics are constrained to be a subset of the ANSI characters, marshalling as LPStr is not deficient, but the capability to marshal UTF8 strings I added this PR, although not actually needed, is likely come in useful later, so we might as well leave this change in here.

src/Confluent.Kafka/Producer.cs
```diff
@@ -480,7 +481,7 @@ public Handle Handle
                 {
                     return new Handle { Owner = this, LibrdkafkaHandle = ownedKafkaHandle };
                 }
-                
+
```
mhowlett, line 484: tried to revert this whitespace, my editor refused to cooperate.

mhowlett:
@edenhill ready for review
cc: @ChadJessup - this is how it ended up, FYI - thanks for the initial version.

There is a requirement to be able to set a default custom partitioner (i.e. what to use if a topic partitioner isn't set). This could be implemented by intercepting each produce request, and adding a new partitioner if the topic is unknown, but this is inefficient, on the hot path, and the capability would be good in librdkafka. Leaving this as a separate scope of work.

test/Confluent.Kafka.IntegrationTests/Tests/Producer_CustomPartitioner.cs
```diff
@@ -0,0 +1,79 @@
...
+                Assert.Equal(expectedPartition, (int)dr.Partition);
+                Assert.True(dr.Offset >= 0);
+                Assert.Equal(TimestampType.CreateTime, dr.Message.Timestamp.Type);
+                Assert.True(Math.Abs((DateTime.UtcNow - dr.Message.Timestamp.UtcDateTime).TotalMinutes) < 1.0);
+            };
+
+            using (var topic = new TemporaryTopic(bootstrapServers, PARTITION_COUNT))
+            using (var producer =
+                new ProducerBuilder<string, string>(producerConfig)
+                    .SetPartitioner(topic.Name, (string topicName, int partitionCount, ReadOnlySpan<byte> keyData, bool keyIsNull) => {
```
edenhill, line 59: Is there a default case for when the topic name is not known upfront (or there are too many)?

src/Confluent.Kafka/Producer.cs
```diff
@@ -671,7 +673,31 @@ internal Producer(ProducerBuilder<TKey, TValue> builder)
...
+                Librdkafka.topic_conf_set_partitioner_cb(topicConfigPtr,
+                    (IntPtr rkt, IntPtr keydata, UIntPtr keylen, int partition_cnt, IntPtr rkt_opaque, IntPtr msg_opaque) =>
+                    {
+                        unsafe
+                        {
+                            var keyIsNull = keydata == IntPtr.Zero;
+                            var keyBytes = keyIsNull
+                                ? ReadOnlySpan<byte>.Empty
+                                : new ReadOnlySpan<byte>(keydata.ToPointer(), (int)keylen);
+                            return partitioners[topic](topic, partition_cnt, keyBytes, keyIsNull);
```
edenhill, line 692: Is there any way for the application to get hold of the original Key object in the partitioner?
It shouldn't be available by default (we dont want to hold on to the object due to object lifetime concerns), but some opaque-like construct to pass an arbitrary object to the partitioner (and delivery report delegate) will definitely come in handy.

src/Confluent.Kafka/ProducerBuilder.cs
```diff
@@ -20,6 +20,40 @@
...
+    ///     <paramref name="partitionCount" /> and serialized <paramref name="keyData" />.
+    /// </summary>
+    /// <remarks>
+    ///     A partioner instance may be called in any thread at any time and
+    ///     may be called multiple times for the same message/key.
+    ///
+    ///     A partitioner:
+    ///     - MUST NOT call any method on the producer instance.
+    ///     - MUST NOT block or execute for prolonged periods of time.
+    ///     - MUST return a value between 0 and partition_cnt-1, or the
```
edenhill, line 34: partitionCount

src/Confluent.Kafka/ProducerBuilder.cs
```diff
@@ -20,6 +20,40 @@
...
+    ///       special <seealso cref="Confluent.Kafka.Partition.Any"/> value.
```
edenhill, line 35: What does that do?

test/Confluent.Kafka.IntegrationTests/Tests/Producer_CustomPartitioner.cs
```diff
@@ -0,0 +1,79 @@
...
+                        var keyString = System.Text.UTF8Encoding.UTF8.GetString(keyData.ToArray());
+                        return int.Parse(keyString.Split(" ").Last()) % partitionCount;
+                    })
+                .Build())
+            {
+                for (int i=0; i<PARTITION_COUNT; ++i)
```
edenhill, line 65: Not test for the .Any case

edenhill (approved):
LGTM, with commentary.

test/Confluent.Kafka.IntegrationTests/Tests/Producer_CustomPartitioner.cs
mhowlett: yes, there is a requirement for this, and a customer has asked for it. Librdkafka does not support specifying a default custom partitioner right now afaict, right?

from the pr commit message:

> There is a requirement to be able to set a default custom partitioner (i.e. what to use if a topic partitioner isn't set). This could be implemented by intercepting each produce request, and adding a new partitioner if the topic is unknown, but this is inefficient, on the hot path, and the capability would be good in librdkafka. Leaving this as a separate scope of work.

mhowlett: ok, thanks for bringing my attention to `conf_set_default_topic_conf`, i'll add this in this PR.

src/Confluent.Kafka/Producer.cs
mhowlett: delegates can have associated state, either because they are a closure which captures the state from their context, or because they point to a method on a class instance (which references state in the instance). same with the delivery report handler. there's no need for this to come through librdkafka (which would be more difficult due to lifetimes).

src/Confluent.Kafka/ProducerBuilder.cs
mhowlett: it returns -1, which I believe should send to a random partition (perhaps sticky). will add a note.

test/Confluent.Kafka.IntegrationTests/Tests/Producer_CustomPartitioner.cs
mhowlett: given this is tested in librdkafka, a test here has marginal benefit. however i suspect it's not tested in librdkafka, so i'll extend the test case a bit.

src/Confluent.Kafka/ProducerBuilder.cs
mhowlett: huh, my assumption was wrong, removing this as an option. value can only be 0 .. partitionCnt - 1.

mhowlett: we should perhaps support this in librdkafka.

src/Confluent.Kafka/Producer.cs
edenhill: But since the partitioner is set on producer instantiation it is not possible to bind any per-message context to the partitioner.

mhowlett: What's the use case for that? the user can of course explicitly produce to a partition if more context is required from produce time (which is the work-around currently suggested in lieu of custom partitioners), with the downside being they can't produce before knowing the partition count.

The upgrade path to supporting this if not added now is to add a `PartitionerDelegateWithOpaqe` and overload to the builder + a property to the Message class. Which is all acceptable. I don't think we'll ever need to do it though.

src/Confluent.Kafka/ProducerBuilder.cs
```diff
@@ -130,6 +174,38 @@ public ProducerBuilder(IEnumerable<KeyValuePair<string, string>> config)
...
+            }
+            this.Partitioners.Add(topic, partitioner);
+            return this;
+        }
+
+        /// <summary>
+        ///     Set a custom partitioner that will be used for all topics
+        ///     except those for which a partitioner has been explicitly configured.
+        /// </summary>
+        public ProducerBuilder<TKey, TValue> SetDefaultPartitioner(PartitionerDelegate partitioner)
```
mhowlett, line 199: Alternatively, I could have overloaded the name `SetPartitioner`, but i think it adds value to have the word `Default` in there.

src/Confluent.Kafka/Producer.cs
```diff
@@ -634,16 +636,46 @@ internal Producer(ProducerBuilder<TKey, TValue> builder)
...
+            };
+
+            // Configure the default custom partitioner, if specified, by creating a new
+            // default topic config object. It's important that this occurs before the
+            // other properties are set, since if any of those were fallthru properties,
+            // their values would get clobbered by the default ones in the topic config
+            // object we create here.
+            if (defaultPartitioner != null)
+            {
+                var defaultTopicConfigHandle = createTopicConfigWithPartitioner(defaultPartitioner);
```
edenhill, line 668: Any time you specify a topic_conf object to topic_new() you will loose whatever default topic config is accumulated on the global config, which is not what we want, so I think what we want to do is get the default_topic_config of the rk and then apply the partitioner to that and pass that object to topic_new(), both in the topic-specific and default case.
See `rd_kafka_default_topic_conf_dup()`

edenhill (changes requested):
Add test cases that mixes string-based partitioner, DefaultPartitioner, and topic-specific Partitioner to verify that the correct one is used.

src/Confluent.Kafka/Producer.cs
mhowlett: thanks for picking up on this ...

src/Confluent.Kafka/Producer.cs
```diff
@@ -671,7 +673,41 @@ internal Producer(ProducerBuilder<TKey, TValue> builder)
             }
 
             this.ownedKafkaHandle = SafeKafkaHandle.Create(RdKafkaType.Producer, configPtr, this);
-            configHandle.SetHandleAsInvalid(); // config object is no longer usable.
+            configHandle.SetHandleAsInvalid();  // ownership was transferred.
+                                                // configPtr remains valid over the lifetime of this method.
+
+            Func<PartitionerDelegate, IntPtr> createTopicConfigWithPartitioner = (partitioner) => {
+                var topicConfigPtr = SafeTopicConfigHandle.DuplicateDefaultTopicConfig(configPtr);
```
edenhill, line 680: Shouldn't this be `DuplicateDefaultTopicConfig(this.ownedKafkaHandle)`?

src/Confluent.Kafka/Producer.cs
```diff
@@ -671,7 +673,41 @@ internal Producer(ProducerBuilder<TKey, TValue> builder)
...
+                Librdkafka.topic_conf_set_partitioner_cb(configPtr,
+                    (IntPtr rkt, IntPtr keydata, UIntPtr keylen, int partition_cnt, IntPtr rkt_opaque, IntPtr msg_opaque) =>
+                    {
+                        unsafe
+                        {
+                            var topicNamePtr = Librdkafka.topic_name(rkt);
```
edenhill, line 686: What's the performance implication of doing a alloc-copy-utf8transform for each partitioned message?
E.g., what's the thruput drop for 100 byte topics going at full speed?

src/Confluent.Kafka/ProducerBuilder.cs
```diff
@@ -20,6 +20,39 @@
 
 namespace Confluent.Kafka
 {
+    /// <summary>
+    ///     Calculate a partition number given a <paramref name="topic" />,
```
edenhill, line 24: The topic name is typically not relevant in keying.

src/Confluent.Kafka/ProducerBuilder.cs
```diff
@@ -20,6 +20,39 @@
...
+    ///     The topic.
+    /// </param>
+    /// <param name="partitionCount">
+    ///     The number of partitions in <paramref name="topic" />.
+    /// </param>
+    /// <param name="keyData">
+    ///     The serialized key data.
+    /// </param>
+    /// <param name="keyIsNull">
+    ///     Whether or not the key is null (distinguishes the null and empty case).
```
edenhill, line 47: It's too bad we can't utilize sticky partitioning when a custom partitioner is set.
We could add a new API, SetNonRandomPartitioner (and the analogue on the rdkafka side) to have a custom partitioner that is only passed non-NULL keys.

test/Confluent.Kafka.IntegrationTests/Tests/Producer_MultiPartitioner.cs
```diff
@@ -0,0 +1,91 @@
...
+                    {
+                        Assert.Equal("hello", dr.Message.Key);
+                    }
+                    else
+                    {
+                        Assert.Equal("world", dr.Message.Key);
+                    }
+                };
+
+                producer.Produce(topic1.Name, new Message<string, Null> { Key = "hello" }, dh);
```
edenhill, line 77: There is no test-case for Null keys.

edenhill (changes requested):

src/Confluent.Kafka/Producer.cs
mhowlett: how could this have possibly been working is the question on my mind.

src/Confluent.Kafka/Producer.cs
mhowlett: I don't know. we don't have a choice in the default all-topic partitioner case. if it's slow someone will complain and we can put energy into it then maybe. I don't reckon it is slow. the alloc will be cheap because it will be short-lived.

src/Confluent.Kafka/ProducerBuilder.cs
mhowlett: it's there and in the docs because it's part of the partitioner api. it could be relevant if someone is setting a default partitioner... but i can't think of a specific use case. i will add a note that it is not typically relevant.

mhowlett: maybe this is where you were coming from with the perf. question.

if it's slow (don't reckon it'll be too material), we can add another type in the future `SimplePartitionerDelegate` or something, but i think for now the more general option is the one to go with.

src/Confluent.Kafka/ProducerBuilder.cs
mhowlett: i'd do that with another argument on `SetPartitioner` / `SetDefaultPartitioner` (a boolean). it can be added later as a non breaking change by giving it a default value.

src/Confluent.Kafka/Producer.cs
edenhill: Maybe you can compare throughput with and without this alloc, use the mock broker.

mhowlett: prioritization! i would rather work on other things. this is getting stuck in the details when we have bigger stuff to do. waste of time

test/Confluent.Kafka.IntegrationTests/Tests/Producer_MultiPartitioner.cs
mhowlett: added to Producer_CustomPartitioner

src/Confluent.Kafka/Producer.cs
edenhill: Knowing the performance of a feature is part of the deliverable. Even though this particular case here with a string object instantiation might not be that interesting, you/we should know the throughput numbers with and without custom partitioners.

src/Confluent.Kafka/Producer.cs
```diff
@@ -670,8 +672,50 @@ internal Producer(ProducerBuilder<TKey, TValue> builder)
...
+                    });
+            };
+
+            // Configure the default custom partitioner.
+            if (defaultPartitioner != null)
+            {
+                // The default topic config may have been modified by topic-level
+                // configuraton parameters passed down from the top level config.
+                // If that's the case, duplicate the default topic config to avoid
+                // colobbering any already configured values.
```
edenhill, line 699: clobbering

src/Confluent.Kafka/Producer.cs
```diff
@@ -670,8 +672,50 @@ internal Producer(ProducerBuilder<TKey, TValue> builder)
...
+                            var keyIsNull = keydata == IntPtr.Zero;
+                            var keyBytes = keyIsNull
+                                ? ReadOnlySpan<byte>.Empty
+                                : new ReadOnlySpan<byte>(keydata.ToPointer(), (int)keylen);
+                            return partitioner(topic, partition_cnt, keyBytes, keyIsNull);
+                        }
+                    });
+            };
+
+            // Configure the default custom partitioner.
```
edenhill, line 693: Is this part always performed after all other (global) config has been applied?

test/Confluent.Kafka.IntegrationTests/Tests/Producer_CustomPartitioner.cs
```diff
@@ -0,0 +1,140 @@
...
+                                return int.Parse(keyString.Split(" ").Last()) % partitionCount;
+                            });
+                            break;
+                        case 2:
+                            // Default custom partitioner in case where default topic config is present due to topic level config in top-level config.
+                            var producerConfig2 = new ProducerConfig
+                            {
+                                BootstrapServers = bootstrapServers
+                            };
+                            producerConfig2.Set("queuing.strategy", "lifo");
```
edenhill, line 90: This is deprecated and will be removed eventually.

test/Confluent.Kafka.IntegrationTests/Tests/Producer_CustomPartitioner.cs
```diff
@@ -0,0 +1,140 @@
...
+
+
+            // Null key
+
+            using (var topic = new TemporaryTopic(bootstrapServers, PARTITION_COUNT))
+            using (var producer = new ProducerBuilder<Null, string>(producerConfig)
+                .SetDefaultPartitioner((string topicName, int partitionCount, ReadOnlySpan<byte> keyData, bool keyIsNull) =>
+                {
+                    Assert.True(keyIsNull);
+                    Assert.True(!keyIsNull);
```
edenhill, line 126: How can both of these be true?

edenhill (approved):

test/Confluent.Kafka.IntegrationTests/Tests/Producer_CustomPartitioner.cs
mhowlett: ok. the reason this got here is the test wasn't working and i was trying something obscure. changed.

src/Confluent.Kafka/Producer.cs
mhowlett: yes - it's immediately before the handle is created.

src/Confluent.Kafka/Producer.cs
mhowlett: nice word, huh?

test/Confluent.Kafka.IntegrationTests/Tests/Producer_CustomPartitioner.cs
mhowlett: no idea how the second assert got there. i was testing the test at one point is all i can think of.

PR #1443, opened 2020-10-30
Mono support

mhowlett:
These changes are based closely on the work by @4garbage in #788 
I'm structuring the PR as a series of incremental commits in order to make the changes easy to follow and check.

### DIFF
--- a/src/Confluent.Kafka/Impl/LibRdKafka.cs
+++ b/src/Confluent.Kafka/Impl/LibRdKafka.cs
@@ -365,7 +365,7 @@ namespace Confluent.Kafka.Impl
 #else
                 const int RTLD_NOW = 2;
 
-                var nativeMethodTypes = new List<Type>();
+                List<Type> nativeMethodTypes;
 
                 if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                 {
@@ -378,8 +378,7 @@ namespace Confluent.Kafka.Impl
                             throw new InvalidOperationException($"Failed to load librdkafka at location '{userSpecifiedPath}'. Win32 error: {Marshal.GetLastWin32Error()}");
                         }
                     }
-
-                    nativeMethodTypes.Add(typeof(NativeMethods.NativeMethods));
+                    nativeMethodTypes = new List<Type> { typeof(NativeMethods.NativeMethods) };
                 }
                 else if (RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
                 {
@@ -391,7 +390,7 @@ namespace Confluent.Kafka.Impl
                         }
                     }
 
-                    nativeMethodTypes.Add(typeof(NativeMethods.NativeMethods));
+                    nativeMethodTypes = new List<Type> { typeof(NativeMethods.NativeMethods) };
                 }
                 else if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
                 {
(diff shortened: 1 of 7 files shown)


src/Confluent.Kafka/Impl/NativeMethods/NativeMethods_Alpine.cs
```diff
@@ -40,7 +40,11 @@ namespace Confluent.Kafka.Impl.NativeMethods
     /// </remarks>
     internal class NativeMethods_Alpine
     {
+#if NET45 || NET46 || NET47
+         public const string DllName = "alpine-librdkafka.so";
```
mhowlett, line 44: prior to this PR, we would never try to load librdkafka in the NET45/46/47 case on linux (which is what mono allows you to do), so this change does not modify any scenario that was currently operational.

mhowlett:
This is now ready for review. I have manually verified that I can run a simple test program that uses Confluent.Kafka.Producer on linux with mono in the official mono 6.8.0.123 docker image (and that I can't without this change). Setting up testing for this would be much more work than actually making the change, and not of commensurate value, so I suggest not doing that, at least until we have a more general automated test framework set up for many environments.

This change is conceptually very simple - before, in the .NET Framework case, we never tried to load librdkafka unless on windows. Now we do (which is the problem mono is solving for). In order to realize this, a bit of a refactor was necessary resulting in a difficult to follow diff. To make that easier to follow i did the change as a series of focussed commits.

The main thing to check for is that I didn't screw up existing scenarios (I don't think so).

I followed what @4garbage did closely, and I thought about it from scratch, so a review of this is effectively a 3rd check.

src/Confluent.Kafka/Impl/LibRdKafka.cs
```diff
@@ -495,6 +442,67 @@ public static bool Initialize(string userSpecifiedPath)
...
+        {
+            string path = userSpecifiedPath;
+            if (path == null)
+            {
+                // in net45, librdkafka.dll is not in the process directory, we have to load it manually
+                // and also search in the same folder for its dependencies (LOAD_WITH_ALTERED_SEARCH_PATH)
+                var is64 = IntPtr.Size == 8;
+                var baseUri = new Uri(Assembly.GetExecutingAssembly().GetName().EscapedCodeBase);
+                var baseDirectory = Path.GetDirectoryName(baseUri.LocalPath);
+                var dllDirectory = Path.Combine(
```
edenhill, line 430: Maybe move these four path blocks to a for loop and try to load each one until one succeeds?

src/Confluent.Kafka/Impl/LibRdKafka.cs
```diff
@@ -502,8 +457,71 @@ private static bool LoadNetFrameworkDelegates(string userSpecifiedPath)
...
+                if (PosixNative.dlopen(userSpecifiedPath, RTLD_NOW) == IntPtr.Zero)
+                {
+                    throw new InvalidOperationException($"Failed to load librdkafka at location '{userSpecifiedPath}'. dlerror: '{PosixNative.LastError}'.");
+                }
+            }
+
+            return TrySetDelegates(new List<Type> { typeof(NativeMethods.NativeMethods) });
+        }
+
+        private static bool LoadLinuxDelegates(string userSpecifiedPath)
```
edenhill: Since LoadOSX and LinuxDelegates are identical (from what I can see), and it is really a POSIX API, maybe call it LoadPosixDelegates()?

edenhill (approved):
LGTM, but needs a CI worker that runs the test suite with Mono before merging.

src/Confluent.Kafka/Impl/LibRdKafka.cs
mhowlett: the LinuxDelegates method tries up to 4 different `NativeMethods` classes to enable the alpine, centos and debian .so's, so it's pretty tied to linux. the macos one just tries the general one.

src/Confluent.Kafka/Impl/LibRdKafka.cs
mhowlett: that's a good idea, but minor detail + this is copy/paste from existing implementation with lots of runtime, and any change opens up possibility of bugs + it takes effort.

mhowlett:
merging since we won't claim support for mono. if we do start claiming this, I can add CI tests then.

PR #1446, opened 2020-11-02
Refactor RequestTimeConsumer.

dogac00:
We can use `BackgroundService` which implements `IHostedService` in order to avoid using our own cancellation token. The token passed to the `ExecuteAsync` method is fired when application is shutting down. We should catch `OperationCanceledException` which may get thrown in Consume if the token is fired. And we loop unless the operation is canceled.


mhowlett (approved):
Thanks @dogac00 ! - this is much neater.

dogac00:
[clabot:check]

PR #1452, opened 2020-11-08
Changelog updates for v1.6.0-PRE1 release

mhowlett:


CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
 # 1.6.0-PRE1
 
-- References librdkafka 1.6.0. Refer to the [release notes](https://github.com/edenhill/librdkafka/releases/tag/v1.6.0) for more information. Headline features:
+- References librdkafka 1.6.0-PRE1. Refer to the [release notes](https://github.com/edenhill/librdkafka/releases/tag/v1.6.0) for more information. Headline features:
   - KIP-429: Incremental rebalancing.
   - KIP-480: Sticky partitioner.
+- KIP-22: Support for custom partitioners.
+- Librdkafka is no longer prevented from being loaded on Linux or MacOS when targeting `net452` (allows Confluent.Kafka to be used with Mono on these platforms).
```
edenhill, line 7: Double negative, maybe focus more on what it is now able to do, than what it used to be unable to, nor not be able to, or unable to not able to be to unable to not be.
> Confluent.Kafka can now be used with Mono on Linux and MacOS when targeting net452. **NOTE**: Mono is not a supported runtime.

CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
...
+- Exceptions thrown by `Producer.Produce` now include an inner exception with additional context on the error ([joostas](https://github.com/joostas)).
```
edenhill, line 8: A small code blob showing how to get at the inner exception would be fantastic!

edenhill (approved):
Needs a bit of twiddling

CHANGELOG.md
mhowlett: it's standard practice and the norm everywhere else except this method, so i think it would be out of place. it's automatically covered by our examples since they print exceptions with ToString, which will print the name of the InnerException by default.

PR #1500, opened 2021-01-12
fix schema reference order on schema registration

pzaloznyi:
There is an issue in schema reference registration when we have few dependencies in schema. Currently all references are registering asynchronously and are adding to result list without strict sequence. The fix just orders the resulting sequence so it will be the same on every schema registration.


pzaloznyi:
[clabot:check]

PR #1521, opened 2021-02-05
add IsInvalid property to Handle class

volgunin:
Reason for change:
The ConsumerBuilder starts logging messages to LogHandler from within Build method with partially constructed IConsumer object  (Kafka library handle is null). There is no way to determine whether the IConsumer properties can be accessed without triggering an exception.

This property will allow to check whether Handle property in IConsumer interface is valid and therefore other properties and methods could be called.

### DIFF
--- a/src/Confluent.Kafka/Handle.cs
+++ b/src/Confluent.Kafka/Handle.cs
@@ -25,6 +25,14 @@ namespace Confluent.Kafka
     /// </summary>
     public class Handle
     {
+        /// <summary>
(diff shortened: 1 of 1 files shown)


volgunin:
[clabot:check]

volgunin:
> [clabot:check]

mhowlett:
LGTM - can you:
1. indent the text in the doc comments to match the style of other methods / properties in the project
2. open this against 1.6.x

then i'll merge - thanks!

volgunin:
A new PR for branch 1.6 was created.

Issue #1523, opened 2021-02-07
ProtobufSerializer has a race condition in RegisterOrGetReferences method


yurii-hunter:
Description
===========
The bug is in RegisterOrGetReferences method from the ProtobufSerializer class.
Several asynchronous tasks update `result` variable simultaneously. Where `result` is variable of type `List<SchemaReference>` which is not thread safe. 
I would suggest to use pure function to avoid side effects and race condition.

`Confluent.SchemaRegistry.Serdes.Protobuf - 1.5.3`

How to reproduce
================
If I simplify the function [RegisterOrGetReferences](https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/src/Confluent.SchemaRegistry.Serdes.Protobuf/ProtobufSerializer.cs#L169) just for showing the problem it may look like this:
```
[Fact]
public async Task Test()
{
    var results = await GetAsync().ConfigureAwait(false);
    Assert.Equal(3, results.Count);
}

public static async Task<List<int>> GetAsync()
{
    List<int> result = new List<int>();

    var tasks = new Task[3];
    for (int i = 0; i < 3; ++i)
    {
        int dependency = i;
        Func<Task> t = async () =>
        {
            await Task.Delay(100).ConfigureAwait(false);
            result.Add(dependency);
        };
        tasks[i] = t();
    }

    await Task.WhenAll(tasks.ToArray()).ConfigureAwait(false);
    return result;
}


```
Checklist
=========

Please provide the following information:
 - [x] A complete (i.e. we can run it), minimal program demonstrating the problem. No need to supply a project file.
 - [x] Confluent.Kafka nuget version.
 - [ ] Apache Kafka version.
 - [ ] Client configuration.
 - [ ] Operating system.
 - [ ] Provide logs (with "debug" : "..." as necessary in configuration).
 - [ ] Provide broker log excerpts.
 - [x] Critical issue.

yurii-hunter:
Hi @mhowlett !
Could you please look at this issue and PR whenever you able?
That is really critical issue for clients that use Protobuf

Thank you in advance!

mhowlett:
good point, well spotted.

yurii-hunter:
Fixed in [v1.6.2](https://github.com/confluentinc/confluent-kafka-dotnet/releases/tag/v1.6.2)

PR #1524, opened 2021-02-07
Fix race condition in RegisterOrGetReferences method

yurii-hunter:
This PR fixes issue #1523
List of result were updating from different tasks.


mhowlett:
thanks for pointing this out @yurii-hunter - LGTM. can you please rebase onto 1.6.x? we will release 1.6.0 soon. ideally we'd backport all bugs like this, but usually don't due to time constraints.

yurii-hunter:
@mhowlett I've rebased onto 1.6.x and also changed the target branch to 1.6.x

mhowlett (approved):

mhowlett:
thanks!

Issue #1529, opened 2021-02-10
ConsumerGroupMetadata is internal and there is no (clear) way for an external library to wrap it


agayev169:
`IConsumerGroupMetadata` is an empty interface, that does not have a public property/method to access the metadata itself, while `ConsumerGroupMetadata` has an internal field containing the serialized version of metadata. 

https://github.com/confluentinc/confluent-kafka-dotnet/blob/2b54dd2ab07f6eaa886d88c57143838627b2512f/src/Confluent.Kafka/ConsumerGroupMetadata.cs#L23

Due to this design, the only way to get metadata of a consumer group is to cast `IConsumerGroupMetadata` to `ConsumerGroupMetadata`.

https://github.com/confluentinc/confluent-kafka-dotnet/blob/2b54dd2ab07f6eaa886d88c57143838627b2512f/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs#L553

A better way to do it would be to have a public property to get metadata as `byte[]` in `IConsumerGroupMetadata` itself.
Please, consider this change.

P.S. I will try to do these changes in my free time, but maybe there is a samaritan that could do this now.

mhowlett:
serializedMetadata is an internal format that isn't intended to be publicly visible.

you are right that it would be good for ConsumerGroupMetadata to expose something useful, but that will be a bigger task. probably a flaw in how we did this tbh (though nothing that can't be worked around).

edenhill:
@agayev169 The ConsumerGroupMetadata's purpose is to pass consumer group state from a consumer to a transactional producer, the state information is of no use to the application and thus there's no reason to expose it.

mhowlett:
the java client actually exposes information on the instance returned from `groupMetadata`, which is what is passed to the transactional producer. whether on not this is really useful (seems like @agayev169 want's it), our api does expose less than the Java one does.

agayev169:
> @agayev169 The ConsumerGroupMetadata's purpose is to pass consumer group state from a consumer to a transactional producer, the state information is of no use to the application and thus there's no reason to expose it.

[...]

PR #1530, opened 2021-02-11
Make public getter for consumer group metadata

agayev169:
Read #1529 for the purpose of the change


agayev169:
> [clabot:check]

[clabot:check]

mhowlett:
refer to my response in #1529 for possible work-around.
if we get more demand for this, i'll consider adding the feature.
also, i'd prefer not explicitly exposing the byte[] array like that as it's exposing an internal detail. i'd probably have a serialize method.

PR #1548, opened 2021-03-03
ReadOnlySpan for internal Produce methods

alec-anikin:
The first step to low-level producer without serializers. 
Proposed producer interface can looks like
```        
Task<DeliveryResult> ProduceAsync(
            string topic,
            ReadOnlySpan<byte> key,
            ReadOnlySpan<byte> value,
            Headers headers = null,
            Timestamp timestamp = default,
            CancellationToken cancellationToken = default(CancellationToken));
``` 
If you are interested in such changes, I can prepare a pull request with an implementation. 

PS: Thank you for great library :)

### DIFF
--- a/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
+++ b/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
@@ -286,13 +286,13 @@ namespace Confluent.Kafka.Impl
             return topicHandle;
         }
 
-        private IntPtr marshalHeaders(IEnumerable<IHeader> headers)
+        private IntPtr marshalHeaders(IReadOnlyCollection<IHeader> headers)
         {
             var headersPtr = IntPtr.Zero;
 
-            if (headers != null && headers.Any())
+            if (headers != null && headers.Count > 0)
             {
-                headersPtr = Librdkafka.headers_new((IntPtr) headers.Count());
+                headersPtr = Librdkafka.headers_new((IntPtr) headers.Count);
                 if (headersPtr == IntPtr.Zero)
                 {
                     throw new Exception("Failed to create headers list.");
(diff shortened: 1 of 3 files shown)


src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -344,46 +348,40 @@ private IntPtr marshalHeaders(IEnumerable<IHeader> headers)
             var gchValue = default(GCHandle);
             var gchKey = default(GCHandle);
 
-            if (val == null)
+            if (val.Length == 0 && val != ReadOnlySpan<byte>.Empty)
```
mhowlett, line 351: good idea

src/Confluent.Kafka/Confluent.Kafka.csproj
```diff
@@ -18,6 +18,7 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <SignAssembly>true</SignAssembly>
     <AssemblyOriginatorKeyFile>Confluent.Kafka.snk</AssemblyOriginatorKeyFile>
+    <LangVersion>7.3</LangVersion>
```
mhowlett, line 21: are we breaking anything with this? i don't think so, but maybe

src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
```diff
@@ -284,13 +286,13 @@ internal SafeTopicHandle NewTopic(string topic, IntPtr config)
             return topicHandle;
         }
 
-        private IntPtr marshalHeaders(IEnumerable<IHeader> headers)
+        private IntPtr marshalHeaders(IReadOnlyCollection<IHeader> headers)
         {
             var headersPtr = IntPtr.Zero;
 
-            if (headers != null && headers.Any())
+            if (headers != null && headers.Count > 0)
```
mhowlett, line 293: `Any` may still be better here

alec-anikin: but not from performance perspective. Any allocates HeadersEnumerator + MoveNext, when Count is just access to the List field.

mhowlett:
this is looking like we want to merge it to me, but i'll spend more time thinking about it before doing so.

re: low level producing without serializers - i'm hesitant (and explicitly didn't) add this to the `K, V` typed producer because I don't think it belongs there. I'd like to ultimately add an untyped Producer, where the serializer is chosen automatically based on type - i think this may fit in well there. Alternatively, maybe we have a whole new `LowLevelProducer` or `BytesProducer` class or something. The latter approach isn't seeming too bad to me ATM. Of course, both these would require a fair bit of refactoring under the covers so we get good code reuse.

alec-anikin:
Yes, methods with Spans are not at all suitable for the current KV producer interface. I've already tried making some draft. I'll try to polish it and create another pull request.

PR #1587, opened 2021-04-19
Use returned schema id as key

jeremy001181:


mhowlett (approved):
thanks! good catch.

jeremy001181:
Hey @mhowlett wonder if it is possible to get this fix into RC4 so I can get it from there before the official 1.7.0 live?

mhowlett:
whoops, that's a mistake. it's in the change log on the 1.7.0 branch! will do.

PR #1590, opened 2021-04-21
Web Example Authentication Updates

cjgalione:
Updated SecurityProtocol to use C#-specific 'SaslSsl' value instead of librdkafka-specific 'sasl_ssl' value, as the example results in 'System.FormatException: sasl_ssl is not a valid value for SecurityProtocol.' when using the librdkafka value. Additionally, changed SaslMechanism from plain to Plain to align with enum value


cjgalione:
[clabot:check]

mhowlett:
thanks @cjgalione !

PR #1628, opened 2021-06-01
Enhance serializers with use.latest.version config

rayokota:
Also 
- add skip.known.types config for Protobuf serializer
- fix serialization of message indexes per https://docs.confluent.io/platform/current/schema-registry/serdes-develop/index.html#wire-format

### DIFF
--- a/src/Confluent.SchemaRegistry.Serdes.Avro/GenericSerializerImpl.cs
+++ b/src/Confluent.SchemaRegistry.Serdes.Avro/GenericSerializerImpl.cs
@@ -34,6 +34,7 @@ namespace Confluent.SchemaRegistry.Serdes
     {
         private ISchemaRegistryClient schemaRegistryClient;
         private bool autoRegisterSchema;
+        private bool useLatestVersion;
         private int initialBufferSize;
         private SubjectNameStrategyDelegate subjectNameStrategy;
 
@@ -46,11 +47,13 @@ namespace Confluent.SchemaRegistry.Serdes
         public GenericSerializerImpl(
             ISchemaRegistryClient schemaRegistryClient,
             bool autoRegisterSchema,
+            bool useLatestVersion,
             int initialBufferSize,
             SubjectNameStrategyDelegate subjectNameStrategy)
         {
             this.schemaRegistryClient = schemaRegistryClient;
             this.autoRegisterSchema = autoRegisterSchema;
+            this.useLatestVersion = useLatestVersion;
             this.initialBufferSize = initialBufferSize;
             this.subjectNameStrategy = subjectNameStrategy;
         }
@@ -131,14 +134,26 @@ namespace Confluent.SchemaRegistry.Serdes
                     if (!registeredSchemas.Contains(subjectSchemaPair))
                     {
                         int newSchemaId;
-                        // first usage: register/get schema to check compatibility
-                        if (autoRegisterSchema)
+                        if (useLatestVersion)
                         {
-                            newSchemaId = await schemaRegistryClient.RegisterSchemaAsync(subject, writerSchemaString).ConfigureAwait(continueOnCapturedContext: false);
+                            var latestSchema = await schemaRegistryClient.GetLatestSchemaAsync(subject)
+                                .ConfigureAwait(continueOnCapturedContext: false);
+                            newSchemaId = latestSchema.Id;
                         }
                         else
                         {
-                            newSchemaId = await schemaRegistryClient.GetSchemaIdAsync(subject, writerSchemaString).ConfigureAwait(continueOnCapturedContext: false);
+                            // first usage: register/get schema to check compatibility
+                            if (autoRegisterSchema)
+                            {
+                                newSchemaId = await schemaRegistryClient
+                                    .RegisterSchemaAsync(subject, writerSchemaString)
+                                    .ConfigureAwait(continueOnCapturedContext: false);
+                            }
+                            else
+                            {
+                                newSchemaId = await schemaRegistryClient.GetSchemaIdAsync(subject, writerSchemaString)
+                                    .ConfigureAwait(continueOnCapturedContext: false);
+                            }
                         }
 
                         if (!schemaIds.ContainsKey(writerSchemaString))
--- a/src/Confluent.SchemaRegistry.Serdes.Avro/SpecificSerializerImpl.cs
+++ b/src/Confluent.SchemaRegistry.Serdes.Avro/SpecificSerializerImpl.cs
@@ -82,6 +82,7 @@ namespace Confluent.SchemaRegistry.Serdes
 
         private ISchemaRegistryClient schemaRegistryClient;
         private bool autoRegisterSchema;
+        private bool useLatestVersion;
         private int initialBufferSize;
         private SubjectNameStrategyDelegate subjectNameStrategy;
 
@@ -97,11 +98,13 @@ namespace Confluent.SchemaRegistry.Serdes
         public SpecificSerializerImpl(
             ISchemaRegistryClient schemaRegistryClient,
             bool autoRegisterSchema,
+            bool useLatestVersion,
             int initialBufferSize,
             SubjectNameStrategyDelegate subjectNameStrategy)
         {
             this.schemaRegistryClient = schemaRegistryClient;
             this.autoRegisterSchema = autoRegisterSchema;
+            this.useLatestVersion = useLatestVersion;
             this.initialBufferSize = initialBufferSize;
             this.subjectNameStrategy = subjectNameStrategy;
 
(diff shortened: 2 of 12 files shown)


src/Confluent.SchemaRegistry.Serdes.Avro/SpecificSerializerImpl.cs
```diff
@@ -208,13 +211,23 @@ public async Task<byte[]> Serialize(string topic, T data, bool isKey)
...
-                        currentSchemaData.WriterSchemaId = autoRegisterSchema
-                            ? await schemaRegistryClient
-                                .RegisterSchemaAsync(subject, currentSchemaData.WriterSchemaString)
-                                .ConfigureAwait(continueOnCapturedContext: false)
-                            : await schemaRegistryClient.GetSchemaIdAsync(subject, currentSchemaData.WriterSchemaString)
+                        if (useLatestVersion)
+                        {
+                            var latestSchema = await schemaRegistryClient.GetLatestSchemaAsync(subject)
                                 .ConfigureAwait(continueOnCapturedContext: false);
+                            currentSchemaData.WriterSchemaId = latestSchema.Id;
```
mhowlett, line 218: if `use.latest.schema` is `true`, then the schema must exist (will not be auto-registered). however this is at odds with the default for `auto.register.schema`, which is `true`. same in the generic case.

test/Confluent.SchemaRegistry.Serdes.IntegrationTests/Tests_Avro/AutoRegisterSchemaDisabled.cs
```diff
@@ -135,6 +135,17 @@ public static void AutoRegisterSchemaDisabled(string bootstrapServers, string sc
                 {
                     producer.ProduceAsync(topic.Name, new Message<string, int> { Key = "test", Value = 112 }).Wait();
                 }
+
+                // config with avro.serializer.use.latest.version == true should also work now.
+                using (var schemaRegistry = new CachedSchemaRegistryClient(new SchemaRegistryConfig { Url = schemaRegistryServers }))
```
mhowlett, line 140: an important test i think (which you don't have) would be if a newer schema exists in SR, and the client is specifying an older schema, right? ideally, we'd have this test for avro, json and protobuf... i think implementing it for at least one of the cases has value (i recognize it's a bit of a pain).

finally, we have a protobuf problem. the protobuf tests have actually been broken on newer SR versions for some time, and i haven't prioritized looking into it. Since I think your changes actually overlap (the tests may need updating due to your changes), i'd love if you took a look ... (if not, i will get to it).

to run the tests you need .net core 2.1 installed. then `dotnet test -f netcoreapp2.1`. here are the test failures (as occur on your branch, but as i said, some of this is pre-existing).

```
[xUnit.net 00:02:58.3318311]     Confluent.SchemaRegistry.Serdes.IntegrationTests.Tests.ProduceConsumeNestedProtobuf(bootstrapServers: "localhost:9092", schemaRegistryServers: "localhost:8081") [FAIL]
  Failed Confluent.SchemaRegistry.Serdes.IntegrationTests.Tests.ProduceConsumeNestedProtobuf(bootstrapServers: "localhost:9092", schemaRegistryServers: "localhost:8081") [46 s]
  Error Message:
   Assert.Equal() Failure
Expected: 3
Actual:   6
  Stack Trace:
     at Confluent.SchemaRegistry.Serdes.IntegrationTests.Tests.ProduceConsumeNestedProtobuf(String bootstrapServers, String schemaRegistryServers) in /Users/matt/git/confluent-kafka-dotnet/test/Confluent.SchemaRegistry.Serdes.IntegrationTests/Tests_Protobuf/ProduceConsumeNested.cs:line 78
[xUnit.net 00:03:52.0111808]     Confluent.SchemaRegistry.Serdes.IntegrationTests.Tests.ProduceConsumeSchemaManyMessagesProtobuf(bootstrapServers: "localhost:9092", schemaRegistryServers: "localhost:8081") [FAIL]
  Failed Confluent.SchemaRegistry.Serdes.IntegrationTests.Tests.ProduceConsumeSchemaManyMessagesProtobuf(bootstrapServers: "localhost:9092", schemaRegistryServers: "localhost:8081") [46 s]
  Error Message:
   Assert.Equal() Failure
Expected: 1
Actual:   2
  Stack Trace:
     at Confluent.SchemaRegistry.Serdes.IntegrationTests.Tests.ProduceConsumeSchemaManyMessagesProtobuf(String bootstrapServers, String schemaRegistryServers) in /Users/matt/git/confluent-kafka-dotnet/test/Confluent.SchemaRegistry.Serdes.IntegrationTests/Tests_Protobuf/ProduceConsumeSchemaManyMessages.cs:line 77
[xUnit.net 00:05:00.0445875]     Confluent.SchemaRegistry.Serdes.IntegrationTests.Tests.ConsumePartitionEOF(bootstrapServers: "localhost:9092", schemaRegistryServers: "localhost:8081") [FAIL]
  Failed Confluent.SchemaRegistry.Serdes.IntegrationTests.Tests.ConsumePartitionEOF(bootstrapServers: "localhost:9092", schemaRegistryServers: "localhost:8081") [1 s]
  Error Message:
   Assert.NotNull() Failure
  Stack Trace:
     at Confluent.SchemaRegistry.Serdes.IntegrationTests.Tests.ConsumePartitionEOF(String bootstrapServers, String schemaRegistryServers) in /Users/matt/git/confluent-kafka-dotnet/test/Confluent.SchemaRegistry.Serdes.IntegrationTests/Tests_Avro/ConsumePartitionEOF.cs:line 77

Failed!  - Failed:     3, Passed:    23, Skipped:     0, Total:    26, Duration: 4 m 59 s - /Users/matt/git/confluent-kafka-dotnet/test/Confluent.SchemaRegistry.Serdes.IntegrationTests/bin/Debug/netcoreapp2.1/Confluent.SchemaRegistry.Serdes.IntegrationTests.dll (netcoreapp2.1)
```

src/Confluent.SchemaRegistry.Serdes.Avro/GenericSerializerImpl.cs
```diff
@@ -131,14 +134,26 @@ public async Task<byte[]> Serialize(string topic, GenericRecord data, bool isKey
...
                     {
                         int newSchemaId;
-                        // first usage: register/get schema to check compatibility
-                        if (autoRegisterSchema)
+                        if (useLatestVersion)
                         {
-                            newSchemaId = await schemaRegistryClient.RegisterSchemaAsync(subject, writerSchemaString).ConfigureAwait(continueOnCapturedContext: false);
+                            var latestSchema = await schemaRegistryClient.GetLatestSchemaAsync(subject)
+                                .ConfigureAwait(continueOnCapturedContext: false);
+                            newSchemaId = latestSchema.Id;
```
mhowlett, line 141: the latest schema is thrown away - isn't being used to serialize the data, so aren't there scenarios in which this breaks? (if schema changes is not forward compatible)?

src/Confluent.SchemaRegistry.Serdes.Avro/SpecificSerializerImpl.cs
rayokota: I can add a check that if `use.latest.version` is `true`, then `auto.register.schemas` must be `false`

src/Confluent.SchemaRegistry.Serdes.Avro/GenericSerializerImpl.cs
rayokota: Yes, `use.latest.version=true` is an advanced option.  For the Java client, we can optionally check that the latest schema is backward compatible with the schema of the object being passed.   However, we don't have the code for compatibility checks available in C#.

src/Confluent.SchemaRegistry.Serdes.Avro/SpecificSerializerImpl.cs
mhowlett: that works

src/Confluent.SchemaRegistry.Serdes.Avro/GenericSerializerImpl.cs
mhowlett: I think it's a bit dangerous. Would definitely note this in the property docstrings. maybe "WARNING: There is no check that the latest schema is backwards compatible with the schema used for serialization". Also, I like the idea of changing the config name to `dangerously.use.latest.version`.

rayokota: I'll add the docstring but would like to keep the config name the same as the Java config name.

test/Confluent.SchemaRegistry.Serdes.IntegrationTests/Tests_Avro/AutoRegisterSchemaDisabled.cs
rayokota: I wasn't able to get my environment set up successfully to run these tests, so it would be great if you could take a look sometime.  In any case, since these tests were already failing, and so as to not overload this PR, I'll create a separate issue for this so we can have a separate PR to fix this issue later.

rayokota: Created https://github.com/confluentinc/confluent-kafka-dotnet/issues/1630

src/Confluent.SchemaRegistry.Serdes.Avro/SpecificSerializerImpl.cs
mhowlett: can you also update the Python PR?

src/Confluent.SchemaRegistry.Serdes.Avro/GenericSerializerImpl.cs
mhowlett: ok. i do have a definite preference for having the prefix, but not strong enough to push the point.

mhowlett: can you also update the Python PR? thanks!

test/Confluent.SchemaRegistry.Serdes.IntegrationTests/Tests_Avro/AutoRegisterSchemaDisabled.cs
mhowlett: sounds good - thanks for looking. i only brought it up on the off chance that as the domain expert, you might see what was wrong with very little effort. i'll take a look myself.

src/Confluent.SchemaRegistry.Serdes.Avro/SpecificSerializerImpl.cs
rayokota: Yup, updated the Python PR as well

src/Confluent.SchemaRegistry.Serdes.Avro/GenericSerializerImpl.cs
rayokota: Yup, updated the Python PR as well

mhowlett (approved):
ok, LGTM, thanks!

PR #1639, opened 2021-06-22
Clarify commit information in IConsumer.Close() documentation.

BillyFromAHill:
Seems that better add clarification about auto commit for Close docs. https://github.com/edenhill/librdkafka/issues/1301


mhowlett (approved):
thanks!

Issue #1644, opened 2021-06-30
DESIGN: Unnecessary locking during serialization promotes thread starvation


kmcclellan:
Description
===========
The async SERDES (like `AvroSerializer<T>` and `AvroDeserializer<T>`) use `SemaphoreSlim` under the hood to synchronize access to cached schemas. Every time a Kafka message is serialized or deserialized, [`WaitAsync()` is called](https://github.com/confluentinc/confluent-kafka-dotnet/blob/348c4e34bb996ea0d49e6de71445a28181ccfcfc/src/Confluent.SchemaRegistry.Serdes.Avro/SpecificSerializerImpl.cs#L181) and the rest of the method executes as a continuation.

When used in tandem with `AsSyncOverAsync()`, this can cause thread starvation issues (as alluded to [here](https://github.com/confluentinc/confluent-kafka-dotnet/blob/348c4e34bb996ea0d49e6de71445a28181ccfcfc/src/Confluent.Kafka/SyncOverAsyncSerializer.cs#L29-L44)). My application was running into problems using `Produce(...)` from a few dozen different threads (which all end up blocking while fighting over the semaphore). One solution, of course, is to switch to `ProduceAsync()` instead...

It occurs to me that, for most applications, the vast majority of (de)serialization calls will reuse cached schemas and therefore _ought_ to be able to complete synchronously. Only occasionally will a schema load be in progress such that we'd have to schedule a continuation - a case we could check for synchronously and only `await` conditionally. If so, I suspect we'd see substantial threading efficiency gains and resolve much of the problem with "SyncOverAsync."

This could be accomplished with an additional layer of synchronous locking over the existing `SemaphoreSlim`. A simpler solution may be to use `ConcurrentDictionary<,>` and allowing threads to overwrite cached schemas in those rare competition scenarios. Given the idempotence of the schema registry API, it's not clear to me why we need to explicitly prevent multiple threads from making identical HTTP calls.

Creating an issue to gauge interest in a PR and/or discuss if there's some lurking complexity I'm overlooking here.

Checklist
=========

Please provide the following information:
 - [ ] A complete (i.e. we can run it), minimal program demonstrating the problem. No need to supply a project file.
 - [ ] Confluent.Kafka nuget version.
 - [ ] Apache Kafka version.
 - [ ] Client configuration.
 - [ ] Operating system.
 - [ ] Provide logs (with "debug" : "..." as necessary in configuration).
[...]

PR #1646, opened 2021-07-02
POC: Less locking during async serialization

kmcclellan:
See #1644. Using `AvroSerializer<T>` as a proof of concept. We could do something similar for all the SERDES...


kmcclellan:
[clabot:check]

PR #1656, opened 2021-07-22
Feature/custom reference name strategy

aminzam:
# Description
This feature allows producers to specify a custom subject name strategy for protobuf references. It is done by allowing the callers of the API to provide a Type object for a class that implements IReferenceSubjectNameStrategy. If no such class is provided, the serializer falls back to the current implementation that uses reference name as subject name.

## Background
Currently the serializer can only use the reference name from the protobuf schema as subject name in Schema Registry. This is while Schema Registry allows registering schemas with custom subject names for references. There are cases where this feature is handy specially when auto registration of schemas is disabled. For example in Confluent Cloud where the Schema Registry is shared between all clusters in the environment and isolation of schemas is desirable. With this change clients are able to implement and use their own strategy. For example a strategy that contacts schema registry to find subject names under which references are registered.

## Alternative designs considered
An alternative design is to add another value to the strategy enum for "custom" and add an extra property for a type that provides the subject name resolution. However this means introducing a new config value and diverging more from how Java client works.

## Backward compatability
The default behavior (when no strategy is specified) has stayed intact. This change will break old code only if it explicitly specifies the reference subject name strategy. Considering the fact that there was only one option (default) the chances of this happening is very slim. 

## Other information
This change is related to Confluent Cloud support request 67865 tied to internal ticket FF-548.

### DIFF
--- a/src/Confluent.SchemaRegistry.Serdes.Protobuf/ProtobufSerializerConfig.cs
+++ b/src/Confluent.SchemaRegistry.Serdes.Protobuf/ProtobufSerializerConfig.cs
@@ -209,26 +209,21 @@ namespace Confluent.SchemaRegistry.Serdes
         ///     
         ///     default: ReferenceSubjectNameStrategy.ReferenceName
         /// </summary>
-        public ReferenceSubjectNameStrategy? ReferenceSubjectNameStrategy
+        public Type ReferenceSubjectNameStrategy
         {
             get
             {
                 var r = Get(PropertyNames.ReferenceSubjectNameStrategy);
-                if (r == null) { return null; }
+                if (r == null) { return typeof(DefaultReferenceSubjectNameStrategy); }
                 else
                 {
-                    ReferenceSubjectNameStrategy result;
-                    if (!Enum.TryParse<ReferenceSubjectNameStrategy>(r, out result))
-                        throw new ArgumentException(
-                            $"Unknown ${PropertyNames.ReferenceSubjectNameStrategy} value: {r}.");
-                    else
-                        return result;
+                    return Type.GetType(r);
                 }
             }
             set
             {
                 if (value == null) { this.properties.Remove(PropertyNames.ReferenceSubjectNameStrategy); }
-                else { this.properties[PropertyNames.ReferenceSubjectNameStrategy] = value.ToString(); }
+                else { this.properties[PropertyNames.ReferenceSubjectNameStrategy] = value.AssemblyQualifiedName; }
             }
         }
 
--- a/src/Confluent.SchemaRegistry.Serdes.Protobuf/ProtobufSerializerConfig.cs
+++ b/src/Confluent.SchemaRegistry.Serdes.Protobuf/ProtobufSerializerConfig.cs
@@ -217,7 +217,13 @@ namespace Confluent.SchemaRegistry.Serdes
                 if (r == null) { return typeof(DefaultReferenceSubjectNameStrategy); }
                 else
                 {
-                    return Type.GetType(r);
+                    var type = Type.GetType(r);
+                    if (typeof(IReferenceSubjectNameStrategy).IsAssignableFrom(type)) { return type; }
+                    else
+                    {
+                        throw new ArgumentException(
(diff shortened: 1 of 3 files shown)


aministrator2k:
[clabot:check]

src/Confluent.SchemaRegistry.Serdes.Protobuf/ProtobufSerializerConfig.cs
```diff
@@ -206,29 +206,30 @@ public static class PropertyNames
 
         /// <summary>
         ///     Reference subject name strategy.
-        ///     
-        ///     default: ReferenceSubjectNameStrategy.ReferenceName
+        ///     The provided type must be an implementation of <see cref="IReferenceSubjectNameStrategy" />
+        ///     default: <see cref="DefaultReferenceSubjectNameStrategy" />
         /// </summary>
-        public ReferenceSubjectNameStrategy? ReferenceSubjectNameStrategy
+        public Type ReferenceSubjectNameStrategy
```
mhowlett, line 212: Thanks for the PR @aministrator2k 

A problem with this approach is it breaks backward compatibility, so we can't do it.

I'm also not into the idea of using `Type` and `Activator.CreateInstance` here. You've reminded me why I didn't implement this in the initial scope - the analogous thing to do with other parts of the library is to make a `ProtobufSerializerBuilder` class (in this case we would keep the ability to construct `ProtobufSerializer` without the builder as well) and have a set method for this, but that's such a lot of boilerplate... the other thing we could do is just add two additional optional arguments to the constructor for the `SubjectNameStrategy` and `ReferenceSubjectNameStrategy`. That might be more pragmatic. Don't have a strong opinion ATM.

Finally, you've switched from using a delegate to `IReferenceSubjectNameStrategy`. Is there a good reason for that? My preference is for a delegate unless there is good reason it won't be able to satisfy requirements (unsure).

aministrator2k: Thanks for the feedback @mhowlett.

We share the concern for using `Activator.CreateInstance`. The reasoning behind it for me was to stay as close as possible to Java client design. Same goes for the use of interface.

I like your idea of adding a `ReferenceSubjectNameStrategy` and considered it initially. I gave up that idea because I wasn't sure about the desired direction of the library.

I can work on an alternate pull request by implementing a new `ReferenceSubjectNameStrategy`. The most important thing for us is the support for custom reference names and I see both solutions viable. How does that sound?

mhowlett: > The reasoning behind it for me was to stay as close as possible to Java client design.

better off in this case to stay idiomatic to other parts of the .NET client i think.

i'm hopeful i'll have time to slip in doing this myself, but can't guarantee it. i'm erring on the side of making builder classes ATM, but won't have a clear view on it until i try to do it and see how it feels.

I'm interested in your feedback on:

> Finally, you've switched from using a delegate to IReferenceSubjectNameStrategy. Is there a good reason for that?

aministrator2k: For the use of interface I considered the case of stateful strategies. For example for the case of the strategy we are interested in that have to query schema registry first to grab the full schema and extract registered reference subject names, it is a good idea to cache said schemas in the object.

aministrator2k: > better off in this case to stay idiomatic to other parts of the .NET client i think.

I agree however the java style that supports injection of custom strategies gives users of the library more freedom and control.

aministrator2k: > A problem with this approach is it breaks backward compatibility, so we can't do it.

I was wondering if there is a chance for this PR to go through in its current form. To me the risk for breaking backward compatibility is low because the enumeration contained only one value which makes it improbable for someone to try to set it explicitly. And that is the only case this would break old code since the default behavior is the same.

esskar:
Is there a chance that this gets merged eventually?

PR #1674, opened 2021-08-11
Throw the base exception in the serializer which also takes care of the inner exception

danielmarbach:


danielmarbach:
[clabot:check]

danielmarbach:
No interest in this contribution?

mhowlett:
Can you elaborate on why you why throwing the base exception is better?

IIRC, it's the way it is so you don't get nested Aggregate exceptions (for convenience). If you throw the base exception, you'll need to traverse to get the exception you want (i believe). But maybe you sometimes want the base exception. I don't understand the scenario for that, but as you're requesting the change, I'm guessing there is a reason.

Also, this is a breaking change, so we have to be very careful how we resolve this if there is an issue.

danielmarbach:
This is already a long time I go, so my memories might no longer be entirely correct. I believe I introduced this change because the actual root exception was masked by throwing the inner exception. With the `GetBaseException` method the aggregate exception is properly traversed until the actual root exception is found (even if the inner exception is another aggregate exception) and then that exception is thrown. 

The example in the documentation shows that behavior

https://docs.microsoft.com/en-us/dotnet/api/system.exception.getbaseexception?view=net-6.0

> A chain of exceptions consists of a set of exceptions such that each exception in the chain was thrown as a direct result of the exception referenced in its InnerException property. For a given chain, there can be exactly one exception that is the root cause of all other exceptions in the chain. This exception is called the base exception and its InnerException property always contains a null reference.
> 
> For all exceptions in a chain of exceptions, the GetBaseException method must return the same object (the base exception).
> 
> Use the GetBaseException method when you want to find the root cause of an exception but do not need information about exceptions that may have occurred between the current exception and the first exception.

I'd argue that is the thing a consumer of the exception is actually interested in to be able to quickly troubleshoot what is going on. On top of that I would say it might even make sense to combine it with `ExceptionDispatchInfo` to properly freeze the stack trace.

danielmarbach:
Closing because of the lack of interest or interaction.

PR #1718, opened 2021-11-04
Enhance HTTP Authentication Support

henrydaly:
Decouple HTTP Basic authentication from RestService HttpClient. Provide
open ended interface for user-implemented authentication schemes.
Implement simple HTTP Bearer authentication solution for token-based
authorization. 

Motivation: my team is running a Schema Registry instance behind a proxy API with token-based authentication. This PR will allow us and others to more easily hook into the HTTP authentication mechanism in cases where we're not running a vanilla Schema Registry deployment or don't want to use basic authentication


src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -128,7 +128,12 @@ private static SubjectNameStrategyDelegate GetValueSubjectNameStrategy(IEnumerab
         /// <param name="config">
         ///     Configuration properties.
         /// </param>
-        public CachedSchemaRegistryClient(IEnumerable<KeyValuePair<string, string>> config)
+        /// <param name="authenticator">
+        ///     The client authenticator
+        ///
+        ///     default: null
+        /// </param>
+        public CachedSchemaRegistryClient(IEnumerable<KeyValuePair<string, string>> config, IHttpAuthenticator authenticator = null)
```
mhowlett, line 136: this isn't backwards compatible without a recompile, so I think it would be better to get rid of the default value for `authenticator` and make a second constructor that only takes `config` which should call this constructor.

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -193,6 +198,9 @@ public CachedSchemaRegistryClient(IEnumerable<KeyValuePair<string, string>> conf
             {
                 throw new ArgumentException($"Invalid value '{basicAuthSource}' specified for property '{SchemaRegistryConfig.PropertyNames.SchemaRegistryBasicAuthCredentialsSource}'");
             }
+            if (username != null && password != null && authenticator == null)
```
mhowlett, line 201: coding style is to always use `{ }`'s

src/Confluent.SchemaRegistry/Rest/Authentication/BearerHttpAuthenticator.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Net.Http.Headers;
+
+namespace Confluent.SchemaRegistry
+{
+    /// <summary>
+    ///     A HTTP authenticator implementing the 'Bearer' scheme
+    ///
+    ///     See: https://datatracker.ietf.org/doc/html/rfc6750
+    /// </summary>
+    public class BearerHttpAuthenticator : IHttpAuthenticator
```
mhowlett, line 26: we don't want to include this specific case because it relates to a setup that we don't document or talk about.

src/Confluent.SchemaRegistry/Rest/RestService.cs
```diff
@@ -282,6 +282,8 @@ private HttpRequestMessage CreateRequest(string endPoint, HttpMethod method, par
                 content.Headers.ContentType.CharSet = string.Empty;
                 request.Content = content;
             }
+            if (httpAuthenticator != null)
```
mhowlett, line 285: we always use `{ }`'s

src/Confluent.SchemaRegistry/Rest/Authentication/BearerHttpAuthenticator.cs
mhowlett: i mean we don't want this file.

src/Confluent.SchemaRegistry/Rest/Authentication/IHttpAuthenticator.cs
```diff
@@ -0,0 +1,34 @@
...
+// Refer to LICENSE for more information.
+
+using System.Net.Http.Headers;
+
+namespace Confluent.SchemaRegistry
+{
+    /// <summary>
+    ///     An interface defining Schema Registry HTTP client authentication schemes.
+    /// </summary>
+    public interface IHttpAuthenticator
```
mhowlett, line 24: I think a more specific name here is better `IAuthenticationHeaderValueProvider`. Also update the comment.

src/Confluent.SchemaRegistry/Rest/Authentication/BasicHttpAuthenticator.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Text;
+
+namespace Confluent.SchemaRegistry
+{
+    /// <summary>
+    ///     A HTTP authenticator implementing the 'Basic' scheme.
+    ///
+    ///     See: https://datatracker.ietf.org/doc/html/rfc7617
+    /// </summary>
+    public class BasicHttpAuthenticator : IHttpAuthenticator
```
mhowlett, line 28: `BasicaAuthenticaitonHeaderValueProvider` i think. I know it's a mouth full, but it's going to be little used and i think more descriptive wins.

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
mhowlett: you should do more checking - throw an argument exception on all invalid input cases.

mhowlett:
thanks for the change, lgtm after comments addressed.

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -194,6 +197,29 @@ public CachedSchemaRegistryClient(IEnumerable<KeyValuePair<string, string>> conf
...
+            if (authenticationHeaderValueProvider != null)
+            {
+                if (username != null || password != null)
+                {
+                    throw new ArgumentException($"Invalid authentication header value provider configuration: Cannot specify both custom provider and username/password");
+                }
+            }
+            else
+            {
+                if (username != null && password == null)
```
henrydaly, line 209: I've added checks on username/password being null. This didn't exist in the past (no exception thrown when one is null and the other isn't). Not sure if we want to add this going forward, but this should cover all potential "invalid argument" cases

henrydaly:
Thanks for the review @mhowlett, I've addressed all your comments. Let me know what you think about the new null checks on username/password

henrydaly:
@mhowlett have you had an opportunity to look at my latest changes?

src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
mhowlett: could be written slightly neater, but logic lgtm. lgtm.

mhowlett:
thanks for the change @henrydaly - LGTM. this'll be in v1.9.

PR #1719, opened 2021-11-09
Change Slack URL

rmoff:
If users aren't signed up then the existing URL just takes them to a login page with no indication of how to sign up.


mhowlett (approved):
LGTM

PR #1720, opened 2021-11-09
Add links out to Confluent Developer resources

rmoff:


mhowlett (approved):
LGTM.

I notice the example you link to references version 1.6.3, should be 1.8.2. Without a relatively easy systematic process for keeping it updated, this is going to tend to be out of date.

PR #1725, opened 2021-11-22
Improve serialization memory management

ChristopherHaas:
Currently, the `ISerializer` and `IAsyncSerializer` interfaces force an implementation to return a byte array with exact size. This approach effectively makes pooling impossible and often requires expensive memory copies and additional allocations in serializer implementations. See ProtobufSerializer as an example.

This PR changes the memory management for serialization. Instead of relying on serializer implementations to return a byte array, they are given an IBufferWriter<byte> to write to.

`void Serialize(T data, SerializationContext context, IBufferWriter<byte> bufferWriter);`

The producer implementation will then supply a buffer writer implementation that internally uses ArrayPool<byte> to allocate the memory. Users can overwrite this implementation by setting a custom ISerializationBufferProvider.

### DIFF
--- a/src/Confluent.Kafka/ArrayPoolBufferWriter.cs
+++ b/src/Confluent.Kafka/ArrayPoolBufferWriter.cs
@@ -1,0 +1,135 @@
+﻿// Copyright 2016-2018 Confluent Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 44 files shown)


mhowlett:
Nice work @ChristopherHaas - apologies for the delayed response.

One important thing to note is we can't break the API - the old `ISerializer<T>` is going to need to stay - we'll need to support both types of serializer.

The other thing worth noting is due to the size, there could be a delay in getting this properly reviewed / merged (as you have already no doubt noticed). The time taken will depend on how much I agree with you. If there is too much disagreeing, it'll likely stall. Based on a quick look, it is seeming good to me so far. 

So I'd encourage you to make changes to make this non-breaking, and hopefully I can get to a proper review before too long.

PR #1751, opened 2022-01-21
 dont allow string key is empty or whitespaces | also fix problem Index 0 out of bounds for length 0

devopg:
https://github.com/confluentinc/ksql/issues/8638


mhowlett:
null and "" are two distinct, valid things that the utf8 serializer should support. also, "" should definitely not serialize as null, that would be surprising behavior.

mhowlett:
where does the "also fix problem Index 0 out of bounds for length 0" error occur. inside this client code?

devopg:
yes . ksql throw exception if try process key  = empty spaces or key = empty (not null)

PR #1760, opened 2022-02-10
Fixes typo

yang-xiaodong:


mhowlett:
thanks @yang-xiaodong !

this is a breaking change. fortunately, i believe we can have multiple enum values with the same int value, so this gives a path to fixing it now, and not waiting to the next version bump. can you change this to the following?:

```
        /// <summary>
        ///     Group coordinator load in progress
        /// </summary>
        [Obsolete("Superseded by GroupLoadInProgress")]
        GroupLoadInProress = 14,

        /// <summary>
        ///     Group coordinator load in progress
        /// </summary>
        GroupLoadInProgress = 14,
```

yang-xiaodong:
Ok, nice practice!

mhowlett:
thanks!

PR #1763, opened 2022-02-16
netcoreapp2.1 -> net5.0

mhowlett:
netcoreapp2.1 is now out of long term support.

Moving to net5.0 not net6.0 as one of the test projects hangs on net6.0. This seems to be a problem with the test framework not the tests according to some quick research. Couldn't quickly figure out how to fix it.

Note: This PR does not affect framework compatibility of any of the libraries.

CancellationTokenRegistration and CancellationToken are structs, so the newer compiler warns me, so can't be null.


mhowlett:
i didn't change the Makefile, which is used for travis builds because we're moving away from that (and there probably isn't a newer version available there anyway).

jliunyu (approved):

PR #1782, opened 2022-03-23
Support using array segments for serializers

webmakersteve:
Adjusts the primary serializer interface to return an ArraySegment-wrapped byte array in favor of the byte array itself. It then uses the resulting count and offset fields to be passed into the produce call. The new interface also supports a Release hook which can be used after the produce call has happened.

The primary use case is to allow for pooled arrays to be written to, or recycled streams' internal buffers. This helps a lot when writing data greater than 85K which would ordinary require LOH allocations and greatly impact the performance of the service. 

I made the adjustment so there should be no breaking changes - the builder interfaces just "upgrade" the old variant serializers by wrapping them. The producer class internally uses the new interfaces only. The release methods are no-op on the wrapped serializers.

### DIFF
--- a/src/Confluent.Kafka/DependentProducerBuilder.cs
+++ b/src/Confluent.Kafka/DependentProducerBuilder.cs
@@ -37,23 +37,23 @@ namespace Confluent.Kafka
         /// <summary>
         ///     The configured key serializer.
         /// </summary>
-        public ISerializer<TKey> KeySerializer { get; set; }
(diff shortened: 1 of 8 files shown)


yinzara:
Just noting I am not an official collaborator, just reading your issue and PR.

This adds capabilities without modifying any of the existing serializers to take advantage of this capability. I'd love to see how this would change the AvroSerializer or ProtobufSerializer etc to now be more efficient or better.

PR #1800, opened 2022-04-10
Use rd_kafka_message_errstr on consume error

mhowlett:


test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Poll_MessageError.cs
```diff
@@ -0,0 +1,66 @@
...
+                ErrorCode code = ErrorCode.NoError;
+                try
+                {
+                    consumer.Subscribe(nonExistantTopic);
+                    consumer.Consume(TimeSpan.FromSeconds(10));
+                }
+                catch (ConsumeException e)
+                {
+                    code = e.Error.Code;
+                    Assert.NotNull(e.Error.Reason);
```
edenhill, line 56: nit: It'd be good if there was also a `Assert.NotTheSamePlease(e.Error.Reason, code2str(e.Error.Code))` since this test would also have passed prior to adding message_errstr.

edenhill (approved):
LGTM

PR #1805, opened 2022-04-13
Change framework order to avoid omnisharp issues

mhowlett:
It seems like omnisharp (IDE intellisense helper) uses the first listed framework in TargetFrameworks. It makes life better therefore to specify a maximally compatible framework first. This change works around issues i'm having with vscode on osx.


jonchiu (approved):
LGTM

PR #1806, opened 2022-04-13
Replace obsolete call

mhowlett:


jonchiu (approved):
I assume it's because of this: https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.codebase?view=net-6.0

mhowlett:
yes

PR #1812, opened 2022-05-13
-> v1.9.0-RC9

mhowlett:


examples/Producer/Producer.csproj
```diff
@@ -9,8 +9,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <!-- nuget package reference: <PackageReference Include="Confluent.Kafka" Version="1.9.0-RC7" /> -->
```
edenhill, line 12: this is an actual change, not sure what it means

edenhill (approved):

examples/Producer/Producer.csproj
mhowlett: oh whoops, thanks.

PR #1825, opened 2022-06-01
Do not calculate statistics json length as it is already provided by librdkafka

VladimirTyrin:
No need to traverse long strings


Ihavesoul:
Cool thing, up!

mhowlett (approved):
thanks! LGTM.

PR #1830, opened 2022-06-08
[docs] Clarify StoreOffset overload behaviour difference

sazarubin:
Currently difference of behaviour between StoreOffset overloads is [unclear](/confluentinc/confluent-kafka-dotnet/issues/1380).

I extended existing doc comments with clarifications.

Hope this helps!

### DIFF
--- a/src/Confluent.Kafka/IConsumer.cs
+++ b/src/Confluent.Kafka/IConsumer.cs
@@ -328,6 +328,11 @@ namespace Confluent.Kafka
         ///     `auto.commit.interval.ms` (and
         ///     `enable.auto.commit`) or manual offset-less
         ///     commit().
+        ///
+        ///     This overload <b>does</b> increment offset, so
+        ///     calling it after event consumption with
+        ///     `ConsumeResult` <b>does</b> move offset
+        ///     to the next event
         /// </summary>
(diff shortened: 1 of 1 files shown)


sazarubin:
Uhmm, I'm not willing to leave my address just for doc clarification 🤦 

I'll be glad if the change gets into mainline by merging this PR or by opening a new one from existing contributors

sazarubin:
Hi, it's me again. 

So, as you supposed, KIP-320 changes really messed things up. Code that used `StoreOffset`, just silently broke when we upgraded Confluent.Kafka from 2.0.2 to 2.1.1.

The reason is that the `leaderEpoch` parameter isn't required by the `TopicPartitionOffset` ctor but at the same time without it `StoreOffset` doesn't do anything without throwing any exception or logging any errors.

I was able to figure it out when switched debug logs on and saw a bunch of `[thrd:main]: Topic <topic_name> [<partition_number>]: stored offset INVALID (leader epoch -1), committed offset 873041 (leader epoch -1): not including in commit` messages.

I think it should be at least documented and/or marked as breaking in release notes, though I'm not sure what the real solution should be.

sazarubin:
Should I post a separate issue?

PR #1835, opened 2022-06-09
Fixed memory leak in SendOffsetsToTransaction

mhowlett:


edenhill (approved):
LGTM

This should probably be mentioned in the release notes.

PR #1836, opened 2022-06-09
fix: freed eventPtr in AdminClient

emasab:


mhowlett (approved):
nice find, thanks @emasab

PR #1840, opened 2022-06-17
Feature/add unload

mrkwllce:
In certain scenarios, the native DLLs need to be unloaded from the process. When hosting in IIS specifically, w3wp retains a reference to the DLL and causes future calls to Initialize to fail.


.travis.yml
```diff
@@ -8,13 +8,13 @@ matrix:
   include:
   - os: linux
     dist: trusty # Ubuntu 14.04
-    dotnet: 2.1.300
+    dotnet: 5.0.408
```
mhowlett, line 11: can you revert the changes to .travis.yml? it will clash with another PR, and also we're deprecating travis.

src/Confluent.Kafka/Library.cs
```diff
@@ -114,6 +114,19 @@ public static bool Load()
...
+        ///     Unloads the native librdkafka library from the process.
+        /// </summary>
+        /// <returns>
+        ///     true if librdkafka was unloaded, false if not unloaded
+        /// </returns>
+        /// <remarks>
+        ///     You will not typically need to call this method - only needed under
+        ///     certain hosting requirements, such as IIS.
+        /// </remarks>
+        public static bool Unload()
```
mhowlett, line 127: can you write a unit test for this?

PR #1842, opened 2022-06-17
Don't exit the pool loop if an error as local timeout happens

emasab:


test/Confluent.Kafka.UnitTests/Admin/CreateTopicsError.cs
```diff
@@ -0,0 +1,70 @@
...
+// Refer to LICENSE for more information.
+
+using Xunit;
+using System;
+using System.Collections.Generic;
+using Confluent.Kafka.Admin;
+
+namespace Confluent.Kafka.UnitTests
+{
+    public class CreateTopicsErrorTests
```
mhowlett, line 24: I think i would just make this `AdminClient` - we don't have any other admin client unit tests, and are unlikely to make many (because there's not much you can do).

great idea to make this a unit test, not integration test.

test/Confluent.Kafka.UnitTests/Admin/CreateTopicsError.cs
```diff
@@ -0,0 +1,70 @@
...
+                    new TopicSpecification
+                    {
+                        Name = "my-topic-2",
+                        ReplicationFactor = 1,
+                        NumPartitions = 1
+                    }
+                };
+
+                // null topics
+                await Assert.ThrowsAsync<ArgumentNullException>(() =>
```
mhowlett, line 48: this would ideally be a different method "NullTopic" i think.

test/Confluent.Kafka.UnitTests/Admin/CreateTopicsError.cs
```diff
@@ -0,0 +1,70 @@
...
+using System;
+using System.Collections.Generic;
+using Confluent.Kafka.Admin;
+
+namespace Confluent.Kafka.UnitTests
+{
+    public class CreateTopicsErrorTests
+    {
+        [Fact]
+        public async void Errors()
```
mhowlett, line 27: maybe call this "LocalTimeout"

test/Confluent.Kafka.UnitTests/Admin/CreateTopicsError.cs
```diff
@@ -0,0 +1,70 @@
...
+                        adminClient.CreateTopicsAsync(
+                            new List<TopicSpecification>{topic},
+                            options)
+                    );
+                    Assert.Equal("Failed while waiting for controller: Local: Timed out", ex.Message);
+                }
+            }
+        }
+    }
+}
```
mhowlett, line 70: nit: the convention in this repo is to always have a new line at the end

PR #1881, opened 2022-08-24
A couple of consumer integration tests

mhowlett:


test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Subscription_Regex.cs
```diff
@@ -0,0 +1,86 @@
...
+            string prefix = "dotnet_test_" + Guid.NewGuid().ToString() + "_";
+
+            using (var topic1 = new TemporaryTopic(prefix, bootstrapServers, 1))
+            using (var topic2 = new TemporaryTopic(prefix, bootstrapServers, 1))
+            using (var consumer = new ConsumerBuilder<byte[], byte[]>(consumerConfig).Build())
+            {
+                Util.ProduceNullStringMessages(bootstrapServers, topic1.Name, 100, 100);
+                Util.ProduceNullStringMessages(bootstrapServers, topic2.Name, 100, 100);
+
+                consumer.Subscribe("^" + prefix + "*");
```
milindl, line 57: I feel we should use either `"^" + prefix + ".*$"` or `"^" + prefix` since just `*` is a bit confusing (it won't match anything, but the test does work since there's no `$`)

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Subscription_Regex.cs
```diff
@@ -0,0 +1,86 @@
...
+
+                // Wait for assignment
+                consumer.Consume(TimeSpan.FromSeconds(10));
+                Assert.Equal(2, consumer.Assignment.Count);
+
+                // Create new matching topic, wait long enough for metadata to be
+                // discovered and corresponding rebalance.
+                using var topic3 = new TemporaryTopic(prefix, bootstrapServers, 1);
```
milindl, line 65: Maybe use a loop here, would be less repeated stuff

milindl (commented):
Looks good to me besides a few things, all minor

test/Confluent.Kafka.IntegrationTests/Tests/Consumer_Subscription_Regex.cs
mhowlett: yeah, good idea, but what i have'll do.

milindl (approved):

PR #1884, opened 2022-08-30
Fix schema registry integration test (SslAuth) by regenerating secrets

milindl:
The certificate in the current schema registry uses an outdated hash (SHA1)
which is deprecated at most places and simply doesn't work at others.

This replaces those with new secrets (the certificate uses SHA256). Also
adds a script to regenerate secrets in case of expiry or deprecation of
SHA256.


milindl:
I would also be very glad if someone with a Mac could run this test too, I only have a linux system, amd since the CI only seems to run unit tests and not integration tests.

mhowlett (approved):
LGTM & works on my Apple M1.

PR #1886, opened 2022-08-31
Remove incomplete ProtobufWithReferences test

mhowlett:
This test was never completed. The case is covered by the ProduceConsumeExternalRefProtobuf Serde integration test. So we can just delete this.


milindl (approved):
LGTM, but it's better if we can delete all these also since they are not being used

Confluent.SchemaRegistry.IntegrationTests/Tests/Person.cs
Confluent.SchemaRegistry.IntegrationTests/Tests/PersonName.cs
Confluent.SchemaRegistry.IntegrationTests/Tests/proto/person.proto
Confluent.SchemaRegistry.IntegrationTests/Tests/proto/person.proto

PR #1893, opened 2022-09-12
Add additional error codes

mhowlett:


mhowlett:
c.f. https://github.com/edenhill/librdkafka/blob/master/src/rdkafka.h#L400

milindl (approved):

PR #1900, opened 2022-09-19
Added TLS auth example

mhowlett:


examples/TlsAuth/Program.cs
```diff
@@ -0,0 +1,61 @@
...
+using Confluent.Kafka;
+
+
+namespace Confluent.Kafka.Examples.TlsAuth
+{
+    public class Program
+    {
+        public static async Task Main(string[] args)
+        {
+            if (args.Length < 4)
```
milindl, line 30:
```
if ((args.Length < 4) || (args[2] == "auth" && args.Length < 6))
```

milindl (approved):
looks good to me, I followed the entire setup on a debian(broker)/ubuntu(client) system, worked smoothly for me. 
Minor comment for better error message.

PR #1901, opened 2022-09-20
Added NotUserConfigurableCheck and Configuration example

mhowlett:


mhowlett:
`NotUserConfigurableCheck` is really a workaround for a poor API, but we've got what we've got and it's not too bad - it's fairly neat doing config like this.

emasab (approved):

PR #1906, opened 2022-09-22
Config stuff

mhowlett:
an idea for how to set up config for your new EoS example @emasab

### DIFF
--- a/examples/ExactlyOnceBasic/Program.cs
+++ b/examples/ExactlyOnceBasic/Program.cs
@@ -1,0 +1,49 @@
+using System;
+using Microsoft.Extensions.Configuration;
+using Confluent.Kafka;
+
+
+/// <remarks>
+///     Set Sasl credentials by adding to appsettings.json or via the env vars:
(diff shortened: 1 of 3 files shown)


examples/ExactlyOnceBasic/Program.cs
```diff
@@ -0,0 +1,49 @@
...
+
+
+/// <remarks>
+///     Set Sasl credentials by adding to appsettings.json or via the env vars:
+///     KAFKA_EXAMPLE__Common__SaslUsername
+///     KAFKA_EXAMPLE__Common__SaslPassword
+/// </remarks>
+class ProcessConfig
+{
+    public static T GetConfig<T>(string configSection)
```
mhowlett, line 13: on reflection i think it would be better to instantiate the ProcessConfig class, and not re-read the config every time it's needed, as i'm doing here. i originally figured this was no big deal to be inefficient as it doesn't matter... but since this is serving as an example, we probably should do it properly.

PR #1907, opened 2022-09-23
Exactly once semantics basic implementation example

emasab:
This is the example used to test long running EOS tests during the development of this librdkafka [fix](https://github.com/edenhill/librdkafka/pull/3971). It contains the needed checks for retriable, abortable or fatal error the producer can throw when using transactions.

### DIFF
--- a/examples/ExactlyOnce/Program.cs
+++ b/examples/ExactlyOnce/Program.cs
@@ -74,13 +74,13 @@ namespace Confluent.Kafka.Examples.ExactlyOnce
             switch (command)
             {
                 case "create_words":
-                    await new WordCreator().Execute(configuration, transactionalId, adminClient, cts);
+                    await WordCreator.Execute(configuration, transactionalId, adminClient, cts);
                     break;
                 case "reverse_words":
-                    await new WordReverser().Execute(configuration, transactionalId, adminClient, cts);
+                    await WordReverser.Execute(configuration, transactionalId, adminClient, cts);
                     break;
                 case "print_words":
-                    await new WordPrinter().Execute(configuration, adminClient, cts);
+                    await WordPrinter.Execute(configuration, adminClient, cts);
                     break;
                 default:
                     logger.LogError("Unknown command");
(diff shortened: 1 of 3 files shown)


examples/ExactlyOnceBasic/Program.cs
```diff
@@ -0,0 +1,680 @@
...
+        public static async Task Main(string[] args)
+        {
+            CancellationTokenSource cancelled = new CancellationTokenSource();
+            CancellationTokenSource completed = new CancellationTokenSource();
+            Console.CancelKeyPress += (_, e) =>
+            {
+                e.Cancel = true; // prevent the process from terminating.
+                cancelled.Cancel();
+            };
+            AppDomain.CurrentDomain.ProcessExit += (_, e) =>
```
mhowlett, line 79: seems too fancy to use AppDomain. Why not just wait on the cancellation token at the end of Main? `cts.Token.WaitHandle.WaitOne();`

examples/ExactlyOnceBasic/Program.cs
```diff
@@ -0,0 +1,680 @@
...
+            }
+        }
+
+        /// <summary>
+        ///     Main method.
+        /// </summary>
+        public static async Task Main(string[] args)
+        {
+            CancellationTokenSource cancelled = new CancellationTokenSource();
+            CancellationTokenSource completed = new CancellationTokenSource();
```
mhowlett, line 73: very much doubt you need both of these.

examples/ExactlyOnceBasic/Program.cs
```diff
@@ -0,0 +1,680 @@
...
+            var wordsConfig = generalConfig.GetSection("Words");
+            var reversedWordsConfig = generalConfig.GetSection("ReversedWords");
+
+            var wordsTopicName = wordsConfig.GetValue<string>("TopicName");
+            var wordsTopicPartitions = wordsConfig.GetValue<int>("TopicPartitions");
+
+            var reversedWordsTopicName = reversedWordsConfig.GetValue<string>("TopicName");
+            var reversedWordsTopicPartitions = reversedWordsConfig.GetValue<int>("TopicPartitions");
+
+            var adminClient = new AdminClientBuilder(producerConfig).Build();
```
mhowlett, line 114: It would be more proper to have separate admin client config section, even though you can just reuse the producerConfig.

examples/ExactlyOnceBasic/Program.cs
```diff
@@ -0,0 +1,680 @@
...
+            var transactionalId = args[1];
+
+            IConfiguration configuration = new ConfigurationBuilder()
+                .SetBasePath(Directory.GetCurrentDirectory())
+                .AddJsonFile("./appsettings.json")
+                .Build();
+
+            var producerConfig = configuration.GetSection("Producer").Get<ProducerConfig>().ThrowIfContainsNonUserConfigurable();
+            var consumerConfig = configuration.GetSection("Consumer").Get<ConsumerConfig>().ThrowIfContainsNonUserConfigurable();
+            var generalConfig = configuration.GetSection("General");
```
mhowlett, line 104: It would be nicer to define a class `Config`, and bind to that (in the same way as ProducerConfig etc works above) rather than writing it out explicitly.

examples/ExactlyOnceBasic/Program.cs
mhowlett: and call the one you do use `cancellationTokenSource` or `cts`, not `cancelled`.

examples/ExactlyOnceBasic/Program.cs
```diff
@@ -0,0 +1,680 @@
...
+            {
+                case "create_words":
+                    new WordsCreator
+                    {
+                        OutputTopic = wordsTopicName,
+                        ProduceRate = wordsConfig.GetValue<TimeSpan>("ProduceRate"),
+                        RetryInterval = retryInterval,
+                        LocalTransactionOperationTimeout = localTransactionOperationTimeout,
+                        ProducerBuilder = producerBuilder,
+                        Log = Program.Log,
```
mhowlett, line 134: I think `Logger` is a better name for this property.

examples/ExactlyOnceBasic/Program.cs
```diff
@@ -0,0 +1,680 @@
...
+                        CTS = cancelled
```
mhowlett, line 135: I think CancellationTokenSource is a better name for this property.

examples/ExactlyOnceBasic/Program.cs
```diff
@@ -0,0 +1,680 @@
...
+
+            var command = args[0];
+            var transactionalId = args[1];
+
+            IConfiguration configuration = new ConfigurationBuilder()
+                .SetBasePath(Directory.GetCurrentDirectory())
+                .AddJsonFile("./appsettings.json")
+                .Build();
+
+            var producerConfig = configuration.GetSection("Producer").Get<ProducerConfig>().ThrowIfContainsNonUserConfigurable();
```
mhowlett, line 102: I would fully encapsulate the configuration logic inside the WordCreator and WordReverser classes - don't do it outside and pass it in.

I'd completely use binding - don't get values directly. I.e. define WordCreatorConfig and WordReverserConfig classes that line up to the config properties and bind (as per the ProduceConfig etc here).

put the producer and consumer config inside these (not outside). this will result in duplicate config, yes, but i don't think it's too bad, and i would err on the side of this being better. Binding works with nested classes - just have a `Producer` property of type `ProducerConfig` of your `WordCreatorConfig` class, and it'll all automatically work.

examples/ExactlyOnceBasic/Program.cs
```diff
@@ -0,0 +1,680 @@
...
+    /// <summary>
+    ///     This transactional producer allows to use a producer an consumer for transactions,
+    ///     by recreating the producer when a fatal error happens, sending offsets to transaction
+    ///     and handling aborts and retries correctly.
+    /// </summary>
+    public class TransactionalProducer<K, V> : IDisposable
+    {
+        public string OutputTopic { get; set; }
+
+        public Action<string> Log { get; set; }
```
mhowlett, line 173: Logger.

examples/ExactlyOnceBasic/Program.cs
```diff
@@ -0,0 +1,680 @@
...
+
+        public TimeSpan RetryInterval { get; set; }
+
+        public TimeSpan LocalTransactionOperationTimeout { get; set; }
+
+        public ProducerBuilder<K, V> ProducerBuilder { get; set; }
```
mhowlett, line 179: I wouldn't maintain a builder class like this - create a builder class every time you need to create the producer is how i intended. it is lightweight.

examples/ExactlyOnceBasic/Program.cs
emasab: I've added it because, you don't, it handles SIGINT correctly but not SIGTERM and doesn't exit gracefully. It happens the same if you don't wait for completedCTS.

examples/ExactlyOnceBasic/Program.cs
emasab: I've called it Log because this is not a complete logger, it's an action that corresponds to the Program method.
It seems better to call `Log("message")`  than `Logger("message")`.

examples/ExactlyOnceBasic/Program.cs
emasab: See previous reply, Renamed

examples/ExactlyOnceBasic/Program.cs
emasab: See previous reply

examples/ExactlyOnce/Program.cs
```diff
@@ -14,789 +14,677 @@
...
+
+        private ILogger logger = Program.LoggerFactory.CreateLogger<TransactionalProducer<K, V>>();
+
+        public TimeSpan RetryInterval { get; set; }
+
+        public TimeSpan LocalTransactionOperationTimeout { get; set; }
+
+        public ProducerConfig ProducerConfig { get; set; }
+
+        public IConsumer<K, V> Consumer { get; set; }
```
mhowlett, line 168: a thing called "producer" should not have access to a thing called "consumer" - indication of bad structure.

examples/ExactlyOnce/Program.cs
```diff
@@ -14,789 +14,677 @@
...
+                        Name = name,
+                        NumPartitions = partitions
+                    }
+                });
+                logger.LogInformation($"Created topic {name}");
+            }
+            catch (CreateTopicsException)
+            {
+                // Continue if topic already exists
+                logger.LogInformation($"Topic {name} already exists");
```
mhowlett, line 67: this is not necessarily the case.

examples/ExactlyOnce/Program.cs
```diff
@@ -14,789 +14,677 @@
...
-            public ClientSession<string, int, int, int, Empty, IFunctions<string, int, int, int, Empty>> Session { get; set; }
-
-            public void Dispose()
+        /// <summary>
+        ///     Retry callback that stops retries only if
+        ///     the error is fatal or abortable.
+        /// </summary>
+        private void RetryUnlessFatalOrAbortable(string operation, KafkaException e)
+        {
+            var TxnRequiresAbort = e is KafkaTxnRequiresAbortException;
```
mhowlett, line 207: these lines are duplicated a bit.
I also don't think this function is needed, RetryIfNeeded will do.

examples/ExactlyOnce/Program.cs
```diff
@@ -14,789 +14,677 @@
...
-        ///     Name of the consumer group used when loading count state
-        ///     from Topic_Counts.
+        ///     Retry callback that stops retries if
+        ///     the error is fatal or abortable or not retriable.
         /// </summary>
-        const string ConsumerGroup_LoadState = "load-state-consumer-group";
-
-
-        class FasterState : IDisposable
+        private void RetryIfNeeded(string operation, KafkaException e)
```
mhowlett, line 188: crappy name.

mhowlett:
i'll add a commit to this pr later today.

mhowlett:
i'm going to put this on hold for now, and shift something else.

examples/ExactlyOnce/Program.cs
```diff
@@ -15,788 +15,698 @@
...
+                default:
+                    logger.LogError("Unknown command");
+                    break;
+            }
+
+            completedCancellationTokenSource.Cancel();
+        }
+    }
 
+    public interface ITransactionalProcess<KC, VC, KP, VP>
```
mhowlett, line 87: To my eye, the current iteration is much better, maybe good enough to go with.

I do think it could be improved though. I wouldn't use interfaces - i would specify the processing method as a delegate, and have config properties/policies to control committing. Making this API flexible gets interesting though - the more flexible you make it the more general / complex the api becomes.

Also, I think the consumer (which may not always be used) should be completely encapsulated in the Processor.

examples/ExactlyOnce/Program.cs
```diff
@@ -15,788 +15,698 @@
...
+        ///     must commit or abort any other transactions it has
+        ///     ongoing.
         /// </summary>
-        const string Topic_Counts = "counts";
+        public void EndTransaction(List<Message<KP, VP>> batch, bool committed);
+    }
 
+    /// <summary>
+    ///     This transactional processor allows to use a producer an consumer for transactions,
+    ///     by recreating the producer when a fatal error happens, sending offsets to transaction
```
mhowlett, line 119: i'm unsure if we want to recreate the producer always. it's kind of papering over a real problem, it entered a fatal state for a reason. maybe a config property controls this behavior, and that makes it ok.

examples/ExactlyOnce/Program.cs
```diff
@@ -15,788 +15,988 @@
...
-        const string Topic_InputLines = "lines";
+        public Action<DeliveryReport<K, V>> DeliveryHandler { get; set; } = null;
+    }
+
+    /// <summary>
+    ///     The transactional processor allows to use a producer an consumer for transactions,
+    ///     by recreating the producer when a fatal error happens, sending offsets to transaction
+    ///     and handling aborts and retries correctly.
+    /// </summary>
+    public sealed class TransactionalProcessor<KC, VC, KP, VP> : IDisposable
```
mhowlett, line 129: i think this is getting big enough that it'd be better to split into different files. `Processor` can probably have its own file. maybe keep everything else in this file.

examples/ExactlyOnce/Program.cs
```diff
@@ -15,788 +15,988 @@
...
+    {
+        private static readonly ILogger logger = Utils.LoggerFactory.CreateLogger<TransactionalProcessor<KC, VC, KP, VP>>();
```
mhowlett, line 131: maybe utils in another file...

examples/ExactlyOnce/Program.cs
```diff
@@ -15,788 +15,988 @@
...
-                this.Session = this.Store.NewSession(funcs);
+            get
+            {
+                return consumerConfig;
             }
+            set
+            {
+                if (value != null && consumer == null)
+                {
+                    consumerConfig = value;
```
mhowlett, line 205: doing this sort of thing is far too tricky/clever !!! 

the producer and consumer don't need to be class properties - they are only needed in the context of the RunLoop - create them there.

examples/ExactlyOnce/Program.cs
```diff
@@ -15,788 +15,988 @@
...
-
         /// <summary>
-        ///     Generate example line input data (using the source code of this example as the source!).
+        ///     Commits transaction's produced and consumed messages.
+        ///     Aborts transaction if an abortable exception is thrown,
+        ///     recreates the producer if a fatal exception is thrown or
+        ///     retries each operation that is throwing a retriable exception.
         /// </summary>
-        static async Task Generator_LineInputData(string brokerList, string clientId, CancellationToken ct)
+        public void CommitKafkaTransaction(ICollection<ConsumeResult<KC, VC>> inputBatch,
```
mhowlett, line 374: should not be public

examples/ExactlyOnce/Program.cs
```diff
@@ -15,788 +15,988 @@
...
+        public IsCommitNeededDelegate IsCommitNeeded { get; set; }
+
+        /// <summary>
+        ///     Called at the end of transaction with the Kafka batch 
+        ///     and the <paramref name="committed" /> param that says if
+        ///     the transaction completed successfully. Here the application
+        ///     must commit or abort any other transactions it has
+        ///     ongoing.
+        /// </summary>
+        public EndTransactionDelegate EndTransaction { get; set; }
```
mhowlett, line 293: this is a good idea 👍

examples/ExactlyOnce/Program.cs
```diff
@@ -15,788 +15,988 @@
...
+                .AddJsonFile("./appsettings.json")
+                .Build();
+
+            var adminClientConfig = configuration.GetSection("AdminClient").Get<AdminClientConfig>();
+            using var adminClient = new AdminClientBuilder(adminClientConfig).Build();
+
+            switch (command)
+            {
+                case "create_words":
+                    await new WordCreator().Execute(configuration, transactionalId, adminClient, cts);
```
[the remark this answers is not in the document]
emasab: Ok, I needed it before when it was using an interface instead of delegates

examples/ExactlyOnce/Program.cs
```diff
@@ -15,788 +15,988 @@
...
+    /// </summary>
+    public class TransactionalProcessorFatalError : Exception
+    {
+    }
+
+    /// <summary>
+    ///     Represents a single message to be sent
+    ///     to an output topic.
+    /// </summary>
+    public class ProduceMessage<K, V>
```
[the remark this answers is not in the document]
emasab: I think it's needed, imagine you want to do transactions involving multiple input and multiple output topics. It's common enough.
About the delivery handler, if you have a producer only and want to do transactions involving messages sent from multiple http calls, let's says many users belonging to the same group that have to be sent all together or none, you may want to give them a response when complete. You could send everyone the same response when EndTransaction is called, but can be useful to have the offset, partition or error code for logging. Also it's optional so can be used only if needed

examples/ExactlyOnce/Program.cs
emasab: the consumer can be put as a local variable, but the producer needs to be recreated and the instance should be passed back and forth to CommitTransaction and it cannot be returned in case it throws

examples/ExactlyOnce/Program.cs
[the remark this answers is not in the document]
emasab: But what happens if the transaction is aborted? I'm thinking that it could be changed to use ProduceAsync, get the delivery reports, await them at the end of transaction and pass them to EndTransaction to avoid having the caller wait for EndTransaction to actually deliver those reports.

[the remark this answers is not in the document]
mhowlett: > I think it's needed, imagine you want to do transactions involving multiple input and multiple output topics. It's common enough.

yes. i thought i'd thought of a better way to do it (using the return mechanism), but i can't think of what i might have been thinking now. A significant issue with what you propose is you don't want to be specifying `Topic = config.OutputTopic.Name,` - kind of defeats the purpose of having OutputTopic.

perhaps we don't want to return the result after all, perhaps we want a `Produce` method for this. but rather than exposing the producer instance to the user and have the user call `Produce` on that, we keep that encapsulated, and put a Produce method on the `Processor` class.

I don't quite see right now how the delivery reports work. Maybe they aren't exposed by the new `Produce` method, but maybe they are made available at transaction end. Since delivery reports contain the message instance, and all the meta info, everything is there to satisfy the use case you point out i think.

Feel free to iterate on the above idea and see if it works - just an idea. I'm convinced that having `ProduceMessage` is going to lead to a worse structure than optimal.

Issue #1747, opened 2022-01-13
Example of Assign() is slightly misleading


skalinets:
Description
===========

While consumer group is not used in Assign() scenarios, it still needs to be provided for whatever reason. In the [samples](https://github.com/confluentinc/confluent-kafka-dotnet/blob/4de84f7c3c4b0d140f643dcb21110fd523efa2cc/examples/Consumer/Program.cs#L163) `new Guid()` is used for naming that group. It seems to be misleading, because in most scenarios Guids are used as unique identifiers, but in this case it is just a constant (`00000000-0000-0000-0000-000000000000`). 

As a result developers often get incorrect assumption that consumer group names for Assign() should be unique (even with each service restart). They just blindly miss that Guid is created via `new Guid()` but not `Guid.NewGuid()`.  And it ends up with hundreds of empty consumer groups with Guid names in clusters. 

I seems to be better to specify some hardcoded constant (and probably explicitly state that that value could be used for ALL services if needed).  

How to reproduce
================

- Change [this line](https://github.com/confluentinc/confluent-kafka-dotnet/blob/4de84f7c3c4b0d140f643dcb21110fd523efa2cc/examples/Consumer/Program.cs#L163) to `GroupId = Guid.NewGuid().ToString(),`
- Run sample few times
- New consumer group with unique name is created each time. Old groups are not removed.

bartelink:
> `new Guid()`
Perhaps `Guid.Empty` might be a better way to represent the fact that it's a constant?

Though I'd favor `"TestConsumerGroupName"` or something more explicit

mhowlett:
yeah, that's probably better. we could probably improve some of the comment text as well.

mhowlett:
https://github.com/confluentinc/confluent-kafka-dotnet/pull/1914

except I didn't do:
> seems to be better to specify some hardcoded constant
(should)

also, there are some comments that don't line up with the code in this example.

also we should demonstrate store offset, as this is best practice.

skalinets:
> #1914
> new Guid() gives 00000000-0000-0000-0000-000000000000, Guid.NewGuid() gives a new unique guid, which is the intent.

I think the intent is the opposite. `Guid.NewGuid()` creates a new consumer group during every service restart, cluttering the cluster. 
These groups are not used in assign scenario. Only one group with hardened id should be enough for all consumers.

anchitj:
Closing the issue as the example has been fixed now.

PR #1918, opened 2022-10-13
Fix consumer example

anchitj:
For #1747.

Changes done:
1. For `Run_Consume` updated to use `StoreOffset` instead of `Commit`.
2. For `Run_ManualAssign` updated `GroupId` to a static id so that a new group isn't created on every restart.

### DIFF
--- a/examples/Consumer/Program.cs
+++ b/examples/Consumer/Program.cs
@@ -41,6 +41,7 @@ namespace Confluent.Kafka.Examples.ConsumerExample
                 BootstrapServers = brokerList,
                 GroupId = "csharp-consumer",
                 EnableAutoOffsetStore = false,
+                EnableAutoCommit = true,
                 StatisticsIntervalMs = 5000,
                 SessionTimeoutMs = 6000,
                 AutoOffsetReset = AutoOffsetReset.Earliest,
@@ -115,7 +116,10 @@ namespace Confluent.Kafka.Examples.ConsumerExample
                             Console.WriteLine($"Received message at {consumeResult.TopicPartitionOffset}: {consumeResult.Message.Value}");
                             try
                             {
-                                // StoreOffset stores offset in an in-memory store which will be automatically committed every AutoCommitIntervalMs.
+                                // Store the offset associated with consumeResult to a local cache. Stored offsets are committed to Kafka by a background thread every AutoCommitIntervalMs. 
+                                // The offset stored is actually the offset of the consumeResult + 1 since by convention, committed offsets specify the next message to consume. 
+                                // If EnableAutoOffsetStore had been set to the default value true, the .NET client would automatically store offsets immediately prior to delivering messages to the application. 
+                                // Explicitly storing offsets after processing gives at-least once semantics, the default behavior does not.
                                 consumer.StoreOffset(consumeResult);
                             }
                             catch (KafkaException e)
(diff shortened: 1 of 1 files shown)


examples/Consumer/Program.cs
```diff
@@ -40,7 +40,7 @@ public static void Run_Consume(string brokerList, List<string> topics, Cancellat
             {
                 BootstrapServers = brokerList,
                 GroupId = "csharp-consumer",
-                EnableAutoCommit = false,
+                EnableAutoOffsetStore = false,
```
anchitj, line 43: @mhowlett Referred https://github.com/edenhill/librdkafka/blob/master/INTRODUCTION.md#at-least-once-processing which mentions this to be false and AutoCommit to be enabled for `StoreOffset` to work. 

Please let me know if this is not right.

mhowlett: this is right. i'd probably also explicitly configure `EnableAutoCommit` to `true` for clarity, even though it's the default.

examples/Consumer/Program.cs
```diff
@@ -115,23 +113,14 @@ public static void Run_Consume(string brokerList, List<string> topics, Cancellat
                             }
 
                             Console.WriteLine($"Received message at {consumeResult.TopicPartitionOffset}: {consumeResult.Message.Value}");
-
-                            if (consumeResult.Offset % commitPeriod == 0)
+                            try
+                            {
+                                // StoreOffset stores offset in an in-memory store which will be automatically committed every AutoCommitIntervalMs.
```
mhowlett, line 118: maybe:

"Store the offset associated with `consumeResult` to a local cache. Stored offsets are committed to Kafka by a background thread every `AutoCommitIntervalMs`. The offset stored is actually the offset of the consumeResult + 1 since by convention, committed offsets specify the next message to consume. If `EnableAutoOffsetStore` had been set to the default value `true`, the .NET client would automatically store offsets immediately prior to delivering messages to the application. Explicitly storing offsets after processing gives at-least once semantics, the default behavior does not."

mhowlett: we really should also document the Erroneous state gotcha here...

examples/Consumer/Program.cs
```diff
@@ -160,12 +149,12 @@ public static void Run_ManualAssign(string brokerList, List<string> topics, Canc
             {
                 // the group.id property must be specified when creating a consumer, even 
                 // if you do not intend to use any consumer group functionality.
-                GroupId = Guid.NewGuid().ToString(),
+                GroupId = "csharp-consumer-GROUPID-NOT-USED",
```
mhowlett, line 152: maybe: "groupid-not-used-but-mandatory" ?

mhowlett (approved):
lgtm with suggested changes.

examples/Consumer/Program.cs
anchitj: Thanks for confirming. Also enabled `EnableAutoCommit` in the config.

examples/Consumer/Program.cs
anchitj: Updated.

examples/Consumer/Program.cs
anchitj: Thanks, Updated the comment.

anchitj: @mhowlett For Erroneous state above comment already specifies this will ensure at least once consumption. Did you mean to document something other than that as well here?

anchitj:
@mhowlett Hi Matt, I have made the changes. Can you please take a look and confirm if we are good to merge this?

PR #1939, opened 2022-11-16
Add default rebalance callback

anchitj:
Added default rebalance callback handler to prevent consumers not having partition revoked handlers from entering into the erroneous state.

Consumers might enter into erroneous state as they might have lost the partitions during rebalance but still they are committing the offsets for them as they aren't aware of the changes.

Some changes were needed to the tests to make sure the rebalance is running properly.

### Testing

Verified all the integration tests are working.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 
 ## Fixes
 
-- Added default rebalance callback handler to prevent consumers not having partition revoked handlers from entering into the erroneous state.
+- During a group rebalance, partitions are now always revoked as a side effect of a call to Consume, whether or not a partitions revoked handler has been specified. Previously, if no handler was specified, the timing of when the consumer lost ownership of partitions during a rebalance was arbitrarily, frequently resulting in an erroneous state exception when committing or storing offsets.
 
 
 # 1.9.4
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@
 - Upgraded `NJsonSchema` to v10.6.3
 - Added `LatestCompatibilityStrict` configuration property to JsonSerializerConfig to check the compatibility with latest schema
   when `UseLatestVersion` is set to true.
+- Added DeleteConsumerGroupOffset to AdminClient.
+
+## Fixes
+
+- Added default rebalance callback handler to prevent consumers not having partition revoked handlers from entering into the erroneous state.
 
 
 # 1.9.4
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+# 2.0.0 
+
+## Enhancements
+
+- Added DeleteConsumerGroupOffset to AdminClient.
```
anchitj, line 5: This is for https://github.com/confluentinc/confluent-kafka-dotnet/commit/fa7626c49fc5955a82db276d199b73361a739e16 where I missed updating the changelog.

src/Confluent.Kafka/Consumer.cs
```diff
@@ -691,10 +691,9 @@ internal Consumer(ConsumerBuilder<TKey, TValue> builder)
 
             IntPtr configPtr = configHandle.DangerousGetHandle();
 
-            if (partitionsAssignedHandler != null || partitionsRevokedHandler != null || partitionsLostHandler != null)
-            {
-                Librdkafka.conf_set_rebalance_cb(configPtr, rebalanceDelegate);
-            }
+            // Set the rebalance callback always to prevent the consumer from entering in erroneous state trap.
```
mhowlett, line 694: don't need this comment

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 - Upgraded `NJsonSchema` to v10.6.3
 - Added `LatestCompatibilityStrict` configuration property to JsonSerializerConfig to check the compatibility with latest schema
   when `UseLatestVersion` is set to true.
+- Added DeleteConsumerGroupOffset to AdminClient.
+
+## Fixes
+
+- Added default rebalance callback handler to prevent consumers not having partition revoked handlers from entering into the erroneous state.
```
mhowlett, line 12: maybe:

```
- During a group rebalance, partitions are now always revoked as a side effect of a call to Consume, whether or not a partitions revoked handler has been specified. Previously, if no handler was specified, the timing of when the consumer lost ownership of partitions during a rebalance was arbitrarily, frequently resulting in an erroneous state exception when committing or storing offsets.
```

mhowlett (approved):
lgtm after comments addressed.

src/Confluent.Kafka/Consumer.cs
anchitj: Updated

CHANGELOG.md
anchitj: Updated

PR #1947, opened 2022-11-24
Add OAuth examples

anchitj:
Added examples for using OAuth bearer SASL in Producer and Consumer. Taken reference from https://github.com/confluentinc/confluent-kafka-go/pull/836/files.

Tested all of them with the clusters setup through trivup.


examples/OAuthConsumer/Program.cs
```diff
@@ -0,0 +1,192 @@
...
+        public string TokenValue { get; set; }
+        public long Expiration { get; set; }
+        public String Principal { get; set; }
+        public Dictionary<String, String> Extensions { get; set; }
+    }
+
+    public class Program
+    {
+        private const String OauthConfigRegexPattern = "^(\\s*(\\w+)\\s*=\\s*(\\w+))+\\s*$"; // 1 or more name=value pairs with optional ignored whitespace
+        private const String OauthconfigKeyValueRegexPattern = "(\\w+)\\s*=\\s*(\\w+)"; // Extract key=value pairs from OAuth Config
```
edenhill, line 45: Inconsistent casing, should be OauthConfigKey...

examples/OAuthConsumer/Program.cs
```diff
@@ -0,0 +1,192 @@
...
+            }
+
+            var consumerConfig = new ConsumerConfig
+            {
+                BootstrapServers = bootstrapServers,
+                SecurityProtocol = SecurityProtocol.SaslPlaintext,
+                SaslMechanism = SaslMechanism.OAuthBearer,
+                SaslOauthbearerConfig = oauthConf,
+                GroupId = groupId,
+                SessionTimeoutMs = 6000,
```
edenhill, line 76: no need for this

examples/OAuthConsumer/Program.cs
```diff
@@ -0,0 +1,192 @@
...
+
+            var parsedConfig = new Dictionary<String, String>();
+            foreach (Match match in Regex.Matches(oauthConfig, OauthconfigKeyValueRegexPattern))
+            {
+                parsedConfig[match.Groups[1].ToString()] = match.Groups[2].ToString();
+            }
+
+            if (!parsedConfig.ContainsKey(PrincipalKey) || !parsedConfig.ContainsKey(ScopeKey) || parsedConfig.Count > 2)
+            {
+                Console.WriteLine($"Invalid OAuth config {oauthConfig} passed. Ignoring the token refresh");
```
edenhill, line 155: The proper way to handle this is to raise an error which is then passed to SetTokenFailure() in the refresh delegate.

examples/OAuthProducer/Program.cs
```diff
@@ -0,0 +1,172 @@
...
+            return new OAuthBearerToken
+            {
+                TokenValue = $"{ToUnpaddedBase64(headerJson)}.{ToUnpaddedBase64(payloadJson)}.",
+                Expiration = expiresAt.ToUnixTimeMilliseconds(),
+                Principal = principal,
+                Extensions = new Dictionary<string, string>()
+            };
+        }
+    }
+}
```
edenhill, line 172: missing newline at end of file

examples/OAuthProducer/Program.cs
```diff
@@ -0,0 +1,172 @@
+﻿// Copyright 2022 Confluent Inc.
```
edenhill, line 1: Do we really need this example, it looks identical to the consumer one?

edenhill (changes requested):
Great stuff!

examples/OAuthProducer/Program.cs
anchitj: I was following the go examples, but I guess I can merge both of them. Let me update this.

edenhill: Fair enough, I think it is better to have duplciate examples in that case, rather than having one example be both producer and consumer.

examples/OAuthConsumer/Program.cs
anchitj: Thanks, updated.

examples/OAuthConsumer/Program.cs
anchitj: Makes sense, throwing an exception now and setting `OAuthBearerSetTokenFailure` on catching the exception.

examples/OAuthProducer/Program.cs
anchitj: Updated.

examples/OAuthConsumer/Program.cs
```diff
@@ -0,0 +1,197 @@
...
+            // It must be invoked whenever the client requires a token (i.e. when it first starts and when the
+            // previously-received token is 80% of the way to its expiration time).
+            void OauthCallback(IClient client, string cfg)
+            {
+                try
+                {
+                    var token = retrieveUnsecuredToken(cfg);
+                    client.OAuthBearerSetToken(token.TokenValue, token.Expiration, token.Principal);
+                }
+                catch (Exception e)
```
edenhill, line 91: Neat :+1:

edenhill (approved):
LGTM

PR #1949, opened 2022-11-28
Fix 100% CPU usage with DependentAdminClient

### LINKED ISSUE #1762: Consumer high CPU utilization while waiting for new messages
I have a .NET WebAPI application which acts like a REST-wrapper around Kafka Consumer. Clients can create, assign, delete consumers and consume messages via REST interface.
The issue is that the application consumes a lot of CPU resources just waiting for new data. On my computer with 8 cores one consumer instance takes about 17% CPU. The same behaviour reproduces on all machines there the application runs.

The application creates a consumer, manually assigns partitions, and then executes "Consumer.Consume(timeout)" on each client request with timeout of 15 seconds. And while executing "Consumer.Consume(timeout)" CPU utilisation rises up to 100% in case of multiple active consumer instances even if there are no new incoming messages.

The strange thing is that this behaviour doesn't reproduce on "example" scenarios. I wrote simple console app that just creates a consumer with the same config, assign partitions and wait for new messages in loop - it works fine and doesn't take any CPU resources. The only difference that I can see is that in my web app consumers are created and accessed in different threads, but access is strictly sequential.

Could you please provide me with possible reasons of this behaviour?

Confluent.Kafka 1.8.2 (tried 1.7.0)
Apache Kafka 3.1.0 (tried 2.8.1)
Client configuration:
    bootstrap.servers: localhost:9092
    security.protocol: PLAINTEXT
    group.id: kafkatest3@silent
    auto.offset.reset: earliest
    enable.partition.eof: false
Operating system: Windows 10

Debug logs (consumer,cgrp,topic,fetch) while executing "Consumer.Consume(timeout)":
```
%7|1645037196.207|FETCH|rdkafka#consumer-1| [thrd:localhost:9092/bootstrap]: localhost:9092/1: Fetch topic test5 [2] at offset 0 (v3)
%7|1645037196.207|FETCH|rdkafka#consumer-1| [thrd:localhost:9092/bootstrap]: localhost:9092/1: Fetch topic test5 [1] at offset 0 (v3)
%7|1645037196.207|FETCH|rdkafka#consumer-1| [thrd:localhost:9092/bootstrap]: localhost:9092/1: Fetch topic test5 [0] at offset 0 (v3)
%7|1645037196.207|FETCH|rdkafka#consumer-1| [thrd:localhost:9092/bootstrap]: localhost:9092/1: Fetch 3/3/3 toppar(s)
%7|1645037196.707|FETCH|rdkafka#consumer-1| [thrd:localhost:9092/bootstrap]: localhost:9092/1: Fetch topic test5 [1] at offset 0 (v3)
%7|1645037196.707|FETCH|rdkafka#consumer-1| [thrd:localhost:9092/bootstrap]: localhost:9092/1: Fetch topic test5 [0] at offset 0 (v3)
%7|1645037196.707|FETCH|rdkafka#consumer-1| [thrd:localhost:9092/bootstrap]: localhost:9092/1: Fetch topic test5 [2] at offset 0 (v3)
%7|1645037196.707|FETCH|rdkafka#consumer-1| [thrd:localhost:9092/bootstrap]: localhost:9092/1: Fetch 3/3/3 toppar(s)
... (other strings are the same)
```

anchitj:
Fixes #1762 

Use `ExtractCancellationDelayMaxMs` to extract value of `cancellationDelayMaxMs` when using DependentAdminClient.

Currently when using `DependentAdminClient` cancellationDelayMaxMs is set to 0 therefore librdKafka continuously keeps on polling which pushes the CPU usage to 100%.

### Testing

Verified using the samples given the issue above #1762 that CPU usage is not increasing.

Also ran the current integration tests to confirm no regression.

### DIFF
--- a/src/Confluent.Kafka/AdminClient.cs
+++ b/src/Confluent.Kafka/AdminClient.cs
@@ -769,7 +769,7 @@ namespace Confluent.Kafka
         /// </param>
         internal AdminClient(Handle handle)
         {
-            this.cancellationDelayMaxMs = 100;                          
+            Config.ExtractCancellationDelayMaxMs(new AdminClientConfig(), out this.cancellationDelayMaxMs);                          
             this.ownedClient = null;
             this.handle = handle;
             Init();


src/Confluent.Kafka/AdminClient.cs
```diff
@@ -768,7 +768,8 @@ private SafeKafkaHandle kafkaHandle
         ///     or AdminClient handle.
         /// </param>
         internal AdminClient(Handle handle)
-        {                            
+        {
+            this.cancellationDelayMaxMs = 100;
```
edenhill, line 772: I believe it should use `ExtractCancellationDelayMaxMs()` so that this value can be configured (and that we reuse the default value).

anchitj: `ExtractCancellationDelayMaxMs` needs Config to be passed on which isn't available with DependentAdminClient as DependentAdminClient is created through librdkafka handle corresponding to the producer/consumer.

anchitj: Or do you mean I should pass an empty config here?

anchitj: Updated to use `ExtractCancellationDelayMaxMs` with an empty config.

edenhill (approved):
LgTM

Issue #1955, opened 2022-12-04
Custom Default Partitioner not being invoked when producing a message


victormartinsantiago:
Description
===========
We want to use a custom default partitioner for our producer. 
We set it by calling `SetDefaultPartitioner` in the producer builder, but the delegate is never invoked when producing the message.

As per your [example](https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/test/Confluent.Kafka.IntegrationTests/Tests/Producer_CustomPartitioner.cs)

- Nugget Version: 1.9.3
- Apache Kafka Version: Confluent Cloud
- Client Config:   SecurityProtocol = SecurityProtocol.SaslSsl, SaslMechanism = SaslMechanism.Plain,
- Operating System: Windows

How to reproduce
================

Simple producer implementation for the test

```
var producerBuilder = new ProducerBuilder<Null, string>(config);
producerBuilder.SetDefaultPartitioner((string topicName, int partitionCount, ReadOnlySpan<byte> keyData, bool keyIsNull) =>
{
    return 0;
})

using var producer  = producerBuilder.Build();

await producer.ProduceAsync(topic, new Message<Null, string> { Value = $"Message {Guid.NewGuid()}" });
```

We expect that after calling `await producer.ProduceAsync` the delegate is invoked before producing the message but it never happens.

Checklist
=========

Please provide the following information:
 - [x] A complete (i.e. we can run it), minimal program demonstrating the problem. No need to supply a project file.
 - [x] Confluent.Kafka nuget version.
 - [x] Apache Kafka version.
 - [x] Client configuration.
 - [x] Operating system.
 - [ ] Provide logs (with "debug" : "..." as necessary in configuration).
 - [ ] Provide broker log excerpts.
 - [ ] Critical issue.

anchitj:
Thanks for the report. I am able to reproduce it with `Null` keys. Investigating further.

victormartinsantiago:
Thank you @anchitj. I've also tested the topic-specific partitioner `SetPartitioner` and I get the same behaviour.

```
var topic = "my.topic";
var producerBuilder = new ProducerBuilder<Null, string>(config);
producerBuilder.SetPartitioner(topic, (string topicName, int partitionCount, ReadOnlySpan<byte> keyData, bool keyIsNull) =>
{
    return 0;
});
```

Cheers

vpa1977:
This is behaviour provided by librdkafka. 

[...]

PR #1965, opened 2022-12-15
Disable sticky partitioning for null key producer in CustomPartitionerTest

anchitj:
### Description

From #1955. 

Updated test for null key producer so that assertions inside the partitioner are actually called.

### Testing

Verified through logs that the partitioner is getting called.

### DIFF
--- a/test/Confluent.Kafka.IntegrationTests/Tests/Producer_CustomPartitioner.cs
+++ b/test/Confluent.Kafka.IntegrationTests/Tests/Producer_CustomPartitioner.cs
@@ -117,6 +117,11 @@ namespace Confluent.Kafka.IntegrationTests
 
 
             // Null key
+            producerConfig = new ProducerConfig
+            {
+                BootstrapServers = bootstrapServers,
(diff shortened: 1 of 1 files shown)


test/Confluent.Kafka.IntegrationTests/Tests/Producer_CustomPartitioner.cs
```diff
@@ -117,6 +117,11 @@ public void Producer_CustomPartitioner(string bootstrapServers)
 
 
             // Null key
+            producerConfig = new ProducerConfig
+            {
+                BootstrapServers = bootstrapServers,
+                StickyPartitioningLingerMs = 0 // disable sticky partitioning and run the default partitioner for null keys.
+            };
```
edenhill, line 124: The original test should've failed by also asserting that the partitioner was called, so I recommend adding that too, i.e., a counter in the partitioner that is checked after Flush to be equal to the message count (1).

anchitj: Thanks, updated to include this.

milindl (approved):

PR #1969, opened 2022-12-21
Add service.yml for adding project to the semaphore CI

anchitj:
Taken reference from internal wikis and existing binding files
* https://github.com/confluentinc/confluent-kafka-go/blob/master/service.yml
* https://github.com/confluentinc/confluent-kafka-python/blob/master/service.yml

Will create the semaphore project once this is merged and test the CI from there.

### DIFF
--- a/service.yml
+++ b/service.yml
@@ -1,0 +1,11 @@
+name: confluent-kafka-dotnet
+lang: unknown
+lang_version: unknown
+git:
+  enable: true
+github:
+  enable: true
+  repo_name: confluentinc/confluent-kafka-dotnet
+semaphore:
+  enable: true
+  pipeline_enable: false


service.yml
```diff
@@ -0,0 +1,11 @@
+name: confluent-kafka-dotnet
+lang: unknown
+lang_version: unknown
```
anchitj, line 3: This is unkown as both of these tags don't support C#.

edenhill (commented):
Let's add some build and test steps too. Also consider adding documentation generation, so we know that isn't breaking in commits.

anchitj:
> Let's add some build and test steps too. Also consider adding documentation generation, so we know that isn't breaking in commits.

I think we need to add those to the semaphore.yml which will be created once the current PR (service.yml) is merged and the run-service-bot job runs.

milindl (approved):

PR #1898, opened 2022-09-16
[admin] Add consumer group bindings for KIP-88, 222, 518, 396 (partial)

milindl:
Currently done: List/Alter Consumer Group Offsets; List/Describe Consumer Groups

PR #1981, opened 2023-01-18
 [admin] Add consumer group bindings for KIP-88, 222, 518, 396 (partial) 

milindl:
Creating this to replace #1898 as to work with CI which uses internal branches.


emasab (approved):
Great PR @milindl !

PR #2021, opened 2023-03-23
[KIP-430] DescribeConsumerGroups, DescribeTopics, DescribeCluster with authorized AclOperations

jainruchir:


src/Confluent.Kafka/Admin/DescribeTopicsOptions.cs
```diff
@@ -0,0 +1,45 @@
...
+// Refer to LICENSE for more information.
+
+using System;
+using System.Collections.Generic;
+
+
+namespace Confluent.Kafka.Admin
+{
+    /// <summary>
+    ///     Options for the "AdminClient.DescribeTopicsOptions" method.
```
milindl, line 24:
```suggestion
    ///     Options for the "AdminClient.DescribeTopics" method.
```

src/Confluent.Kafka/Impl/LibRdKafka.cs
```diff
@@ -418,6 +425,7 @@ static bool SetDelegates(Type nativeMethodsClass)
             _Node_id = (_Node_id_delegate)methods.Single(m => m.Name == "rd_kafka_Node_id").CreateDelegate(typeof (_Node_id_delegate));
             _Node_host = (_Node_host_delegate)methods.Single(m => m.Name == "rd_kafka_Node_host").CreateDelegate(typeof (_Node_host_delegate));
             _Node_port = (_Node_port_delegate)methods.Single(m => m.Name == "rd_kafka_Node_port").CreateDelegate(typeof (_Node_port_delegate));
+            _Node_rack = (_Node_rack_delegate)methods.Single(m => m.Name == "rd_kafka_Node_rack_id").CreateDelegate(typeof (_Node_rack_delegate));
```
milindl, line 428: Note to change this as per latest public api

test/Confluent.Kafka.IntegrationTests/AdminClient_DescribeTopics.cs
```diff
@@ -0,0 +1,189 @@
...
+                        TopicCollection.OfTopicNames(topicList),
+                        describeOptionsWithTimeout);
+                }
+                catch (DescribeTopicsException ex)
+                {
+                    var resCount = ex.Results.TopicDescriptions.Count;
+                    Assert.Equal(2, resCount);
+                    
+                    // TODO: remove after fix
+                    ex.Results.TopicDescriptions.Sort(
```
milindl, line 82: Note: can remove now

test/Confluent.Kafka.IntegrationTests/AdminClient_DescribeTopics.cs
```diff
@@ -0,0 +1,189 @@
...
+                        (a,b) =>
+                        {
+                            return topicList.IndexOf(a.Name) -
+                                topicList.IndexOf(b.Name);
+                        }
+                    );
+                    Assert.Empty(ex.Results.TopicDescriptions[0].AuthorizedOperations);
+                    Assert.Empty(ex.Results.TopicDescriptions[1].AuthorizedOperations);
+                    Assert.True(ex.Results.TopicDescriptions[0].Error.IsError);
+                    Assert.True(!ex.Results.TopicDescriptions[1].Error.IsError);
```
milindl, line 92: nit: use Assert.False instead.

test/Confluent.Kafka.IntegrationTests/AdminClient_DescribeTopics.cs
```diff
@@ -0,0 +1,189 @@
...
+                    BootstrapServers = bootstrapServers
+                }).Build())
+                {
+                     descResWithAuthOps = await adminClientUser.DescribeTopicsAsync(topicListAuthOps,
+                        describeOptionsWithAuthOps);
+                }
+
+                var descResAuthOps =
+                    descResWithAuthOps.TopicDescriptions[0].AuthorizedOperations;
+                Assert.Equal(
```
milindl, line 165: I think we should OrderBy before asserting equality, because while the order of the operations will be exactly what you described, there is no reason for it to be so (and it's not something we or the user should assume)
Here and elsewhere

test/Confluent.Kafka.IntegrationTests/AdminClient_DescribeTopics.cs
```diff
@@ -0,0 +1,189 @@
...
+                    
+                    // TODO: remove after fix
+                    ex.Results.TopicDescriptions.Sort(
+                        (a,b) =>
+                        {
+                            return topicList.IndexOf(a.Name) -
+                                topicList.IndexOf(b.Name);
+                        }
+                    );
+                    Assert.Empty(ex.Results.TopicDescriptions[0].AuthorizedOperations);
```
milindl, line 89: Note to change this to asserting null after making change to allow for null authorized operations
here and in other files

test/Confluent.Kafka.IntegrationTests/AdminClient_DescribeTopics.cs
```diff
@@ -0,0 +1,189 @@
...
+                    {
+                        Pattern = new ResourcePattern
+                        {
+                            Type = ResourceType.Topic,
+                            Name = singlePartitionTopic,
+                            ResourcePatternType = ResourcePatternType.Literal
+                        },
+                        Entry = new AccessControlEntry
+                        {
+                            Principal = "User:user",
```
milindl, line 120: Use 
"User:(value of string user)"
rather than hard-coding it.
For all the relevant tests.

milindl (commented):
Seems mostly okay, some minor changes and comments

src/Confluent.Kafka/IAdminClient.cs
```diff
@@ -531,4 +531,77 @@ public interface IAdminClient : IClient
         Task AlterUserScramCredentialsAsync(IEnumerable<UserScramCredentialAlteration> alterations, AlterUserScramCredentialsOptions options = null);
     }
 
+    /// <summary>
+    ///     Extension methods for default <see cref="IAdminClient"/> implementations.
+    /// </summary>
+    public static class IAdminClientExtensions
```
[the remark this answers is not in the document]
milindl: I see we are doing this to avoid breaking binary compatibility but we don't guarantee that for .NET, right? 
Also, as per https://learn.microsoft.com/en-us/dotnet/core/compatibility/categories#binary-compatibility
If we just add methods or method implementations, that doesn't break binary compatibility, too, right?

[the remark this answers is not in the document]
emasab: Yeah, it's extending an interface that breaks it, we're starting to avoid breaking it.

test/Confluent.Kafka.IntegrationTests/AdminClient_DescribeTopics.cs
emasab: The order will be that one because it comes from the order in the enum that corresponds to the bitfield order too

milindl (approved):
Thanks for looking into this @emasab !
Approved with a few comments here and there.

PR #2037, opened 2023-04-07
add trace event

chad122:


bjornbouetsmith:
I think this is a bad idea to merge.
Every TraceEvent call will allocate a string on the heap - even if people have turned off tracing.
If you really want to introduce tracing, which can be a good idea - you should try to make it non-allocating.
Libraries should not put pressure on the GC - if this gets merged into the master branch every single produce will allocate several string.

So you should probably instead of just calling .ToString() - instead make a static ctor that creates a dictionary that you can do a lookup into to get a preallocated string value.

chad122:
> I think this is a bad idea to merge. Every TraceEvent call will allocate a string on the heap - even if people have turned off tracing. If you really want to introduce tracing, which can be a good idea - you should try to make it non-allocating. Libraries should not put pressure on the GC - if this gets merged into the master branch every single produce will allocate several string.
> 
> So you should probably instead of just calling .ToString() - instead make a static ctor that creates a dictionary that you can do a lookup into to get a preallocated string value.

Good suggestion. I'll make some change.

Issue #1743, opened 2021-12-29
ClientConfig.SecurityProtocol not compatible with official config values.


randomravings:
When instantiating a client config, eg:
```csharp
var config = new ProducerConfig(IDictionary<string, string> config)
```
using the string configs from the Apache Kafka documentation for [security.protocol](https://kafka.apache.org/documentation/#producerconfigs_security.protocol) then the ClientConfig throws and ArgumentException because it cannot convert it to a value in the [ClientConfig.SecurityProtocol](https://github.com/confluentinc/confluent-kafka-dotnet/blob/df4fb025d3dfab26179d3e9d4aa6cf954f0acbc5/src/Confluent.Kafka/Config_gen.cs#L103).

This is the line that breaks:
https://github.com/confluentinc/confluent-kafka-dotnet/blob/df4fb025d3dfab26179d3e9d4aa6cf954f0acbc5/src/Confluent.Kafka/Config_gen.cs#L623

As an example, if I were to use the following config from a file:
```
{
  "bootstrap.servers": "localhost:9092",
  "group.id": "some-group-id",
  "auto.offset.reset": "earliest",
  "enable.auto.commit": "true",
  "auto.commit.interval.ms": 5000,
  "max.poll.interval.ms": 300000,
  "security.protocol": "SASL_SSL",
  "sasl.mechanism": "OAUTHBEARER"
}
```

For "sasl.mechanism" it is solved in the [ClientConfig.SaslMechanism](https://github.com/confluentinc/confluent-kafka-dotnet/blob/df4fb025d3dfab26179d3e9d4aa6cf954f0acbc5/src/Confluent.Kafka/Config_gen.cs#L290)

I would suggest a similar approach be taken wrt. ClientConfig.SecurityProtocol.

edenhill:
I think @randomravings is right, there should be strictly typed enums in the dotnet client for things like SecurityProtocol.

But Matt is also right, everyone is right!, librdkafka should be more forgiving.
I'll fix the latter.

edenhill:
Scratch that, librdkafka's enums are already case insensitive.

mhowlett:
right, looks like an issue on the .NET side.

huaxing-yuan:
I encountered the same error following get started document: 
https://developer.confluent.io/get-started/dotnet/#configuration

If in producer.properties, we specify SASL_SSL like in the documentation, SecurityProtocol throws argument exception.
[...]

PR #2045, opened 2023-05-02
Fix: #1743 ClientConfig.SecurityProtocol not compatible with official config values.

huaxing-yuan:
This PR tries to fix issue #1743 `ClientConfig.SecurityProtocol not compatible with official config values.`

By following getting started document, https://developer.confluent.io/get-started/dotnet/#configuration, We want to set `security.protocol=SASL_SSL`

But when creating ProducerConfig (inherits from ClientConfig) with given `producer.properties` file, ClientConfig.SecurityProtocol throws exception as described in #1743.
This PR convert value in lowercase before searching in `EnumNameToConfigValueSubstitutes` (same transformation as implemented in `SetObject` method) , so official config values would be supported correctly.

### DIFF
--- a/src/Confluent.Kafka/Config.cs
+++ b/src/Confluent.Kafka/Config.cs
@@ -160,7 +160,8 @@ namespace Confluent.Kafka
         {
             var result = Get(key);
             if (result == null) { return null; }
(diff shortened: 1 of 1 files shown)


danielrosatellez (approved):
Looks correct.

Issue #2060, opened 2023-06-02
Unintentional Breaking Change to TopicPartitionOffset in 2.1


drinehim:
Description
===========
Version 2.1 (specifically commit 9826d794) introduced a breaking change (looks unintentional as you would think optional parameters would be non-breaking but they aren't). This is causing my applications that reference a library that uses a library that uses Confluent.Kafka to break when the entry app tries to upgrade to the latest Confluent.Kafka.

Library -> Confluent.Kafka 2.0.2
App -> Confluent.Kafka 2.1.1

Exception:
RLI.EventMessaging.Client.ClientHandler[0]
          Method not found: 'Void Confluent.Kafka.TopicPartitionOffset..ctor(System.String, Confluent.Kafka.Partition, Confluent.Kafka.Offset)'.
          System.MissingMethodException: Method not found: 'Void Confluent.Kafka.TopicPartitionOffset..ctor(System.String, Confluent.Kafka.Partition, Confluent.Kafka.Offset)'.
             at Confluent.Kafka.Consumer`2.Consume(Int32 millisecondsTimeout)
             at RLI.EventMessaging.Kafka.Client.KafkaClient`3.<>c__DisplayClass31_0`1.<ConsumeWithRetryAsync>b__0() in C:\code\RLI.EventMessaging\src\RLI.EventMessaging.Kafka\Client\KafkaClient.cs:line 421
             at System.Threading.Tasks.Task`1.InnerInvoke()

How to reproduce
================
[KafkaBreakTest.zip](https://github.com/confluentinc/confluent-kafka-dotnet/files/11638112/KafkaBreakTest.zip)

Library -> Confluent.Kafka 2.0.2
App -> Confluent.Kafka 2.1.1

Kafka Version: Any
Client Configuration: Basic
OS: Linux (probably windows too)
Critical: No
Broker Logs: Can't get that far

Checklist
=========

Please provide the following information:
 - [X] A complete (i.e. we can run it), minimal program demonstrating the problem. No need to supply a project file.
 - [X] Confluent.Kafka nuget version.
 - [X] Apache Kafka version.
 - [X] Client configuration.
 - [X] Operating system.
 - [X] Provide logs (with "debug" : "..." as necessary i.
 - [X] Provide broker log excerpts.
 - [X] Critical issue.
[KafkaBreakTest.zip](https://github.com/confluentinc/confluent-kafka-dotnet/files/11638112/KafkaBreakTest.zip)

[...]

PR #2061, opened 2023-06-02
Fix backwards compatability of TopicPartitionOffset constructor #2060

drinehimer:
Version 2.1 (specifically commit https://github.com/confluentinc/confluent-kafka-dotnet/commit/9826d7941176882ac0d739a08912daa2054b3c95) introduced a breaking change (looks unintentional as you would think optional parameters would be non-breaking but they aren't). This is causing my applications that reference a library that uses a library that uses Confluent.Kafka to break when the entry app tries to upgrade to the latest Confluent.Kafka.

See #2060


rising-4ce (approved):
I completely agree with the importance of merging this pull request.

anchitj:
Closing this as your commit is merged with PR #2066.

PR #2066, opened 2023-06-27
Fix Backward Compatibility Issue in TopicPartitionOffset Constructor

### LINKED ISSUE #2060: Unintentional Breaking Change to TopicPartitionOffset in 2.1

anchitj:
Fix a backward compatibility issue in the TopicPartitionOffset constructor by reintroducing the original constructor and overloading it to separately accommodate the optional parameters.

* Reintroduces the original constructor.
* Adds an overloaded constructor for additional parameters.

Referring - https://github.com/confluentinc/confluent-kafka-dotnet/pull/2061 and fixes #2060

___

### Testing

Integration tests

### DIFF
--- a/src/Confluent.Kafka/TopicPartitionOffset.cs
+++ b/src/Confluent.Kafka/TopicPartitionOffset.cs
@@ -34,6 +34,7 @@ namespace Confluent.Kafka
         /// </param>
         public TopicPartitionOffset(TopicPartition tp, Offset offset)
             : this(tp.Topic, tp.Partition, offset, null) { }
+        
         /// <summary>
         ///     Initializes a new TopicPartitionOffset instance.
         /// </summary>
@@ -62,12 +63,9 @@ namespace Confluent.Kafka
         /// <param name="offset">
         ///     A Kafka offset value.
         /// </param>
-        /// <param name="leaderEpoch">
-        ///     The optional offset leader epoch.
-        /// </param>
         public TopicPartitionOffset(string topic, Partition partition,
                                     Offset offset)
-            : this(topic, tp.Partition, offset, null) { }
+            : this(topic, partition, offset, null) { }
 
         /// <summary>
         ///     Initializes a new TopicPartitionOffset instance.
--- a/src/Confluent.Kafka/TopicPartitionOffset.cs
+++ b/src/Confluent.Kafka/TopicPartitionOffset.cs
@@ -32,12 +32,23 @@ namespace Confluent.Kafka
         /// <param name="offset">
         ///     A Kafka offset value.
         /// </param>
+        public TopicPartitionOffset(TopicPartition tp, Offset offset)
+            : this(tp.Topic, tp.Partition, offset, null) { }
+        /// <summary>
+        ///     Initializes a new TopicPartitionOffset instance.
+        /// </summary>
+        /// <param name="tp">
+        ///     Kafka topic name and partition.
+        /// </param>
+        /// <param name="offset">
+        ///     A Kafka offset value.
+        /// </param>
         /// <param name="leaderEpoch">
         ///     The offset leader epoch (optional).
         /// </param>
         public TopicPartitionOffset(TopicPartition tp, Offset offset,
-                                    int? leaderEpoch = null)
-            : this(tp.Topic, tp.Partition, offset, leaderEpoch) {}
+                                    int? leaderEpoch)
+            : this(tp.Topic, tp.Partition, offset, leaderEpoch) { }
 
         /// <summary>
         ///     Initializes a new TopicPartitionOffset instance.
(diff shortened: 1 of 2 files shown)


src/Confluent.Kafka/TopicPartitionOffset.cs
```diff
@@ -36,8 +48,24 @@ public class TopicPartitionOffset
         ///     The offset leader epoch (optional).
         /// </param>
         public TopicPartitionOffset(TopicPartition tp, Offset offset,
-                                    int? leaderEpoch = null)
```
pranavrth, line 39: I don't know much about dotnet. What does this mean? `int? leaderEpoch = null`. 

If it was in Python, it would mean to take leaderEpoch as null if not provided. In that case the change wouldn't have caused any issue. I don't think this is the case here. Then what does this statement mean?

anchitj: `int?` means its nullable, ` = null` means null by default.

The issue stemmed from a transitive dependency.

Imagine L2 uses an older version of our library, which includes the original `TopicPartitionOffset` constructor. If an application A references the latest versions of both confluent-kafka-dotnet and L2, a `MissingMethodException` will be thrown if A calls an L2 method that relies on the old constructor, which was removed in the latest confluent-kafka-dotnet version.

To resolve this, we’re reintroducing the original constructor to ensure compatibility with libraries like L2 that haven't upgraded to the latest confluent-kafka-dotnet version.

pranavrth: As discussed, this should have been working with optional parameter but its not working in transitive dependency.

pranavrth (approved):
LGTM!

tshowers-bt:
https://github.com/confluentinc/confluent-kafka-dotnet/pull/2027/files also introduced a breaking change to the `IConsumer` interface by adding `TopicPartitionOffset PositionTopicPartitionOffset(TopicPartition partition);` in 2.1

ttd2089:
I opened [this PR](https://github.com/confluentinc/confluent-kafka-dotnet/pull/2067) to replace the new interface method with an extension method on the `IConsumer` type; though it might be better to just keep the new method private.

src/Confluent.Kafka/TopicPartitionOffset.cs
rising-4ce: > I don't know much about dotnet. What does this mean? `int? leaderEpoch = null`.
> 
> If it was in Python, it would mean to take leaderEpoch as null if not provided. In that case the change wouldn't have caused any issue. I don't think this is the case here. Then what does this statement mean?

The difference between a constructor with an optional parameter and one without parameters is very clear in the intermediate code.
C# code:
```csharp
public class TestClass1
{
    public TestClass1(int? someValue = null)
    {
    }
}

public class TestClass2
{
    public TestClass2(int? someValue)
    {
    }

    public TestClass2() : this(null)
    {
    }
}

public static class TestClass
{
    public static void InitializeTestClass1InstanceWithoutParameters()
    {
        var instance = new TestClass1();
    }

    public static void InitializeTestClass2InstanceWithoutParameters()
    {
        var instance = new TestClass2();
    }
}
```

IL code:

image

In the case of optional parameters, their values are still passed to the constructor, but implicitly.

Moreover, the distinction between an optional parameter and no parameter allows the following set of constructors to be used without compilation errors (the default parameter value in this case is redundant: the constructor without an optional parameter takes precedence):
```csharp
public class TestClass1
{
    public TestClass1(int? someValue = null)
    {
    }

    public TestClass1()
    {
    }
}
```

PR #2067, opened 2023-06-27
fix: remove breaking change to IConsumer

ttd2089:
A method was added to the IConsumer interface by #2027 that resulted in a breaking change in the 2.1 release. This moves replaces the interface method and the implementation on Consumer with an extension method on IConsumer to ensure that existing implementations of IConsumer remain unbroken and have the same functionality as the default Consumer implementation.

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -263,7 +263,7 @@ namespace Confluent.Kafka
                     {
                         try
                         {
-                            assignmentWithPositions.Add(PositionTopicPartitionOffset(tp));
+                            assignmentWithPositions.Add(this.PositionTopicPartitionOffset(tp));
                         }
                         catch
                         {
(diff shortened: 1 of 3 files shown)


ttd2089:
It might be actually be best to keep this new method private instead. The references are in the implementation of `Consumer` and in some tests so it doesn't look like the intention was actually to extend the public API.

ttd2089:
This also results in a strange scenario where the `Consumer` implementation of `IConsumer` references the extension method in its own implementation. The `IConsumer` abstraction itself seems like it might be somewhat overloaded. Some of its methods like `Subscribe` and `Consume` seem like places where a user might want to extend the behaviour of `Consumer` by wrapping it in their own implementation of `IConsumer`, but a significant number of the other methods like `OffsetsForTimes` and `Position` seem like you want just want to inherit them from the wrapped implementation. This can by done by forwarding calls but it results in the situation here where you can't extend the interface without introducing a breaking change. A lot of the functionality on `Consumer` is really just using `Handle` and in those cases they could implemented as extension methods on `IConsumer` that use its handle -- except for the fact that they already exist in the public API.

anchitj:
Thanks for the PR. Looks good overall, I've created a PR including this change to run against our CI https://github.com/confluentinc/confluent-kafka-dotnet/pull/2071.

anchitj:
Closing this PR, as we've included your commit in the PR #2071.

PR #2081, opened 2023-07-11
Increase number of network threads

emasab:
in integration test to avoid disconnections from broker 0.

It seems the broker is disconnecting when available threads are exhausted, this problem is more evident when number of broker thread was reduced to 2 instead of the default of 3.


anchitj (approved):
Thanks for fixing this.

PR #2088, opened 2023-07-20
Use object mapping when serializing

emasab:
and deserializing with referenced schemas.


examples/JsonWithReferences/Program.cs
```diff
@@ -165,17 +145,25 @@ static async Task Main(string[] args)
...
+            var jsonSchemaGeneratorSettings = new JsonSchemaGeneratorSettings
+            {
+                SerializerSettings = new JsonSerializerSettings
+                {
+                    ContractResolver = new DefaultContractResolver
+                    {
+                        NamingStrategy = new CamelCaseNamingStrategy()
+                    }
+                }
+            };
```
anchitj, line 158: Is the user always supposed to provide this? Won't it make sense that we keep this as default `JsonSchemaGeneratorSettings` for both `JsonSerializer` and `JsonDeserializer` if the user isn't providing any `JsonSchemaGeneratorSettings`?

emasab: It's not required to provide this, this allows you to set the default strategy and avoid to add attributes like
`[JsonProperty("favoriteColor")]` to the fields, but that is also possible. I'd not change JsonSerializer defaults, also it would be a breaking change (see PersonPoco in tests).
I can add a comment about this.

PR #2096, opened 2023-08-07
Methods for working with compatibility

ISBronny:
Hi!

I've added the missing compatibility types and methods to work with them.


ISBronny:
I made commit without e-mail and idk how to fix it. I will create new PR

PR #2041, opened 2023-04-11
Allow proxy to be specified

djam-nbs:
fixes #620

PR #2128, opened 2023-10-20
Allow proxy to be specified (#2127)

emasab:
Imported from #2041


rayokota (approved):
Thanks @emasab , LGTM

PR #2131, opened 2023-10-24
Fix calls with empty lists

emasab:
DescribeTopics and ListOffsets should return an empty output if and empty input is passed as parameter.
Consistent with Java and user friendly for users using filter and map.


anchitj (approved):
LGTM! CI dependent on https://github.com/confluentinc/confluent-kafka-dotnet/pull/2132

PR #2136, opened 2023-11-04
Added schema id check to the JsonSchemaResolver

gokerakc:
**Issue:** $id property is not required for JSON schemas but in the **JsonSchemaResolver** class we need to use the schema id value and there is no null check in place for it. 

**Current behavior:** `JsonSchemaResolver` throws the following error when the provided schema does not have a schema id;

- `System.ArgumentNullException: 'Value cannot be null. Arg_ParamName_Name'` (error reason is not clear)

**Desired behavior:** `JsonSchemaResolver` throws the following exception when the schema id is null;

- `System.Exception: 'Provided schema does not have $id property.'`

If you have a different solution for this issue or any suggestion for the exception message please let me know.


rayokota:
Null checked added here https://github.com/confluentinc/confluent-kafka-dotnet/pull/2226

PR #2174, opened 2024-01-17
(fix #1897) : skip server SSL certs verification when configured

### LINKED ISSUE #1897: SchemaRegistryConfig.EnableSslCertificateVerification does not allow to bypass CA check when client certificates is not used
Description
===========

I found in the sources, that when HttpClient is created, it bypasses the _server_ SSL CA check only when we're using _client_ certs.

https://github.com/confluentinc/confluent-kafka-dotnet/blob/895b72d1f7e983e9512185aab667ba142e0a9839/src/Confluent.SchemaRegistry/Rest/RestService.cs#L67-L78

This does not work, when we're accessing a SchemaRegistry that is served via https with self-signed certificate, without using client certificates.

How to reproduce
================

* Use a self-signed certificate for the schema-registry
* set `EnableSslCertificateVerification = true`
* try to get schema, receive an error
> HttpRequestException: The SSL connection could not be established, see inner exception.

```c#
var schemaRegistryConfig = new SchemaRegistryConfig
{
  Url = "https://localhost:8085",
  EnableSslCertificateVerification = false  // this does not work
};

using (var schemaRegistry = new CachedSchemaRegistryClient(schemaRegistryConfig))
{
  var subjects = await schemaRegistry.GetAllSubjectsAsync();

  foreach (var subject in subjects)
  {
    Console.WriteLine(subject);
  }
}
```

Checklist
=========

 - [x] A complete (i.e. we can run it), minimal program demonstrating the problem. No need to supply a project file.
 - [x] Confluent.Kafka nuget version - 1.8.2

dolifer:
An attempt to fix #1897 

I've updated `RestService` to respect the`enableSslCertificateVerification` option, even when there are no client certificates provided.

This ensures that in case of self-signed certificate, we can get schema from schema-registry, by skiping certificate validation.


dolifer:
@mhowlett hello, here is my attempt to fix the issue with certificates validation.

jorgebg90:
Hello,
Is there any update on this? @dolifer your solution jumps over the self-signed certificate or uses the specified one with the property SslCaLocation?

Thanks in advance :)

dolifer:
@jorgebg90 it's jumps over, and skips the validation at all by using custom `ServerCertificateCustomValidationCallback` which "trusts" ANY given certificate.

rayokota:
The commits in this PR have been merged via https://github.com/confluentinc/confluent-kafka-dotnet/pull/2347

PR #2177, opened 2024-01-22
Add a new type of Serializer for supporting zero allocation scenarios

nmosafi:
Right now `ISerializer` and `ISerializerAsync` are forced to allocate a new array for every message which is sent. For a high throughput application we'd like to avoid allocating large amounts of memory.

The API exposed by librdkafka supports buffer reuse as it accepts a length and offset alongside each byte array, however this is not exposed by the serialization framework in Kafka dotnet.

The simplest approach here is to allow `ArraySegment<byte>` to be returned by the serializer rather than `byte[]`, however changing all serializers to return `ArraySegment<byte>` would surely be a large breaking change. So instead I decided to implement this as a new type of serializer.

Please give me some feedback and thoughts on this PR.

Thanks

### DIFF
--- a/src/Confluent.Kafka/DependentProducerBuilder.cs
+++ b/src/Confluent.Kafka/DependentProducerBuilder.cs
@@ -33,7 +33,7 @@ namespace Confluent.Kafka
         ///     The configured client handle.
         /// </summary>
         public Handle Handle { get; set; }
-        
+
         /// <summary>
         ///     The configured key serializer.
         /// </summary>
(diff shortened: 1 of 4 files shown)


nanov:
Is there a reason why this is not looked at, such functionality could allow, potentially, huge performance benefits in scenarios where the implementations relays on pooled arrays ( only minimum length is guaranteed, array may be bigger depending on the pool ) or ring buffers.

nmosafi:
Seems there are newer PRs for this out there so I close this one

PR #2193, opened 2024-02-22
#2181 Remove System.Memory from Dotnet 6

### LINKED ISSUE #2181: Only depend on System.Memory when needed
#### summary
On newer TFM'S System.Memory is included as part of the framework hence the dependency is not needed.

### Proposal
Adjust the conditions on the package dependency in Confluent.Kafka so that it won't be added for the newer TFM

thompson-tomo:
Remove System.Memory as a dependency for dotnet 6 TFM

closes #2181


rayokota (approved):
Thanks @thompson-tomo , LGTM

PR #2207, opened 2024-04-11
Add "CreateReadOnlySpanFromNullTerminated" on NET6-specific path

stevenaw:
Adds a .NET6-specific implementation of `PtrToStringUTF8` to use the built-in [`MemoryMarshal.CreateReadOnlySpanFromNullTerminated()`](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.memorymarshal.createreadonlyspanfromnullterminated?view=net-8.0#system-runtime-interopservices-memorymarshal-createreadonlyspanfromnullterminated(system-byte*)) function to satisfy the below existing inline comment in the source:

> // TODO: Is there a built in / vectorized / better way to implement this? 

This API was added in .NET6 for this purpose (when an interop call only returns a pointer and the length of the underlying null-terminated string is unknown) - https://github.com/dotnet/runtime/issues/40202. The internal source uses [`strlen()`](https://github.com/dotnet/runtime/blob/5535e31a712343a63f5d7d796cd874e563e5ac14/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/MemoryMarshal.cs#L267C13-L268C20) which internally delegates to [`SpanHelpers.IndexOfNullByte`](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/SpanHelpers.Byte.cs#L453) which is heavily optimized with vectorization or other special cases.

Performance is comparable on short strings and further improves as string length increases

<details>
  <summary>Benchmarks (using .NET8 runtime) </summary>

```pre
| Method   | Mean     | Error    | StdDev   | Ratio | Gen0   | Gen1   | Allocated | Alloc Ratio |
|--------- |---------:|---------:|---------:|------:|-------:|-------:|----------:|------------:|
| Original | 12.05 ns | 0.110 ns | 0.103 ns |  1.00 | 0.0064 | 0.0000 |      40 B |        1.00 |
| Updated  | 11.70 ns | 0.042 ns | 0.040 ns |  0.97 | 0.0064 |      - |      40 B |        1.00 |
```
  
```
    [MemoryDiagnoser]
    public class UnsafeIntPtrToStr
    {
        GCHandle handle;
        IntPtr strPtr;

        [GlobalSetup]
        public void Setup()
        {
            var strBytes = Encoding.UTF8.GetBytes("TestData");
            byte[] strBytesNulTerminated = new byte[strBytes.Length + 1]; // initialized to all 0's.
            Array.Copy(strBytes, strBytesNulTerminated, strBytes.Length);

            handle = GCHandle.Alloc(strBytesNulTerminated, GCHandleType.Pinned);
            strPtr = handle.AddrOfPinnedObject();
        }

        [Benchmark(Baseline = true)]
        public unsafe string Original()
        {
#if NET6_0_OR_GREATER
                var bytes = MemoryMarshal.CreateReadOnlySpanFromNullTerminated((byte*)strPtr.ToPointer());
                return Encoding.UTF8.GetString(bytes);
#else
            // TODO: Is there a built in / vectorized / better way to implement this?              
            byte* pTraverse = (byte*)strPtr;
            while (*pTraverse != 0) { pTraverse += 1; }
            var length = (int)(pTraverse - (byte*)strPtr);

            return Encoding.UTF8.GetString((byte*)strPtr.ToPointer(), length);
#endif
        }

        [Benchmark]
        public unsafe string Updated()
        {
#if NET6_0_OR_GREATER
                var bytes = MemoryMarshal.CreateReadOnlySpanFromNullTerminated((byte*)strPtr.ToPointer());
                return Encoding.UTF8.GetString(bytes);
#else
            // TODO: Is there a built in / vectorized / better way to implement this?              
            byte* pTraverse = (byte*)strPtr;
            while (*pTraverse != 0) { pTraverse += 1; }
            var length = (int)(pTraverse - (byte*)strPtr);

            return Encoding.UTF8.GetString((byte*)strPtr.ToPointer(), length);
#endif
        }


        [GlobalCleanup]
        public void Cleanup()
        {
            handle.Free();
            strPtr = IntPtr.Zero;
        }
    }
```
  ```
</details>

### DIFF
--- a/src/Confluent.Kafka/Internal/Util.cs
+++ b/src/Confluent.Kafka/Internal/Util.cs
@@ -16,6 +16,9 @@
 
 using System;
 using System.Text;
+#if NET6_0_OR_GREATER
+using System.Runtime.InteropServices;
+#endif
 using SystemMarshal = System.Runtime.InteropServices.Marshal;
 using SystemGCHandle = System.Runtime.InteropServices.GCHandle;
 using SystemGCHandleType = System.Runtime.InteropServices.GCHandleType;
@@ -31,7 +34,7 @@ namespace Confluent.Kafka.Internal
             ///     Convenience class for generating and pinning the UTF8
             ///     representation of a string.
             /// </summary>
-            public class StringAsPinnedUTF8 : IDisposable
+            public sealed class StringAsPinnedUTF8 : IDisposable
             {
                 private SystemGCHandle gch;
 
@@ -60,13 +63,18 @@ namespace Confluent.Kafka.Internal
                 {
                     return null;
                 }
-                
+
+#if NET6_0_OR_GREATER
+                var bytes = MemoryMarshal.CreateReadOnlySpanFromNullTerminated((byte*)strPtr.ToPointer());
+                return Encoding.UTF8.GetString(bytes);
+#else
                 // TODO: Is there a built in / vectorized / better way to implement this?              
                 byte* pTraverse = (byte*)strPtr;
                 while (*pTraverse != 0) { pTraverse += 1; }
                 var length = (int)(pTraverse - (byte*)strPtr);
 
                 return Encoding.UTF8.GetString((byte*)strPtr.ToPointer(), length);
+#endif
             }
 
             public unsafe static string PtrToStringUTF8(IntPtr strPtr, UIntPtr strLength)


Havret:
This can be closed as https://github.com/confluentinc/confluent-kafka-dotnet/pull/2466 was already merged.

PR #2212, opened 2024-04-23
[KIP-848] Make Kafka integration test run with new protocol too

emasab:


test/Confluent.Kafka.TestsCommon/TestConsumerBuilder.cs
```diff
@@ -0,0 +1,32 @@
...
+        var consumerConfig = new ConsumerConfig(
+            new Dictionary<string, string>(config)) {};
+
+        var groupProtocol = TestConsumerGroupProtocol.GroupProtocol();
+        if (groupProtocol != null)
+        {
+            consumerConfig.GroupProtocol = groupProtocol == "classic" ?
+                GroupProtocol.Classic :
+                GroupProtocol.Consumer;
+        }
```
anchitj, line 27: Can be simplified using `TestConsumerGroupProtocol.IsClassic()`

test/Confluent.Kafka.IntegrationTests/Tests/Tests.cs
```diff
@@ -94,7 +94,7 @@ private void LogToFile(string msg)
             {
                 // Uncomment to enable logging to a file. Useful for debugging,
                 // for example, which test caused librdkafka to segfault.
-                // File.AppendAllLines("/tmp/test.txt", new [] { msg });
+                File.AppendAllLines("/tmp/test.txt", new [] { msg });
```
anchitj, line 97: Can be reverted

anchitj (changes requested):

test/Confluent.Kafka.TestsCommon/TestConsumerBuilder.cs
emasab: I left it like it is because `TestConsumerGroupProtocol.IsClassic()` returns true even if it's not set, and in that case we don't want to set property explicitly but rely on the default

test/docker/kraft/Dockerfile
```diff
@@ -4,7 +4,10 @@ RUN apt update
 RUN apt install -y git curl
 RUN apt install -y openjdk-17-jdk
 
-RUN git clone https://github.com/apache/kafka.git
+# Trunk at 2024-04-10, replace with 3.8.0 tag when available
+RUN git clone --single-branch --branch trunk \
+    https://github.com/apache/kafka.git && \
+    (cd /kafka  && git checkout f6c9feea76d01a46319b0ca602d70aa855057b07)
```
anchitj, line 10: Do we need to update the commit to use everytime?

emasab: Until we switch to the tag and only when we need it. To avoid random test failures because of new commits in AK trunk

anchitj (approved):

PR #2225, opened 2024-05-29
Add a UTF-8 statistics handler

verdie-g:
The statistics payload can be quite large (several MB) and currently it can be copied twice in the managed code:
1. The payload is received from the C driver
2. [A .NET string is allocated for the JSON](https://github.com/confluentinc/confluent-kafka-dotnet/blob/32948fe3b70ce03a541fbd5026bb226848525f42/src/Confluent.Kafka/Consumer.cs#L115)
3. The statistics handler is called
4. The subscriber will likely call System.Text.Json.Serialize.Deserialize [which cause the string to be converted back to UTF-8](https://github.com/dotnet/runtime/blob/5c74f63871ed17624279ecbccd011de4d7dc5caa/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/JsonSerializer.Read.String.cs#L426-L439)

This round-trip causes important memory fragmentation which puts pressure on the GC. In .NET 6, this can be avoided using `Span`.

### DIFF
--- a/src/Confluent.Kafka/Consumer.cs
+++ b/src/Confluent.Kafka/Consumer.cs
@@ -17,6 +17,7 @@
 // Refer to LICENSE for more information.
 
 using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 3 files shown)


verdie-g:
@emasab @rayokota could you have a look at this PR, it's been open for more than a year.

PR #2228, opened 2024-06-03
Add data-governance to code owners for schema registry clients

emasab:


.github/CODEOWNERS
```diff
@@ -1 +1 @@
-* @confluentinc/clients
+*	@confluentinc/clients @confluentinc/data-governance
```
anchitj, line 1: Is the spacing in front intentional?

emasab: Taken from a different file, it's a tab instead of a space

PR #2238, opened 2024-06-26
DOCS-29466 - Clarify description

joel-hamill:


milindl (approved):
Approving it as part of the larger verbiage effort

PR #2270, opened 2024-07-29
Fix Azure algorithm to RsaOaep256

rayokota:
Fix algorithm used by Azure Key Vault to be RsaOaep256, to be consistent with the Java and golang clients, so that the dotnet client can interoperate with them.


sakshamkumar1997 (approved):
Thanks @rayokota

sakshamkumar1997 (approved):
lgtm!

sakshamkumar1997 (approved):

PR #2285, opened 2024-08-16
Add editorconfig

anchitj:
Also format the whole codebase


Havret (changes requested):
Big -1. I see value in adding an .editorconfig, but formatting the entire codebase would be disruptive, to say the least.

PR #1603, opened 2021-05-03
Binary Producer with Span<T>

alec-anikin:
Hi Matt,
I've prepared the draft for binary producer with span api. 
I tried to minimize the number of changes to make it easier to review.

PR #2311, opened 2024-09-13
Natively support Memory<byte>

verdie-g:
There are many PRs open aiming to extend the API to allow low allocation produce but they all have some caveats:
- https://github.com/confluentinc/confluent-kafka-dotnet/pull/1603 tries to support `Span` but given the constraints of that type, it means creating a whole new set of API.
- https://github.com/confluentinc/confluent-kafka-dotnet/pull/1725 suggests to use `IBufferWriter` instead, which unlike `byte[]` or `Span`, supports non-contiguous memory, kind of like `Stream` but it enables a lof of ways to reduce allocations. I don't think `IBufferWriter` is ideal here because in the end, a continuous memory buffer is expected for both the key and value. Also, the PR is introducing several breaking changes.
- https://github.com/confluentinc/confluent-kafka-dotnet/pull/2177 and https://github.com/confluentinc/confluent-kafka-dotnet/pull/2219 both propose to support `ArraySegment` which does not break the API and fixes the original problem. Though it would prevent the use of `Memory<byte>` which is the preferred type when dealing with contiguous memory when `Span<byte>` can't be used. This type is for example used in <a href="https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.writeasync?view=net-8.0#system-io-stream-writeasync(system-readonlymemory((system-byte))-system-threading-cancellationtoken)">Stream.WriteAsync</a>, <a href="https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.socket.sendasync?view=net-8.0#system-net-sockets-socket-sendasync(system-readonlymemory((system-byte))-system-threading-cancellationtoken)">Socket.SendAsync</a>,  and also in <a href="https://learn.microsoft.com/en-us/dotnet/api/system.buffers.memorypool-1.rent?view=net-8.0#system-buffers-memorypool-1-rent(system-int32)">MemoryPool.Rent</a>.

Instead of adding a third type of serializer (after `ISerializer` and `IAsyncSerializer`), this PR makes a special case for `Memory<byte>` (and `ReadOnlyMemory<byte>`) to by pass the serializers and directly use `Message.Value` or `Message.Key`.

### DIFF
--- a/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
+++ b/src/Confluent.Kafka/Impl/SafeKafkaHandle.cs
@@ -18,6 +18,7 @@
 // Refer to LICENSE for more information.
 
 using System;
+using System.Buffers;
 using System.Collections.Generic;
 using System.Linq;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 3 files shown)


bmesetovic:
There seems to be no progress on either of the several approaches to reduce allocations on the publishing side of Kafka. All of them are stuck at the review phase.

We are affected by the issue as well and would really like to see how we can push this topic so that it finally gets the attention it deserves. 

@verdie-g Did you end up forking the project and adding your changes?

verdie-g:
We used a fork with https://github.com/confluentinc/confluent-kafka-dotnet/pull/2219 just to confirm the important performance benefits but we were not ready to maintain a fork.

We are still waiting for a review on that PR.

verdie-g:
@Claimundefine @rayokota could you have a look at his PR 🙏 It's been sitting for a while and I believe it fixes an important performance issue in the client.

verdie-g:
or @anchitj @emasab 🥺

Scylin232:
I also would love to see this feature, @anchitj @emasab, give this Dino some hugs from your side

PR #2338, opened 2024-11-02
Ensure proper reader schema in rules for deserializer

rayokota:
Ensure proper reader schema in rules for deserializer.  Also adds a test


kcorman0 (approved):
thanks LGTM

PR #2339, opened 2024-11-06
DGS-18985 Fix JSON Schema ref handling

rayokota:
Fix JSON Schema ref handling.  Before the code relied on the reference name also being the `$id` for the referenced schema.


kcorman0 (approved):
thanks

Issue #2218, opened 2024-05-08
ReferenceSubjectNameStrategy is missing option for QualifiedReferenceSubjectNameStrategy


mattstermiller:
Description
===========
The [ReferenceSubjectNameStrategy](https://github.com/confluentinc/confluent-kafka-dotnet/blob/32948fe3b70ce03a541fbd5026bb226848525f42/src/Confluent.SchemaRegistry/ReferenceSubjectNameStrategy.cs#L38) enum only defines the default "ReferenceName" and does not define the alternate "Qualified" strategy described in [this Schema Registry](https://docs.confluent.io/platform/current/schema-registry/fundamentals/serdes-develop/serdes-protobuf.html#protobuf-serializer) documentation. The implementation of the ProtobufSerializer also keeps the reference subject naming completely closed to outside extension with no way to override this behavior.

## Motivation
We are running into an issue where we have two completely different proto packages with files that have the same relative path that are imported into protos used for schemas.

For Example:

- DomainA/events/ModelA.proto:
  ```protobuf
  import "shared/Enums.proto";
  // ...
  ```

- DomainA/shared/Enums.proto:
  ```protobuf
  package domain_a;
  enum ItemTypes { }
  ```

- DomainB/events/ModelB.proto:
  ```protobuf
  import "shared/Enums.proto";
  // ...
  ```

- DomainB/shared/Enums.proto:
  ```protobuf
  package domain_b;
  enum ThingTypes { }
  ```

Since the reference subject name strategy does not take the package name of the imported proto into account, we have a conflict where our clients are trying to register these different schemas under the same name and we get this error:
> Confluent.SchemaRegistry.SchemaRegistryException: Schema being registered is incompatible with an earlier schema for subject "shared/Enums.proto", details: [{errorType:"PACKAGE_CHANGED", description:"The package at '#/' in the old schema does not match the package in the new schema"}

We want to use the package name as part of the subject name to avoid having to ensure that all of our imported proto file names are unique across our organization.

How to reproduce
================
N/A

Checklist
=========

Please provide the following information:
 - [ ] A complete (i.e. we can run it), minimal program demonstrating the problem. No need to supply a project file.
 - [x] Confluent.Kafka nuget version. `2.4.0`
 - [ ] Apache Kafka version. N/A
 - [ ] Client configuration. N/A
 - [ ] Operating system. N/A
 - [ ] Provide logs (with "debug" : "..." as necessary in configuration). N/A
[...]

PR #2345, opened 2024-11-12
Feature: Added Qualified and Custom reference name strategy approaches for protobuf references (Schema Registry)

RobotechUSA:
# Description
This feature allows producers to specify configuration for ReferenceSubjectNameStrategy the following enumeration values:
- `ReferenceName` -> (default): Use the reference name as the subject name.
- `Qualified` -> Given a reference name, replace slashes with dots, and remove the .proto suffix to obtain the subject name. For example, mypackage/myfile.proto becomes mypackage.myfile. (This provides feature parity with java client)
- `Custom` -> Use a custom reference subject name strategy resolver to determine the subject name. This is done by implementing ICustomReferenceSubjectNameStrategy and assigning instance to ProtobufSerializerConfig property CustomReferenceSubjectNameStrategy

## Background
An internal team is using java has implemented their Schema Registry with `QualifiedReferenceSubjectNameStrategy`. As a dotnet team we need to interoperate with that work. Unfortunately, we cannot without this proposed change (We require feature parity with java). Further, we are not allowed to auto register schemas.

## Other Information
- Addresses Issue: [ReferenceSubjectNameStrategy is missing option for QualifiedReferenceSubjectNameStrategy](https://github.com/confluentinc/confluent-kafka-dotnet/issues/2218)
- [Work inspired by](https://github.com/confluentinc/confluent-kafka-dotnet/pull/1656)

### DIFF
--- a/src/Confluent.SchemaRegistry.Serdes.Protobuf/ProtobufSerializer.cs
+++ b/src/Confluent.SchemaRegistry.Serdes.Protobuf/ProtobufSerializer.cs
@@ -52,7 +52,7 @@ namespace Confluent.SchemaRegistry.Serdes
     ///                            a single 0 byte as an optimization.
     ///                         2. The protobuf serialized data.
     /// </remarks>
-    public class ProtobufSerializer<T> : AsyncSerializer<T, FileDescriptorSet>  where T : IMessage<T>, new()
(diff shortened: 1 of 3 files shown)


rayokota (approved):
Thanks @RobotechUSA , LGTM

PR #2362, opened 2024-11-20
Add support for .NET 9

ffernandolima:
- Removed support for .NET 6 as it has reached the end of support;
- Added support for .NET 9;
- Set 'LangVersion' to 'latest'.

### DIFF
--- a/examples/AdminClient/AdminClient.csproj
+++ b/examples/AdminClient/AdminClient.csproj
@@ -3,9 +3,9 @@
   <PropertyGroup>
     <ProjectTypeGuids>{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}</ProjectTypeGuids>
     <AssemblyName>AdminClient</AssemblyName>
-    <TargetFramework>net6.0</TargetFramework>
+    <TargetFramework>net9.0</TargetFramework>
     <OutputType>Exe</OutputType>
-    <LangVersion>7.3</LangVersion>
+    <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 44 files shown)


Havret:
Would it make sense to add support for a version of .NET that isn't LTS?

ffernandolima:
> Would it make sense to add support for a version of .NET that isn't LTS?

@havret, I agree with you to some extent, and I believe this is open for discussion. However, the primary objective of this PR is to remove .NET 6, as it has reached its end of support and should no longer be used in active development.

Im5tu:
> > Would it make sense to add support for a version of .NET that isn't LTS?
> 
> @Havret, I agree with you to some extent, and I believe this is open for discussion. However, the primary objective of this PR is to remove .NET 6, as it has reached its end of support and should no longer be used in active development.

Could just do a multi-target so you always have current STS + LTS versions, ie: net8/net9.

ffernandolima:
> Could just do a multi-target so you always have current STS + LTS versions, ie: net8/net9.

@Im5tu, that was exactly what I did through this PR. Projects released as Nuget packages support multi-target.

lgcmotta:
> Would it make sense to add support for a version of .NET that isn't LTS?

I think yes, it makes sense. Each yearly release, whether STS or LTS, includes valuable performance improvements, new features, and bug fixes. If you only upgrade every two years, you risk missing out on incremental enhancements. Another point is: that by updating more frequently, you typically face smaller, more manageable changes rather than a large, potentially disruptive jump.
Although you’re unlikely to encounter major API-breaking changes, there could still be under-the-hood modifications that introduce subtle issues if you skip multiple versions.

PR #2370, opened 2024-11-28
Improve lock utilization on the consumer hot path

Havret:
This PR addresses a significant performance issue observed when enabling the schema registry for consumers using the Protobuf deserializer. The consumption time for 600k messages shows a drastic difference:

- **With Schema Registry:**  
  Elapsed time: `00:00:33.7949289`

- **Without Schema Registry:**  
  Elapsed time: `00:00:07.8086635`

The root cause was identified as the usage of `SemaphoreSlim` locks with a regular `.NET Dictionary` on the consumer's hot path. This PR optimizes the implementation to mitigate this bottleneck, aligning the performance closer to Java's implementation, which does not suffer from this issue.

### DIFF
--- a/src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
+++ b/src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
@@ -23,6 +23,7 @@ using System.Collections.Generic;
 using System.Threading.Tasks;
 using System.Linq;
 using System;
+using System.Collections.Concurrent;
 using System.Net;
 using System.Threading;
 using System.Security.Cryptography.X509Certificates;
@@ -65,7 +66,7 @@ namespace Confluent.SchemaRegistry
         private IRestService restService;
         private int identityMapCapacity;
         private int latestCacheTtlSecs;
-        private readonly Dictionary<int, Schema> schemaById = new Dictionary<int, Schema>();
+        private readonly ConcurrentDictionary<int, Schema> schemaById = new ConcurrentDictionary<int, Schema>();
 
         private readonly Dictionary<string /*subject*/, Dictionary<Schema, int>> idBySchemaBySubject =
             new Dictionary<string, Dictionary<Schema, int>>();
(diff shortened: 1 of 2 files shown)


src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
```diff
@@ -657,11 +658,15 @@ public async Task<Schema> GetSchemaAsync(int id, string format = null)
         /// <inheritdoc/>
         public async Task<Schema> GetSchemaBySubjectAndIdAsync(string subject, int id, string format = null)
         {
+            if (this.schemaById.TryGetValue(id, out Schema schema) && !checkSchemaMatchesFormat(format, schema.SchemaString))
```
Havret, line 661: I’m wondering if we could eliminate the `checkSchemaMatchesFormat` method. In my workaround, I created a cache that uses all three parameters as the key, which allowed me to skip this check. One idea that comes to mind is extending the `Schema` object with an additional property to store the format information.  

@rayokota, what are your thoughts on this approach?

Havret: .NET 8 introduces a really nice method that allows you to check if a string is valid Base64 encoding without any allocations, making this check a no-op in terms of performance.

Reference: [Base64.IsValid](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.text.base64.isvalid?view=net-8.0)

We could introduce a helper method that provides a performance boost for .NET 8 and onward compilations while either falling back to the old mechanism for older .NET versions or offering a polyfill.

Here’s the code snippet:

```csharp
internal static bool IsBase64String(string value)
{
#if NET8_0_OR_GREATER
    return Base64.IsValid(value);
#else
    try
    {
        _ = Convert.FromBase64String(value);
        return true;
    }
    catch (FormatException)
    {
        return false;
    }
#endif
}
```

rayokota (approved):
Thanks @Havret , LGTM

PR #2371, opened 2024-11-29
Improve perf of base64 encoding check on a consumer hot path

Havret:
This PR improves the performance of the Base64 encoding check, which is performed for every message processed by the consumer. This change complements the work done in #2370.

Below are the results of benchmarking the new implementation (Base64.IsValid) against the existing Convert.FromBase64String approach:

```
| Method           | Mean          | Error        | StdDev       |
|----------------- |--------------:|-------------:|-------------:|
| IsBase64         |      24.52 ns |     0.074 ns |     0.066 ns |
| FromBase64String | 138,883.49 ns | 2,535.710 ns | 2,247.841 ns |

```

The code used to produce the benchmark results::
```csharp
class Program
{
    static void Main(string[] args)
    {
        BenchmarkRunner.Run<IsBase64StringBenchmark>();
    }
}

public class IsBase64StringBenchmark
{
    private readonly string _invalidBase64 = string.Join("", Enumerable.Repeat("This is not base64!", 10000));

    [Benchmark]
    public bool IsBase64()
    {
        return Base64.IsValid(_invalidBase64);
    }

    [Benchmark]
    public bool FromBase64String()
    {
        try
        {
            _ = Convert.FromBase64String(_invalidBase64);
            return true;
        }
        catch (FormatException)
        {
            return false;
        }
    }
}
```

### DIFF
--- a/src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
+++ b/src/Confluent.SchemaRegistry/CachedSchemaRegistryClient.cs
@@ -592,16 +592,8 @@ namespace Confluent.SchemaRegistry
             // if a format isn't specified, then assume text is desired.
             if (format == null)
             {
-                try
-                {
-                    Convert.FromBase64String(schemaString);
-                }
-                catch (Exception)
-                {
-                    return true; // Base64 conversion failed, infer the schemaString format is text.
-                }
-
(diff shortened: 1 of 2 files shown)


rayokota (approved):
Thanks @Havret , LGTM

PR #2376, opened 2024-12-06
Set skipKnownTypes default to be consistent with other clients

rayokota:
What
----
By default, skip known types when registering a Protobuf schema.  This is because these known types are builtin and not used when resolving a Protobuf schema.  The default is changed to match the default behavior of the other clients.

Checklist
------------------
  - Default of skipKnownTypes is changed to true to match other clients.
  - Existing tests provide sufficient coverage

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------


kcorman0 (approved):
LGTM thanks

PR #2429, opened 2025-03-25
Optimize schema registry client by caching results of LookupSchemaAsync

Havret:
## What
We have a large Protobuf schema in our system (over 300 proto dependencies), and the initial schema registration takes nearly **30 seconds**, which is far too long.  

We've identified that we can easily save around **10 seconds** by caching the results of `LookupSchemaAsync` calls.

This PR introduces that caching mechanism.  
Additionally, I’ve added a utility method for `ConcurrentDictionary` that enables safe caching of asynchronous results without requiring a global mutex.  

In future PRs, I plan to build on this utility to eliminate mutexes from other cache calls as well. This should further improve performance in concurrent scenarios involving the schema registry - such as in the `RegisterOrGetReferences` method of `ProtobufSerializer`.

---

## Checklist
- No changes in public API
- Just a perf optimization, no functional changes. 

---

## Test & Review
- Manual testing
- Full project test suite

@rayokota Could you please take a look? :)


rayokota:
@Havret , can you change the PR to just add a cache for LookupSchemaAsync in the exact same manner as the other caches?  Then we can look at switching to ConcurrentDictionary later.

Havret:
Hey @rayokota. I've updated the PR.

rayokota (approved):
Thanks @Havret , LGTM

PR #2432, opened 2025-03-27
Fix GetHashCode for IEnumerable properties

rayokota:
What
----

Checklist
------------------
- [ ] Contains customer facing changes? Including API/behavior changes 
- [ ] Did you add sufficient unit test and/or integration test coverage for this PR?
  - If not, please explain why it is not required

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------

PR #2440, opened 2025-03-31
Implement proper equality comparers for classes with Set and Dictionary properties

Havret:
What
----
While attempting to implement [this](https://github.com/confluentinc/confluent-kafka-dotnet/pull/2432#issuecomment-2759875756), I noticed that objects with `ISet`- or `IDictionary`-based properties implement equality comparison incorrectly.

Specifically, when two objects have the same set-based values but in a different insertion order (which is expected to be irrelevant for sets), the Equals method returns true, but their hash codes differ. This breaks the contract of equality in .NET, which requires that equal objects must have the same hash code.

To reproduce the issue, please check out the first commit in this PR, which contains only tests. Run it against the current implementation—you'll see that it fails.

For now, I’ve fixed the issue for just one data structure to clearly illustrate the problem.
@rayokota – is it okay if I go ahead and apply similar fixes for the remaining types, implementing proper equality comparison overloads where needed?

Checklist
------------------
  - https://github.com/confluentinc/confluent-kafka-dotnet/pull/2440/files#r2021704308

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------

<!--
Review stakeholders
------------------

### DIFF
--- a/src/Confluent.SchemaRegistry.Encryption/Confluent.SchemaRegistry.Encryption.csproj
+++ b/src/Confluent.SchemaRegistry.Encryption/Confluent.SchemaRegistry.Encryption.csproj
@@ -38,4 +38,9 @@
         <None Include="..\..\README.md" Pack="true" PackagePath="\"/>
     </ItemGroup>
 
+    <ItemGroup>
+        <Compile Include="..\Shared\SetEqualityComparer.cs" Link="Includes/SetEqualityComparer.cs"/>
+        <Compile Include="..\Shared\DictionaryEqualityComparer.cs" Link="Includes/DictionaryEqualityComparer.cs"/>
+    </ItemGroup>    
+
 </Project>
--- a/src/Confluent.SchemaRegistry/Utils.cs
+++ b/src/Confluent.SchemaRegistry/Utils.cs
@@ -27,45 +27,6 @@ namespace Confluent.SchemaRegistry
 {
     public class Utils
     {
-        public static bool DictEquals(IDictionary<string, string> a, IDictionary<string, string> b)
-        {
-            if (ReferenceEquals(a, b)) return true;
-            if (a == null || b == null) return false;
-            if (a.Count != b.Count) return false;
-            foreach (var kvp in a)
-            {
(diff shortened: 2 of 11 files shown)


src/Confluent.SchemaRegistry/Utils.cs
```diff
@@ -39,19 +39,6 @@ public static bool DictEquals(IDictionary<string, string> a, IDictionary<string,
             }
             return true;
         }
-        
-        public static bool DictEquals(IDictionary<string, ISet<string>> a, IDictionary<string, ISet<string>> b)
```
Havret, line 43: I’ve removed this since it’s no longer used. Technically, this could be considered a breaking change to the public API, though I’m not sure utility methods like this should be part of the public surface in the first place.

In the future, we could consider keeping such methods internal and use `System.Runtime.CompilerServices.InternalsVisibleTo` if we need to share them with related packages.

rayokota:
@Havret , yes please go ahead and fix the other types in your PR

Havret:
@rayokota It is ready for review.

src/Confluent.SchemaRegistry.Encryption/Confluent.SchemaRegistry.Encryption.csproj
```diff
@@ -38,4 +38,9 @@
         <None Include="..\..\README.md" Pack="true" PackagePath="\"/>
     </ItemGroup>
 
+    <ItemGroup>
+        <Compile Include="..\Shared\SetEqualityComparer.cs" Link="Includes/SetEqualityComparer.cs"/>
```
Havret, line 42: This is a pattern I came across while working on https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main. It’s particularly useful when you need to share code across different packages without exposing it through the public API, and when using `InternalsVisibleTo` isn't an option—such as with strongly signed assemblies.  

I hope you don't mind me introducing it here.

rayokota (approved):
Thanks @Havret , LGTM

Issue #2158, opened 2023-12-19
Crashes withot logs when using oauth


Cheshirre:
Description
===========
Crashes without logs when using authorization type "OAUTHBEARER".

`client.OAuthBearerSetToken(token.TokenValue, token.Expiration, **null**);`

OS -> Ubuntu Linux, MacOS, etc.
Nuget package: Confluent.Kafka, Version=2.3.0.0

How to reproduce
================
Reproduces on unix OS family: local or in docker container.
Works on windows.

Set **principal** to **null** in client.OAuthBearerSetToken(token.TokenValue, token.Expiration, **null**): https://github.com/confluentinc/confluent-kafka-dotnet/blob/master/test/Confluent.Kafka.IntegrationTests/Tests/OauthBearerToken_PublishConsume.cs 

Works, when **principal** set to **empty string**.

Checklist
=========

Please provide the following information:
 - [ ] A complete (i.e. we can run it), minimal program demonstrating the problem. No need to supply a project file.
 - [+] Confluent.Kafka nuget version.
 - [ ] Apache Kafka version.
 - [ ] Client configuration.
 - [+] Operating system.
 - [ ] Provide logs (with "debug" : "..." as necessary in configuration).
 - [ ] Provide broker log excerpts.
 - [+] Critical issue.

anchitj:
Can you provide debug logs?

Cheshirre:
> Can you provide debug logs?
There are no logs becouse this issue
https://github.com/confluentinc/confluent-kafka-dotnet/issues/1834

shlomiassaf:
Happens to me as well OSX with latest nuget

If `principal` is set to `null` it will exit with **CODE 139** in with or without a debug session.
If `principal` is set to an empty string, all is ok.

ublink:
This is also happening in docker env for me. Works fine outside of docker.

`var adminClientBuilder = new AdminClientBuilder(_adminClientConfig)
                .SetOAuthBearerTokenRefreshHandler(async (c, p) =>
                {
                    try
                    {
                        var token = await _tokenService.GetTokenAsync();
                        if (token != null && !string.IsNullOrEmpty(token.AccessToken))
                        {
                            var tokenExpiration = DateTimeOffset.UtcNow.AddMinutes(15).ToUnixTimeMilliseconds();
                            c.OAuthBearerSetToken(token.AccessToken, tokenExpiration, null);
                        }
                        else
                        {
[...]

PR #2444, opened 2025-04-02
Validate inputs in `OAuthBearerSetToken`

danwkennedy:
What
----

The documentation for `OAuthBearerSetToken` notes that `tokenValue` and `principalName` are required parameters but doesn't check them for null values. As seen in this [issue](https://github.com/confluentinc/confluent-kafka-dotnet/issues/2158), if `principalName` is `null`, the underlying library call will segfault (and do so silently in a Docker/Kubernetes environment. To keep from segfaulting, we can simply check those values for `null` and throw accordingly.

Checklist
------------------

References
----------
Issue: https://github.com/confluentinc/confluent-kafka-dotnet/issues/2158 

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/src/Confluent.Kafka/ClientExtensions.cs
+++ b/src/Confluent.Kafka/ClientExtensions.cs
@@ -41,6 +41,9 @@ namespace Confluent.Kafka
         /// <seealso cref="OAuthBearerSetTokenFailure"/>
         public static void OAuthBearerSetToken(this IClient client, string tokenValue, long lifetimeMs, string principalName, IDictionary<string, string> extensions = null)
         {
+            if (tokenValue == null) throw new System.ArgumentNullException(nameof(tokenValue));
+            if (principalName == null) throw new System.ArgumentNullException(nameof(principalName));
+
             client.Handle.LibrdkafkaHandle.OAuthBearerSetToken(tokenValue, lifetimeMs, principalName, extensions);
         }
 


src/Confluent.Kafka/ClientExtensions.cs
```diff
@@ -41,6 +41,9 @@ public static class ClientExtensions
         /// <seealso cref="OAuthBearerSetTokenFailure"/>
         public static void OAuthBearerSetToken(this IClient client, string tokenValue, long lifetimeMs, string principalName, IDictionary<string, string> extensions = null)
         {
+            if (tokenValue == null) throw new System.ArgumentNullException(nameof(tokenValue));
```
Havret, line 44: It's would be nice to have a test for that.

Havret (changes requested):
It's would be nice to have a test for that.

BraveThrasher:
For a project, I have spent the last few weeks researching and debugging a problem.
Setting up a connection with Azure Event Hubs using OAUTHBEARER, our AKS pods kept crashing with a segfault (exit code 139). 
The `principalName` being `null` was the cause.

In case this PR would have been merged in already, our project would have progressed significantly, instead it cost us time and money.

It is outrageous that the cause and solution have been known for almost two years, but that it has not yet been implemented.
Even when the heavy lifting has already been done almost a year ago by [Daniel Kennedy (@danwkennedy)](https://github.com/danwkennedy) with this PR.

PR #2448, opened 2025-04-11
Refactor serializer implementations to remove unnecessary ContinueWith calls

Havret:
What
----
Remove unnecessary `ContinueWith` call from `ExecuteRules(...)` await chain. The original use of `ContinueWith(t => (T)t.Result)` is redundant, as the result can be directly awaited and cast. This simplifies the code and improves readability without changing behavior.

Checklist
------------------
No changes in public API
Just a optimization, no functional changes.


rayokota (approved):
Thanks @Havret , LGTM

PR #2449, opened 2025-04-11
Remove shared mutex from CachedSchemaRegistryClient

Havret:
What
----
This PR follows up on https://github.com/confluentinc/confluent-kafka-dotnet/pull/2433.

It removes the shared mutex from `CachedSchemaRegistryClient`, allowing multiple concurrent calls for different resources without serializing their execution. To preserve the same semantics as the mutex-based version, a utility method `ConcurrentDictionaryExtensions.GetOrAddAsync` is introduced. This method ensures that the remote call is executed only once per key (excluding the cache cleanup scenario).

The implementation of `GetOrAddAsync` is based on a [gist by David Fowler](https://gist.github.com/davidfowl/3dac8f7b3d141ae87abf770d5781feed).

The utility class is placed in the `/shared` directory, as I plan to apply the same pattern to `AsyncSerde` to eliminate mutex from its critical path as well.

Checklist
------------------
- No changes in public API
- Just a perf optimization, no functional changes.


Havret:
@rayokota It should be good now.

rayokota (approved):
Thanks @Havret , LGTM

PR #2459, opened 2025-04-24
Protobuf serialization perf improvements

Havret:
What
----
Optimize serialization by replacing MemoryStream with pre-allocated byte buffers. This reduces allocations and improves performance.

Performance (serialization only)
```
| Method |     Mean |   Error |   StdDev |   Gen0 |   Gen1 | Allocated |
|------- |---------:|--------:|---------:|-------:|-------:|----------:|
|    New | 280.6 ns | 7.27 ns | 20.61 ns | 0.0229 |      - |     144 B |
|    Old | 444.9 ns | 8.21 ns |  7.68 ns | 0.8736 | 0.0014 |    5480 B |
```

Checklist
------------------
No changes in public API
Just an optimization, no functional changes.

## Test & Review
- Manual testing
- Full project test suite


Havret:
Sorry, I hadn’t pushed all the regenerated protos earlier. It should be all set now.

rayokota:
@Havret , why did the Protobuf classes need to be regenerated?  Is this PR incompatible with customers' older generated classes?

Havret:
If their classes were generated by `protoc` older than five years, then yes, they would need to regenerate them.  
The API I'm using here — [`WriteTo(this IMessage message, Span<byte> output)`](https://github.com/protocolbuffers/protobuf/commit/9039103637cfeba1aca82c8eead94b1479a6612d) — was introduced in that timeframe.

Realistically, they should regenerate their protos anyway, since older versions of the libraries had known CVEs.

rayokota (approved):
Thanks @Havret , LGTM

PR #2465, opened 2025-05-04
Support for schema id in header

rayokota:
What
----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/src/Confluent.SchemaRegistry/SchemaId.cs
+++ b/src/Confluent.SchemaRegistry/SchemaId.cs
@@ -1,0 +1,298 @@
+// Copyright 2025 Confluent Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Refer to LICENSE for more information.
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Net;
+using Confluent.Kafka;
+
+namespace Confluent.SchemaRegistry;
+
+/// <summary>
+///     Represents a schema ID or GUID.
+/// </summary>
+public class SchemaId
+{
+    public const string KEY_SCHEMA_ID_HEADER = "__key_schema_id";
+    public const string VALUE_SCHEMA_ID_HEADER = "__value_schema_id";
+    public const byte MAGIC_BYTE_V0 = 0;
+    public const byte MAGIC_BYTE_V1 = 1;
+
+    private const int DefaultInitialBufferSize = 1024;
+
+    private static List<int> _defaultIndex = new List<int> { 0 };
+
+    /// <summary>
+    ///     The schema type.
+    /// </summary>
+    public SchemaType SchemaType { get; }
+
+    /// <summary>
+    ///     The schema ID.
+    /// </summary>
+    public int? Id { get; set;  }
+
+    /// <summary>
+    ///     The schema GUID.
+    /// </summary>
+    public Guid? Guid { get; set;  }
+
+    /// <summary>
+    ///     The schema GUID.
+    /// </summary>
+    public List<int> MessageIndexes { get; set;  }
+
+    public SchemaId(SchemaType schemaType)
+    {
+        SchemaType = schemaType;
+    }
+
+    public SchemaId(SchemaType schemaType, int id, Guid guid)
+    {
+        SchemaType = schemaType;
+        Id = id;
+        Guid = guid;
+    }
+
+    public SchemaId(SchemaType schemaType, int id, string guid)
+    {
+        SchemaType = schemaType;
+        Id = id;
+        Guid = guid != null ? System.Guid.Parse(guid) : null;
+    }
+
+    public Stream FromBytes(Stream stream)
+    {
+        var reader = new BinaryReader(stream);
+        var magicByte = reader.ReadByte();
+        if (magicByte == MAGIC_BYTE_V0)
+        {
+            Id = IPAddress.NetworkToHostOrder(reader.ReadInt32());
+        }
+        else if (magicByte == MAGIC_BYTE_V1)
+        {
+            Guid = Utils.GuidFromBigEndian(reader.ReadBytes(16));
+        }
+        else
+        {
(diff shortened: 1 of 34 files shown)


Havret:
I'm not sure about it—it seems to partially revert https://github.com/confluentinc/confluent-kafka-dotnet/pull/2459  and introduces several unnecessary memory allocations on the critical path.

src/Confluent.SchemaRegistry/SchemaId.cs
```diff
@@ -0,0 +1,298 @@
...
+            default:
+                throw new ArgumentException($"Unknown SchemaIdDeserializerStrategy: {strategy}");
+        }
+    }
+}
+
+/// <summary>
+///     Interface for schema ID or GUID serialization.
+/// </summary>
+public interface SchemaIdSerializer
```
Havret, line 216: Typically, in .NET, interfaces are prefixed with 'I'—see `IConsumer`, `IClient`, `IProducer`, and other interfaces in this project. I’d suggest getting accustomed to this convention.

src/Confluent.SchemaRegistry/SchemaId.cs
```diff
@@ -0,0 +1,298 @@
...
+    /// </summary>
+    public int? Id { get; set;  }
+
+    /// <summary>
+    ///     The schema GUID.
+    /// </summary>
+    public Guid? Guid { get; set;  }
+
+    /// <summary>
+    ///     The schema GUID.
```
Havret, line 55: This doesn't seem right.

src/Confluent.SchemaRegistry/SchemaId.cs
```diff
@@ -0,0 +1,298 @@
...
+    ///     Serialize a schema ID/GUID.
+    ///     During serialization, the headers or payload may be mutated, and
+    ///     the result is a payload (which may contain a schema ID/GUID).
+    /// </summary>
+    ///
+    /// <param name="payload"></param>
+    /// <param name="context"></param>
+    /// <param name="schemaId"></param>
+    /// <returns></returns>
+    public byte[] Serialize(byte[] payload, SerializationContext context, SchemaId schemaId);
```
Havret, line 228: I’d suggest defining the interface along these lines (though the name might need refining):

```csharp
public interface ISchemaIdEncoder
{
    /// <summary>
    /// Encodes the schema ID or GUID into the provided buffer. The buffer is expected to be large enough 
    /// to hold the data—use <see cref="CalculateSize"/> to determine the required size.
    /// </summary>
    void Encode(Span<byte> buffer, SerializationContext context, SchemaId schemaId);

    /// <summary>
    /// Calculates the number of bytes required to encode the schema ID or GUID. 
    /// Returns 0 if the schema is encoded in headers instead of the buffer.
    /// </summary>
    int CalculateSize(SchemaId schemaId);
}
```

This approach allows us to preallocate the buffer when possible, reducing allocations. It significantly improves performance when handling a large number of messages—something I frequently encounter in typical use cases.

src/Confluent.SchemaRegistry/SchemaId.cs
```diff
@@ -0,0 +1,298 @@
...
+using System.IO;
+using System.Net;
+using Confluent.Kafka;
+
+namespace Confluent.SchemaRegistry;
+
+/// <summary>
+///     Represents a schema ID or GUID.
+/// </summary>
+public class SchemaId
```
Havret, line 28: Since a SchemaId object is created for every deserialized message, it would be ideal to make it a struct to reduce the number of heap allocations.

src/Confluent.SchemaRegistry/SchemaId.cs
```diff
@@ -0,0 +1,298 @@
...
+        if (SchemaType == SchemaType.Protobuf)
+        {
+            var size = stream.ReadVarint();
+            if (size == 0)
+            {
+                MessageIndexes = _defaultIndex;
+            }
+            else
+            {
+                MessageIndexes = new List<int>();
```
Havret, line 104: We're not actually using this for anything in the deserialization flow; nevertheless, this array is allocated for every message we consume. I'm wondering if it would be possible to detach MessageIndexes from the type representing the schema ID, since MessageIndexes are specific to Protobuf serialization anyway.

src/Confluent.SchemaRegistry/SchemaId.cs
rayokota: Fixed

src/Confluent.SchemaRegistry/SchemaId.cs
rayokota: This is only instantiated for Protobuf messages that don't used the first message type in the schema, which is uncommon

src/Confluent.SchemaRegistry/SchemaId.cs
rayokota: The functionality in this PR won't be released for a while.  Let's look at perf improvements in subsequent PRs.  Having concrete numbers would be great.

Havret: Ok

Havret (approved):
LGTM, with the caveat that we'll take a closer look at performance before releasing it.

test/Confluent.SchemaRegistry.Serdes.UnitTests/ProtoSerializeDeserialize.cs
```diff
@@ -100,6 +100,54 @@ public void UInt32SerDe()
...
+                FavoriteNumber = 100,
+                Name = "awesome",
+                OneofString = "oneof"
+            };
+
+            Headers headers = new Headers();
+            var bytes = serializer.SerializeAsync(user, new SerializationContext(MessageComponentType.Value, testTopic, headers)).Result;
+            var result = deserializer.DeserializeAsync(bytes, false, new SerializationContext(MessageComponentType.Value, testTopic, headers)).Result;
+
+            Assert.Equal("awesome", result.Name);
```
Claimundefine, line 146: nit: why is this hard-coded

test/Confluent.SchemaRegistry.Serdes.UnitTests/JsonSerializeDeserialize.cs
```diff
@@ -513,6 +513,52 @@ public async void ValidationUseLatest()
...
+                FavoriteColor = "blue",
+                FavoriteNumber = 100,
+                Name = "awesome"
+            };
+
+            Headers headers = new Headers();
+            var bytes = serializer.SerializeAsync(user, new SerializationContext(MessageComponentType.Value, testTopic, headers)).Result;
+            var result = deserializer.DeserializeAsync(bytes, false, new SerializationContext(MessageComponentType.Value, testTopic, headers)).Result;
+
+            Assert.Equal("awesome", result.Name);
```
Claimundefine, line 557: hardcoded value

Claimundefine (approved):
some nits, otherwise lgtm

PR #2469, opened 2025-05-09
Remove redundant ConfigureAwait

Havret:
What
----
This PR removes a redundant call to .ConfigureAwait(false) in a blocking context (.GetAwaiter().GetResult()), as it has no effect in such cases.
See Stephen Cleary’s [article](https://blog.stephencleary.com/2023/11/configureawait-in-net-8.html) on ConfigureAwait in .NET 8 for detailed explanation.

Checklist
------------------
* [ ] Contains customer facing changes? Including API/behavior changes
  * No
* [ ] Did you add sufficient unit test and/or integration test coverage for this PR?
  * Not required — behavior remains unchanged and is already covered.


rayokota (approved):
Thanks @Havret , LGTM

PR #2470, opened 2025-05-12
Implement IAsyncDisposable on AdminClient

verdie-g:
What
----

AdminClient.Dispose does sync-over-async which impacts the throughput of the thread pool. There is a comment on the method that says
> In the current implementation, this method may block for up to 100ms. This will be replaced with a non-blocking version in the future.

This change fixes that by implementing [IAsyncDisposable](https://learn.microsoft.com/en-us/dotnet/api/system.iasyncdisposable).

The AdminClient class partially implemented a dispose pattern (`Dispose(bool)` + `GC.SuppressFinalize(this)`) but it was missing a finalizer. Anyway, this class doesn't own any native resources so this is not needed.

Checklist
------------------
- [X] Contains customer facing changes? A new method is added to `IAdminClient`
- [X] Did you add sufficient unit test and/or integration test coverage for this PR?

### DIFF
--- a/src/Confluent.Kafka/AdminClient.cs
+++ b/src/Confluent.Kafka/AdminClient.cs
@@ -30,7 +30,7 @@ namespace Confluent.Kafka
     /// <summary>
     ///     Implements an Apache Kafka admin client.
     /// </summary>
-    internal class AdminClient : IAdminClient
+    internal sealed class AdminClient : IAdminClient
     {
         private int cancellationDelayMaxMs;
 
@@ -1654,51 +1654,51 @@ namespace Confluent.Kafka
 
         /// <summary>
         ///     Releases all resources used by this AdminClient. In the current
-        ///     implementation, this method may block for up to 100ms. This 
-        ///     will be replaced with a non-blocking version in the future.
+        ///     implementation, this method may block for up to 100ms. It is
+        ///     recommended to use <see cref="DisposeAsync"/> instead.
         /// </summary>
         public void Dispose()
         {
-            Dispose(true);
-            GC.SuppressFinalize(this);
-        }
-
-
-        /// <summary>
-        ///     Releases the unmanaged resources used by the
-        ///     <see cref="Confluent.Kafka.AdminClient" />
-        ///     and optionally disposes the managed resources.
-        /// </summary>
-        /// <param name="disposing">
-        ///     true to release both managed and unmanaged resources;
-        ///     false to release only unmanaged resources.
-        /// </param>
-        protected virtual void Dispose(bool disposing)
-        {
-            if (disposing)
+            callbackCts.Cancel();
+            try
             {
-                callbackCts.Cancel();
-                try
-                {
-                    callbackTask.Wait();
-                }
-                catch (AggregateException e)
-                {
-                    if (e.InnerException.GetType() != typeof(TaskCanceledException))
-                    {
-                        // program execution should never get here.
-                        throw e.InnerException;
-                    }
-                }
-                finally
-                {
-                    callbackCts.Dispose();
-                }
-
-                DisposeResources();
+                callbackTask.Wait();
             }
+            catch (AggregateException e)
+            {
+                if (e.InnerException.GetType() != typeof(TaskCanceledException))
+                {
+                    // program execution should never get here.
+                    throw e.InnerException;
+                }
+            }
+            finally
+            {
+                callbackCts.Dispose();
+            }
+
+            DisposeResources();
         }
 
+#if NET6_0_OR_GREATER
+        public async ValueTask DisposeAsync()
+        {
+            callbackCts.Cancel();
+            try
+            {
+                await callbackTask;
+            }
+            catch (TaskCanceledException)
+            {
+            }
+            finally
+            {
+                callbackCts.Dispose();
+            }
+
+            DisposeResources();
+        }
+#endif
 
         private void DisposeResources()
         {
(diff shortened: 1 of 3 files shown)


src/Confluent.Kafka/AdminClient.cs
```diff
@@ -1654,51 +1654,51 @@ public Handle Handle
...
 
+#if NET6_0_OR_GREATER
+        public async ValueTask DisposeAsync()
+        {
+            callbackCts.Cancel();
+            try
+            {
+                await callbackTask;
+            }
+            catch (TaskCanceledException)
```
Havret, line 1691: This approach won’t work. Calling ThrowIfCancellationRequested throws an OperationCanceledException, not a TaskCanceledException.

src/Confluent.Kafka/AdminClient.cs
```diff
@@ -1654,51 +1654,51 @@ public Handle Handle
 
         /// <summary>
         ///     Releases all resources used by this AdminClient. In the current
-        ///     implementation, this method may block for up to 100ms. This 
-        ///     will be replaced with a non-blocking version in the future.
+        ///     implementation, this method may block for up to 100ms. It is
```
Havret, line 1657: This comment may be counterintuitive, as you’re implementing IAsyncDisposable conditionally (behind a compiler flag).

test/Confluent.Kafka.UnitTests/Admin/DisposeAsync.cs
```diff
@@ -0,0 +1,15 @@
...
+using Xunit;
+
+namespace Confluent.Kafka.UnitTests;
+
+public class DisposeAsync
+{
+    [Fact]
+    public async Task TestDisposeAsyncDoesNotThrow()
+    {
+        var adminClient = new AdminClientBuilder(new AdminClientConfig { BootstrapServers = "localhost:90922" }).Build();
```
Havret, line 11: I’m not sure this tests anything. To properly verify that this method works, it would be good to have at least one iteration of the pooling loop.

Havret (changes requested):
This implementation needs some adjustments.

src/Confluent.Kafka/AdminClient.cs
verdie-g: Good catch!

src/Confluent.Kafka/AdminClient.cs
verdie-g: Updated.

test/Confluent.Kafka.UnitTests/Admin/DisposeAsync.cs
verdie-g: Yeah I'm not exactly sure how to test that since the loop doesn't do any actions when it's canceled. For example, it doesn't cancel pending tasks. I added a create/delete topic so at least it goes though the loop once.

verdie-g:
> This class owns native resources (see the DisposeResources method), including:
> • handle
> • resultQueue
> • ownedClient (optional)

- handle doesn't seem to be owned by the AdminClient
- ownedClient is managed, it may own some unmanaged fields, but it's its own responsibility to dispose them in its finalizer
- resultQueue is indeed unmanaged, I missed that one. So I completed the dispose pattern following https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync

src/Confluent.Kafka/AdminClient.cs
```diff
@@ -1651,63 +1651,72 @@ public string Name
...
-        ///     Releases the unmanaged resources used by the
-        ///     <see cref="Confluent.Kafka.AdminClient" />
-        ///     and optionally disposes the managed resources.
-        /// </summary>
-        /// <param name="disposing">
-        ///     true to release both managed and unmanaged resources;
-        ///     false to release only unmanaged resources.
-        /// </param>
-        protected virtual void Dispose(bool disposing)
+            Dispose(false);
```
Havret, line 1704: I'm not sure I understand why it's false here. Doesn't that mean ownedClient survives when we call DisposeAsync?

verdie-g: I agree it's a little confusing. See https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#the-disposeasync-method

> One primary difference in the async dispose pattern compared to the dispose pattern, is that the call from [DisposeAsync()](https://learn.microsoft.com/en-us/dotnet/api/system.iasyncdisposable.disposeasync#system-iasyncdisposable-disposeasync) to the Dispose(bool) overload method is given false as an argument. When implementing the [IDisposable.Dispose()](https://learn.microsoft.com/en-us/dotnet/api/system.idisposable.dispose#system-idisposable-dispose) method, however, true is passed instead. This helps ensure functional equivalence with the synchronous dispose pattern, and further ensures that finalizer code paths still get invoked. In other words, the DisposeAsyncCore() method will dispose of managed resources asynchronously, so you don't want to dispose of them synchronously as well. Therefore, call Dispose(false) instead of Dispose(true).

Havret: I read the cited document, but I still don't understand how it relates to our use case. In the end, we want all our disposable resources to be disposed of—regardless of whether we call Dispose or DisposeAsync. The only difference is the nature of the execution: asynchronous non-blocking vs. synchronous blocking.

The description you quoted applies more to scenarios where a resource can be disposed of through either interface—IAsyncDisposable or IDisposable. In our case, that doesn't seem to apply... or am I completely missing something

verdie-g: No you're right, I spoke too fast on that one. Just changing this line to Dispose(true) should do it.

Havret (approved):

PR #2501, opened 2025-07-18
Fix protobuf serialization - Normalize config not taken into account since 2.10.1

PierreBougon:
Fixes normalize schema passed at the wrong position in the parameter arguments in the ProtobufSerializer.cs SerializeAsync call to schemaRegistryClient.LookupSchemaAsync.

This bug was introduced with the 2.10.1, breaking our schema validation from this version while it is working fine with previous version.
After analysis the normalized parameter passed into the http call was changed from this version and I was able to trace back the issue to this change in the diff.

It might be missing a test, considering the change at the moment I only did minimal setup on my local, but if it's needed we can cover the case.

What
----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/src/Confluent.SchemaRegistry.Serdes.Protobuf/ProtobufSerializer.cs
+++ b/src/Confluent.SchemaRegistry.Serdes.Protobuf/ProtobufSerializer.cs
@@ -250,7 +250,7 @@ namespace Confluent.SchemaRegistry.Serdes
                                 .ConfigureAwait(continueOnCapturedContext: false)
                             : await schemaRegistryClient.LookupSchemaAsync(subject,
                                     new Schema(value.Descriptor.File.SerializedData.ToBase64(), references,
-                                        SchemaType.Protobuf), normalizeSchemas)
+                                        SchemaType.Protobuf), true, normalizeSchemas)
                                 .ConfigureAwait(continueOnCapturedContext: false);
 
                         // note: different values for schemaId should never be seen here.
--- a/src/Confluent.SchemaRegistry.Serdes.Protobuf/ProtobufSerializer.cs
+++ b/src/Confluent.SchemaRegistry.Serdes.Protobuf/ProtobufSerializer.cs
@@ -173,7 +173,7 @@ namespace Confluent.SchemaRegistry.Serdes
                     var schemaId = autoRegisterSchema
                         ? await schemaRegistryClient.RegisterSchemaAsync(subject, schema, normalizeSchemas).ConfigureAwait(continueOnCapturedContext: false)
                         : await schemaRegistryClient.GetSchemaIdAsync(subject, schema, normalizeSchemas).ConfigureAwait(continueOnCapturedContext: false);
-                    var registeredDependentSchema = await schemaRegistryClient.LookupSchemaAsync(subject, schema, true, normalizeSchemas).ConfigureAwait(continueOnCapturedContext: false);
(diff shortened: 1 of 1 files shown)


src/Confluent.SchemaRegistry.Serdes.Protobuf/ProtobufSerializer.cs
```diff
@@ -250,7 +250,7 @@ await RegisterOrGetReferences(value.Descriptor.File, context, autoRegisterSchema
                                 .ConfigureAwait(continueOnCapturedContext: false)
                             : await schemaRegistryClient.LookupSchemaAsync(subject,
                                     new Schema(value.Descriptor.File.SerializedData.ToBase64(), references,
-                                        SchemaType.Protobuf), normalizeSchemas)
+                                        SchemaType.Protobuf), true, normalizeSchemas)
```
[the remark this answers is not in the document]
PierreBougon: This is how it's passed from the other call in the file, I would agree that a default setting somewhere would be more descriptive but we should probably locate all relevant places to be changed to make it coherent everywhere

Havret: To avoid this parameter ordering error in the future I'd suggest using named parameters in scenarios where there are multiple consecutive boolean flags:

```csharp
await schemaRegistryClient.LookupSchemaAsync(
    subject,
    new Schema(
        value.Descriptor.File.SerializedData.ToBase64(),
        references,
        SchemaType.Protobuf
    ),
    ignoreDeletedSchemas: true,
    normalize: normalizeSchemas);

```

Havret (changes requested):
Please do not hardcode the value; instead, pass it in the correct order. It would also be nice to have a test in place so that any future regression is caught early.

src/Confluent.SchemaRegistry.Serdes.Protobuf/ProtobufSerializer.cs
PierreBougon: Absolutely fine with doing it, what should be the overall convention for this repo about bool literals or even all literals ?
As I only took the pattern that seemed the most used in the repo and copied the style from the other call in the same file to try to comply with the repo conventions.

PierreBougon: I updated with the recommended changes, if we want a broader style alignment I would suggest to make it a dedicated PR probably ? To make this fix available as soon as possible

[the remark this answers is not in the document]
PierreBougon: Used named param to make it clearer

Havret: That's probably a question for @rayokota. My remark was mainly about the correctness of the code. I'm not in a position to suggest changes to the style in this repo.

PierreBougon: Does the code looks ok to you in the current form ? I might be misunderstanding some part of your remark as I don't understand : "instead, pass it in the correct order" given that the goal of the PR is to rectify the order.
But if it was about the named parameters I am actually in favor of using them to clear up some ambiguity.

rayokota:
Thanks @PierreBougon , @Havret , the other serdes had the same problem, so created a new PR here:  https://github.com/confluentinc/confluent-kafka-dotnet/pull/2503

PierreBougon:
Good news, I guess we can close this PR now then

PR #2502, opened 2025-07-21
DGS-21472: use SocketsHttpHandler to support connection pooling for Schema Registry client, for .NET 6.0+

naxinf:
What
----

Previously there have been issues reported on using Schema Registry client in .NET, particularly when running it in serverless environments (e.g. AWS Lambda) where opening too many connections could exceed the file descriptor limit

For build targets .NET 6.0 and .NET 8.0, use `SocketsHttpHandler` to support connection pooling, via the fields:
- `MaxConnectionsPerServer` (https://learn.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler.maxconnectionsperserver?view=net-8.0): max number of connections per host
- `ConnectTimeout` (https://learn.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler.connecttimeout?view=net-8.0): timespan to wait before the connection establishing times out

Checklist
------------------

References
----------
JIRA: https://confluentinc.atlassian.net/browse/DGS-21472

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/src/Confluent.SchemaRegistry/Rest/RestService.cs
+++ b/src/Confluent.SchemaRegistry/Rest/RestService.cs
@@ -25,6 +25,7 @@ using System.Threading.Tasks;
 using X509Certificate2 = System.Security.Cryptography.X509Certificates.X509Certificate2;
 
 using System.Net.Security;
+using System.Security.Cryptography.X509Certificates;
 
 
 namespace Confluent.SchemaRegistry
@@ -40,6 +41,8 @@ namespace Confluent.SchemaRegistry
         public const int DefaultRetriesWaitMs = 1000;
 
         public const int DefaultRetriesMaxWaitMs = 20000;
+
+        public const int DefaultMaxConnectionsPerServer = 20;
 
         /// <summary>
         ///     The index of the last client successfully used (or random if none worked).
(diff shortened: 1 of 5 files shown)


rayokota (approved):
Thanks @fangnx , LGTM

src/Confluent.SchemaRegistry/Rest/RestService.cs
```diff
@@ -71,21 +74,33 @@ public RestService(string schemaRegistryUrl, int timeoutMs,
...
             this.retriesMaxWaitMs = retriesMaxWaitMs;
 
+            // For .NET 6 and later, use SocketsHttpHandler to support connection pooling
+#if NET6_0_OR_GREATER
+            this.clients = schemaRegistryUrl
+                .Split(',')
+                .Select(SanitizeUri) // need http or https - use http if not present
+                .Select(client => new HttpClient(CreateSocketsHandler(certificates, enableSslCertificateVerification, sslCaCertificate, proxy, maxConnectionsPerServer, timeoutMs))
+                {
+                    BaseAddress = new Uri(client, UriKind.Absolute),
```
daniel-petrov-flows, line 91: I believe not setting the http client timeout here like in the else block has introduced a bug, see https://github.com/confluentinc/confluent-kafka-dotnet/issues/2622.

PR #2507, opened 2025-08-06
Fixes ProtobufSerializer's reversed list of message indexes

### LINKED ISSUE #2505: Protobuf serializer serializes message type indices in reverse order since v2.10.1
The way nested protobuf schema message type indexes are encoded in the serializer provided by package `Confluent.SchemaRegistry.Serdes.Protobuf` was changed in version 2.10.1. They are now in reverse order. This causes problems for anything that tries to read it, expecting to follow the type indexing in the descriptor from the root. For one, Redpanda Console trying to list messages where there is a reverse-indexed message errors out and shows an empty list instead.

I've created a small project with a build configuration using Confluent.SchemaRegistry.Serdes.Protobuf 2.10.0 and another using 2.10.1 to demonstrate the issue:
[ProtoSchemaIndices.zip](https://github.com/user-attachments/files/21621255/ProtoSchemaIndices.zip)

When it runs, it serializes an example nested protobuf message type, InnerMessage2, and prints the bytes for the message indexes. The proto is defined as:
```proto
message OuterMessage {
  string outer_field = 1;

  message InnerMessage1 {
    string inner_field_1 = 1;
  }

  message InnerMessage2 {
    string inner_field_2 = 1;
  }
}
```

InnerMessage2's type indices should be "0, 1" because the type we are using is in the first message, and is the second message within that. The app prints the bytes, which because of the encoding, shows "2" for the value 1.

- Running `dotnet run -c v2_10_0` prints "0 2" which is correct.
- Running `dotnet run -c v2_10_1` prints "2 0" which is incorrect.

I will try submitting a PR to fix this.

mattstermiller:
What
----
Fixes #2505 

ProtobufSerializer was changed in [commit 3f10fc8f](https://github.com/confluentinc/confluent-kafka-dotnet/commit/3f10fc8f829e7540e135fe8fe343371f1d4f5f2b#diff-d709ebb934a4cf3f9b8ac7470b75cccb36598a795b54999a782bd1277972d989) in such a way that message indexes are encoded in reverse order. The message indexes are built up in reverse as it walks up the type tree. The previous code would write the message index bytes by reading from the list in reverse order, while the new code that writes the bytes in SchemaId writes them in the same order given.

I've added a unit test that uses the ProtobufSerializer to serialize the bytes for a nested proto message (copied from the integration tests) and reads it with SchemaId to assert the correct order. The ProduceConsumeNestedProtobuf integration test  did not catch this because the ProtobufDeserializer does not actually use the message indexes for anything; the message type is known at compile time and given as the generic type argument for `DeserializeAsync()`.

Checklist
------------------
- [x] Contains customer facing changes?
  - Contains behavior change to revert incorrect reversed protobuf message index order since v2.10.1
  - Added unit test in ProtobufSerializeDeserializeTests


rayokota (approved):
Thanks for catching and fixing this @mattstermiller !  LGTM

PR #2521, opened 2025-09-22
Minor fix to avoid extra trailing zeroes during serialization

rayokota:
What
----

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------


kevin-mueller-dg:
@wserr +1, if the key is serialised, this breaks the partition assignment logic and has huge downstream effects.
Basically, we ended up having the same key in different partitions.

PR #2539, opened 2025-11-03
Restservice httprequestexception propagation

mkopcil:
## **What**

There are two changes:

1. Added `AggregateException` as an inner exception of the final thrown `HttpRequestException`.
2. Added an additional try-catch block when parsing the content of a non-retriable `HttpResponse`. If the content is not JSON, we throw an `HttpRequestException` as an InnerException with the corresponding `StatusCode`.

## **Why**

1. This change was made to avoid suppressing exceptions when communicating with the `SchemaRegistry`. We discovered that when using the AVRO schema for deserializing a Kafka message, we would get an error:

   ```
   Confluent.Kafka.ConsumeException: Local: Value deserialization error ---> System.Net.Http.HttpRequestException: [https://xxx/] HttpRequestException: An error occurred while sending the request.
   ```

   without propagating the response `StatusCode`. In this case, the status was 403: Forbidden. This change makes it easier to debug networking errors.

2. Similar to the first change, this occurred when serializing a Kafka message in the producer using the AVRO schema registry. We received an error:

   ```
   Local: Value serialization error ---> ; error code: -1
   ```

   Debugging revealed that this happened when the `HttpResponse.StatusCode` was 403: Forbidden, and parsing into JSON failed. The default value of `errorCode` (-1) was then thrown as a `SchemaRegistryException`.

## **Checklist**

* [ ] Contains customer-facing changes? Including API/behavior changes

  * The exceptions thrown will remain the same, but they will now include inner exceptions with the necessary information.
* [ ] Did you add sufficient unit and/or integration test coverage for this PR?

  * Since the changes only add additional information to the exceptions, I ensured that existing tests still pass and did not add new ones.

## **Open Questions / Follow-ups**

The method `ExecuteOnOneInstanceAsync` may need further review, as these fixes might not be sufficient for different `HttpResponse` scenarios.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -177,7 +177,7 @@ after_pipeline:
       - name: SonarQube
         commands:
           - checkout
-          - sem-version java 11
+          - sem-version java 17
           - mise use dotnet@8
           - dotnet tool install --tool-path $PWD dotnet-coverage
           - artifact pull workflow unit-coverage.xml || true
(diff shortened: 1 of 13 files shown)


rayokota:
Fixed by https://github.com/confluentinc/confluent-kafka-dotnet/pull/2571

PR #2540, opened 2025-11-03
Ensure schemaId initialization is thread-safe

### LINKED ISSUE #2538: Concurrency bug in ProtobufSerializer leading to invalid messages being produced
We have observed random messages with invalid 5-byte value prefix being produced to our topics after upgrading from 2.4 to 2.10. We have then analyzed the `ProtobufSerializer` and found what we believe to be a serious concurrency bug when multiple threads produce using same Producer instance at same time.

Consider lines 231-285 in `ProtobufSerializer.cs` and consider case where two threads A and B execute concurrently:

1. Thread A enters `serdeMutex` and creates new `SchemaId` instance in line 240. When it leaves the lock at line 268, the `schemaId` class member variable is set to an **incomplete**  `SchemaId` with no `MessageIndices` set.
2. Thread A sets the `MessageIndexes` in line 285.
3. Now Thread B enters the `serdeMutex` and overwrites the `schemaId` class member with a new instance at line 240. This new instance is **incomplete** without `MessageIndexes`.
4. Thread A now uses the incomplete `schemaId` member to write a message value in lines 295-299. This includes a call to `SchemaId.WriteArrayIndexesToBytes` which silently skips the mandatory "message index" byte from value prefix leading to an **invalid 5-byte**  prefix being written.

The root cause is that all threads manipulate the same shared member, partly outside the lock. One fix could be to only assign to the `schemaId` class member when assigned value is complete.

rayokota:
What
----
Ensure schemaId initialization is thread-safe

Fixes https://github.com/confluentinc/confluent-kafka-dotnet/issues/2538

Checklist
------------------

References
----------
JIRA: 

Test & Review
------------

Open questions / Follow-ups
--------------------------

### DIFF
--- a/src/Confluent.SchemaRegistry.Serdes.Protobuf/ProtobufSerializer.cs
+++ b/src/Confluent.SchemaRegistry.Serdes.Protobuf/ProtobufSerializer.cs
@@ -237,8 +237,7 @@ namespace Confluent.SchemaRegistry.Serdes
                     
                     if (latestSchema != null)
                     {
-                        schemaId = new SchemaId(SchemaType.Protobuf, latestSchema.Id, latestSchema.Guid);
-                        schemaId.MessageIndexes = indexArray;
+                        schemaId = new SchemaId(SchemaType.Protobuf, latestSchema.Id, latestSchema.Guid, indexArray);
                     }
                     else if (!subjectsRegistered.Contains(subject))
                     {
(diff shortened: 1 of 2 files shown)


tobiogunbi (approved):

JRobTS:
I applaud this fix. We actually ran into this race condition in testing (we happened to be using 2.12.0 from Oct).

PR #2571, opened 2026-02-18
Preserve inner exceptions in RestService

### LINKED ISSUE #1437: Don't throw away the inner exception
Description
===========

The RestClient for SchemaRegistry throws away the inner exception:

https://github.com/confluentinc/confluent-kafka-dotnet/blob/9c5dfffbbd8ae80e3a08494a48b9a239b9872ffa/src/Confluent.SchemaRegistry/Rest/RestService.cs#L248

So that error messages become:

image

without an inner exception. Simply pass the caught exception into the c'tor to fix this issue.

How to reproduce
================

Make the schema registry TLS endpoint be something that this client doesn't like. 

Similarly, it's impossible to override the logic for validating TLS certificates due to how the HttpClient is created:

https://github.com/confluentinc/confluent-kafka-dotnet/blob/9c5dfffbbd8ae80e3a08494a48b9a239b9872ffa/src/Confluent.SchemaRegistry/Rest/RestService.cs#L68-L75

This is doubly annoying because it would seem that, by explicitly creating a new `HttpRequestMessage` in `CreateRequest`,  the global `ServicePointManager.ServerCertificateValidationCallback` that I tried setting, is overridden and defaulted.

Finally it's annoying because it would seem that on macOS, .Net Core doesn't default to using the system-configured certificates, so even if this certificate is valid for me outside .net, it's not inside .net.

Checklist
=========

I haven't followed it because this issue is a review of the source code.

 - [ ] A complete (i.e. we can run it), minimal program demonstrating the problem. No need to supply a project file.
 - [ ] Confluent.Kafka nuget version.
 - [ ] Apache Kafka version.
 - [ ] Client configuration.
 - [ ] Operating system.
 - [ ] Provide logs (with "debug" : "..." as necessary in configuration).
 - [ ] Provide broker log excerpts.
 - [ ] Critical issue.

yykkibbb:
## Summary
- Fixes #1437
- Preserve inner exceptions in `RestService.ExecuteOnOneInstanceAsync` instead of discarding them
- Collect caught `HttpRequestException`s and pass them as `innerException` (single → direct, multiple → `AggregateException`)
- Pass the caught exception as `innerException` for the Unauthorized case

## Test plan
- [x] Added unit tests verifying inner exception preservation for single and multiple URLs
- [x] All 85 existing unit tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# 2.13.1
+
+## Fixes
+
+* Processing arrays and maps sequentially to avoid race conditions (#2569)
+
+
 # 2.13.0
 
 ## Enhancements
--- a/examples/AdminClient/AdminClient.csproj
+++ b/examples/AdminClient/AdminClient.csproj
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <!-- nuget package reference: <PackageReference Include="Confluent.Kafka" Version="2.13.1-RC1" /> -->
+    <!-- nuget package reference: <PackageReference Include="Confluent.Kafka" Version="2.13.1" /> -->
     <ProjectReference Include="../../src/Confluent.Kafka/Confluent.Kafka.csproj" />
   </ItemGroup>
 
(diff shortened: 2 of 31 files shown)


rayokota (approved):
Thamks @yykkibbb , LGTM

PR #2627, opened 2026-06-19
Add RawProducer overload accepting two header sets

nanov:
## What

Adds a `RawProduce` overload that takes **two** `KafkaHeaders` sets and writes them into a **single** message — entries concatenated in order (`headers1` then `headers2`). This lets callers combine a shared/common header set with per-message headers without copy-merging them into one collection.

## Changes

**`IRawProducer.cs`**
- Two public `RawProduce(..., in KafkaHeaders headers1, in KafkaHeaders headers2, ...)` overloads (default-partition and explicit-partition).
- Internal `ProduceRawWithHeaders(..., in KafkaHeaders headers1, in KafkaHeaders headers2, ...)`.

**`RawProducer.cs`**
- Extracted the per-header add loop into a private `AddHeadersToHandle` helper (no behavior change to the existing path).
- Added `BuildHeadersHandle(in KafkaHeaders first, in KafkaHeaders second)` — allocates the native list sized to `first.Count + second.Count`, appends both sharing one `stackalloc` name buffer. Same ownership/cleanup contract as the single-set builder.
- Implemented the two-set `ProduceRawWithHeaders` and the two public `RawProduce` overloads, reusing the existing `MSG_F_COPY` + pinning pattern.

**Tests**
- `Produce_WithTwoHeaderSets_Concatenated_RoundTrip` — produces with a 2-entry shared set + 1-entry per-message set and asserts the consumer reads all three in order.

## Notes

- Matched existing file style (Allman / 4-space) to minimize upstream merge drift on these synced files.
- Fork-specific files (not upstream-patched), so no `sync-upstream.sh` change needed.
- `RawProducerMarshal` (no-copy API) left untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nanov:
Opened against the wrong repo (intended for my fork). Closing.

PR #2646, opened 2026-07-27
Add s390x CI coverage, and fix CSFLE ciphertext version byte order on big-endian

Devarsh010:
Adds CI coverage for `linux-s390x` (IBM Z, big-endian), and fixes a real byte-order bug
found while doing it.

**1. A CSFLE bug : (`EncryptionExecutor.cs`).**
`PrefixVersion`/`ExtractVersion` paired `BinaryWriter`/`BinaryReader` with
`IPAddress.HostToNetworkOrder`/`NetworkToHostOrder`. `BinaryWriter`/`BinaryReader` are
always little-endian on every platform, while `HostToNetworkOrder` only byte-swaps on
little-endian hosts and is a no-op on big-endian ones , so the pair produces network byte
order *only* on little-endian. On s390x the ciphertext DEK-version prefix was written
little-endian, meaning **encrypted fields could not be exchanged across architectures**: a
consumer there reads a version of `1` written by an x86 producer as `16777216` and fails
with `No dek found for <kek> during consume`.

Fixed with explicit big-endian conversion. **On little-endian hosts the emitted bytes are
byte-identical to before**, so existing data and existing producers are unaffected .

**2. An s390x CI block.** Semaphore has no native s390x agent, so an amd64 agent offloads over
SSH to a dedicated IBM Z host and runs the tests in Red Hat's UBI9 .NET image , the same
approach `librdkafka` uses (confluentinc/librdkafka#5458). Runs the four unit test projects.

**3. Supporting changes.** This branch merges current master, whose test suite is now on xunit.v3 / Microsoft.Testing.Platform (#2656); scripts/run-tests-s390x.sh invokes the s390x tests accordingly (build per-project, then dotnet test --no-build) , validated on a native s390x host, 770 tests, 0 failures. Two unit tests had a latent little-endian assumption in their setup (the serializers themselves were already correct). README/CHANGELOG updated.


rayokota (approved):
Thanks @Devarsh010 , LGTM

PR #2656, opened 2026-08-25
Fix NJsonSchema security warnings and migrate Kafka tests to xunit.v3

emasab:
## Summary

- Bumps `NJsonSchema` -> `NJsonSchema.NewtonsoftJson` 11.0.2 in the OAuth examples (`OAuthConsumer`, `OAuthOIDC`, `OAuthProducer`) to resolve security advisories against the old `NJsonSchema` 10.8.0, and drops the now-redundant `System.Net.Http` reference from `ConfigGen`.
- That NJsonSchema bump surfaced a pre-existing bump of `Confluent.Kafka.UnitTests` to `Microsoft.NET.Test.Sdk` 18.9.0 / `xunit.v3` 4.0.0 / `Moq` 4.20.72, which no longer built. This PR completes that migration properly for both `Confluent.Kafka.UnitTests` and `Confluent.Kafka.IntegrationTests`, and cleans up every warning the new xunit.v3 analyzers surfaced along the way.

## Details

**Build fixes (xunit v2 -> v3):**
- Converted `async void` test methods to `async Task` — xunit.v3 removed support for `async void` tests (`xUnit1049`).
- Replaced `Xunit.SkippableFact`/`Skip.If` (incompatible with xunit.v3's execution engine) with xunit.v3's native dynamic-skip API, `Assert.SkipWhen`.

**Warning cleanup (xunit.v3 analyzer):**
- Converted test methods that blocked on `.Wait()`/`.Result` to `async`/`await`, passed `TestContext.Current.CancellationToken` to cancellable APIs, and unwrapped `catch (AggregateException) { ex.InnerException }` patterns to catch the concrete exception type directly now that `await` propagates it unwrapped (`xUnit1031`/`xUnit1051`).
- A handful of tests deliberately block on purpose (thread-pool-exhaustion/deadlock-avoidance tests, sync-over-async tests, and one test that intentionally verifies `.Wait()`'s `AggregateException`-wrapping behavior) — those were left blocking, with a scoped `#pragma warning disable/restore` and a comment explaining why.
- Converting a couple of methods to `async` surfaced two pre-existing fire-and-forget bugs (an admin-client call that was never awaited) — fixed alongside the warning cleanup.
- Fixed assertion-style warnings: `Assert.Equal` argument order, `Assert.True(false, msg)`/`Assert.False(true, msg)` -> `Assert.Fail(msg)`, and `Assert.Empty`/`Assert.Single` over a `.Where(...)` -> `Assert.DoesNotContain`/`Assert.Single` with a predicate (`xUnit2000`/`xUnit2020`/`xUnit2029`/`xUnit2031`).

## Test plan

- [x] `Confluent.Kafka.UnitTests` builds with 0 warnings / 0 errors on `net8.0` and `net10.0`; all 259 tests pass on both target frameworks.
- [x] `Confluent.Kafka.IntegrationTests` builds with 0 warnings / 0 errors on `net8.0` and `net10.0`.
- [x] Ran the full `Confluent.Kafka.IntegrationTests` suite against a live local cluster: 99 total, 0 errors, 0 failed, 1 skipped (pre-existing `[SKIP] FIXME: Review and fix this test`).

### DIFF
--- a/test/Confluent.SchemaRegistry.UnitTests/RestServiceTests.cs
+++ b/test/Confluent.SchemaRegistry.UnitTests/RestServiceTests.cs
@@ -170,14 +170,14 @@ namespace Confluent.SchemaRegistry.UnitTests
                 {
                     while (true)
                     {
-                        accepted.Add(await listener.AcceptTcpClientAsync());
+                        accepted.Add(await listener.AcceptTcpClientAsync(TestContext.Current.CancellationToken));
                     }
                 }
                 catch
                 {
                     // listener stopped - expected on cleanup.
                 }
-            });
+            }, TestContext.Current.CancellationToken);
 
             try
             {
@@ -202,7 +202,7 @@ namespace Confluent.SchemaRegistry.UnitTests
                 // enumerating the accepted clients, to avoid mutating the list
                 // while it is being read.
                 listener.Stop();
-                await acceptLoop.ConfigureAwait(false);
+                await acceptLoop;
                 foreach (var client in accepted)
                 {
                     client.Dispose();
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -65,10 +65,10 @@ blocks:
             - powershell -ExecutionPolicy ByPass -File dotnet-install.ps1 -Channel 10.0 -Quality GA -InstallDir C:\dotnet
             - $Env:Path += ";C:\dotnet"
             - dotnet restore
-            - dotnet test -f net10.0 -c ${CONFIGURATION} --no-build test/Confluent.Kafka.UnitTests/Confluent.Kafka.UnitTests.csproj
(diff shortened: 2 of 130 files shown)


test/Confluent.SchemaRegistry.UnitTests/RestServiceTests.cs
```diff
@@ -202,7 +202,7 @@ public async Task Timeout_IsTreatedAsNetworkError_AndDoesNotPropagateRaw()
                 // enumerating the accepted clients, to avoid mutating the list
                 // while it is being read.
                 listener.Stop();
-                await acceptLoop.ConfigureAwait(false);
+                await acceptLoop;
```
prashah-confluent, line 205: nit: Dropped on purpose `.ConfigureAwait(false)` ?

prashah-confluent (approved):
1 minor nit, otherwise LGTM !! Thanks @emasab

